PR #22, opened 2026-07-01
fix: use v3 stream events for agent output

christian-bromann:
## Summary
- Switches the agent runtime from `stream` with explicit stream modes to `streamEvents` v3.
- Parses message and tool protocol events directly using `@langchain/protocol` types.
- Removes legacy tuple/subgraph chunk normalization and reduces debug noise for expected protocol channels.


colifran (dismissed):
Nice!

colifran (approved):

PR #24, opened 2026-07-01
feat: add `--config` to reconfigure credentials without running the agent

### LINKED ISSUE #23: No way to reconfigure API key once credentials exist; `--init` immediately calls the API
## Summary

Once `~/.openwiki/.env` holds a full set of credentials, there is no way to reconfigure them from the CLI. `openwiki --init` — which the README describes as *"configure your model and API key, then generate documentation"* — skips setup entirely and immediately fires an agent request against the existing key. A user whose key is dead (expired trial, exhausted quota, wrong tier) gets an error on boot and no path to switch keys.

## Steps to reproduce

1. Have a complete `~/.openwiki/.env` (provider + API key + model + `LANGSMITH_API_KEY`), where the OpenAI key has no remaining quota.
2. Run `openwiki --init`.

## Observed

The agent runs on mount and returns immediately:

```
* Error 429 You exceeded your current quota, please check your plan and billing details.
```

Credential diagnostics confirm every value is present and loaded, so setup is considered "done":

```
OPENWIKI_PROVIDER   source=process.env over ~/.openwiki/.env  preview="openai"
OPENAI_API_KEY      source=process.env over ~/.openwiki/.env  length=164
OPENWIKI_MODEL_ID   source=process.env over ~/.openwiki/.env  preview="gpt-5.4-mini"
LANGSMITH_API_KEY   source=process.env over ~/.openwiki/.env  length=0
```

There is no way to reach the key prompt again:
- `--init` maps to `command: "init"` with `shouldStart: true`, so the App effect calls `runOpenWikiAgent` on mount (`commands.ts`, `cli.tsx`).
- The only pre-run interceptor is `shouldRunInteractiveCredentialSetup`, gated on `needsCredentialSetup()`. That returns `true` only when provider / key / model / LangSmith are **missing** (`credentials.ts`). With a full `.env`, `getInitialStep` returns `null` and `InitSetup` completes without prompting.
- Interactive chat has no key command. Slash commands are `provider / model / init / update / clear / help / exit`; `/provider` only writes the provider + default model and prints "ensure KEY is set" — nothing lets you enter an API key.

Net: `InitSetup` is only ever reachable as a side effect of *missing* credentials, and `--init` conflates "configure" with "generate docs."

## Secondary: env-precedence footgun

`loadOpenWikiEnv` only sets `process.env[key]` when it is `undefined`, so an exported shell var wins over `~/.openwiki/.env` (diagnostics: `source=process.env over ~/.openwiki/.env`). Editing the `.env` — or a future reconfigure flow — takes effect in the current process (setup patches `process.env`), but a fresh shell that still exports the old key silently reverts. Setup should warn when the key it saves is shadowed by an exported var.

## Proposed fix

- Add a dedicated, side-effect-free config entry point (`openwiki --config`) that renders `InitSetup` unconditionally and never starts an agent run.
- Give `InitSetup` a reconfigure mode that walks every step even when values already exist, pre-selecting current values, so credentials are updated rather than skipped — and never nulls out settings the user leaves untouched.
- (Follow-ups) a `/key` slash command for the in-session case; a hint on 429/`insufficient_quota` errors pointing at `--config`; the env-shadow warning above.

bath-tub:
<h4>Description</h4>

Adds `openwiki --config`, a settings-only entry point for changing the provider, API key, model, or LangSmith key.

Today, once `~/.openwiki/.env` holds a complete set of credentials there is no way to change them from the CLI:

- `--init` maps to a documentation run (`shouldStart: true`), so it invokes the agent on mount. A user whose key is dead (expired trial, exhausted quota, wrong tier) just gets an error on boot with no path to switch keys.
- `needsCredentialSetup()` is only `true` when a value is *missing*, so `InitSetup` is never reached when the file is complete.
- Interactive chat has no key command — `/provider` only writes the provider + default model and prints "ensure KEY is set".

This fixes https://github.com/langchain-ai/openwiki/issues/23

<h4>What changed</h4>

- **`--config` flag** (`commands.ts`): new `{ kind: "config" }` command. It errors if combined with `--init`/`--update`/`--print`/a message, and requires a TTY.
- **`reconfigure` mode in `InitSetup`** (`credentials.tsx`): walks every step (provider → key → model → LangSmith) even when values already exist.
- **Non-destructive by design**: on the API key and LangSmith prompts, an empty entry means *keep the current value* — `completeSetup` only writes non-null fields, so untouched secrets are never cleared. A saved custom model that isn't one of the provider presets is surfaced as a selectable "current" option so pressing Enter keeps it instead of silently falling back to the first preset.
- **Never runs the agent** (`cli.tsx`): the config flow renders the outcome and exits; the agent-run effect is gated on `command.kind === "run"`. Env is now loaded for `--config` too so existing values are detected.
- Docs + help updated.

<h4>Testing</h4>

`pnpm build`, `pnpm lint:check`, `pnpm format:check` all pass. Drove the interactive flow through a PTY against an isolated `HOME` with a complete `.env`:

- **Keep all** (Enter through every step): every value preserved, no network call.
- **Change key** (type a new key, keep the rest): key replaced; provider, model, and LangSmith key preserved.
- **Custom model + keep all**: a non-preset `OPENWIKI_MODEL_ID` is preserved rather than overwritten.

<h4>Follow-ups (not in this PR)</h4>

- A `/key` (or `/login`) slash command for the in-session case.
- A hint on 429/`insufficient_quota` errors pointing at `--config`.
- Warn when the key being saved is shadowed by an exported shell var (`loadOpenWikiEnv` only fills `process.env` when unset, so an exported key wins over `~/.openwiki/.env` on the next shell).

Made with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,17 @@ Update existing documentation:
 openwiki --update
 ```
 
+Reconfigure your provider, API key, model, or LangSmith key without running the
+agent (for example, to switch to a different key):
+
+```sh
+openwiki --config
+```
+
+`--config` walks the setup prompts even when credentials already exist. Press
+Enter on the API key or LangSmith prompt to keep the current value; only what you
+retype is changed.
+
 Show help:
 
 ```sh
--- a/src/cli.tsx
+++ b/src/cli.tsx
@@ -70,6 +70,11 @@ type RunState =
       errorDiagnostics?: ErrorDiagnostic[];
     };
 
+type ConfigState =
+  | { status: "editing" }
+  | { status: "saved"; result: InitSetupResult }
+  | { status: "error"; message: string };
+
 type RunLogItem = {
   actionCount?: number;
   activeToolCallIds?: string[];
@@ -136,6 +141,9 @@ function App({ command }: AppProps) {
   const activeRunLog = useRef<RunLogItem[]>([]);
   const [runState, setRunState] = useState<RunState>({ status: "idle" });
   const [completedRuns, setCompletedRuns] = useState<CompletedRun[]>([]);
+  const [configState, setConfigState] = useState<ConfigState>({
+    status: "editing",
+  });
   const [activeUserMessage, setActiveUserMessage] = useState<string | null>(
     command.kind === "run" ? command.userMessage : null,
   );
@@ -225,13 +233,13 @@ function App({ command }: AppProps) {
       return;
     }
 
-    if (command.dryRun) {
-      process.exitCode = 0;
-      app.exit();
+    if (command.kind !== "run") {
       return;
     }
 
-    if (command.kind !== "run") {
+    if (command.dryRun) {
+      process.exitCode = 0;
+      app.exit();
       return;
     }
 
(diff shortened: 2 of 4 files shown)


colifran:
Thanks for PR! This will be superseded by: https://github.com/langchain-ai/openwiki/pull/374 which makes init idempotent so re-running allows updating config.

Issue #21, opened 2026-07-01
initialization failed


christian-bromann:
Installed openwiki globally, ran init, selected GPT-5.5 as model and put in my LangSmith key, then got this:

```
openwiki --init
  ___                  __        ___ _    _ 
 / _ \ _ __   ___ _ __ \ \      / (_) | _(_)
| | | | '_ \ / _ \ '_ \ \ \ /\ / /| | |/ / |
| |_| | |_) |  __/ | | | \ V  V / | |   <| |
 \___/| .__/ \___|_| |_|  \_/\_/  |_|_|\_\_|
      |_|

╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ >_ OpenWiki v0.0.1 agent docs for codebases                                                                                               │
│ provider: OpenAI                                                                                                                          │
│ model: gpt-5.5                                                                                                                            │
│ directory: ~/Sites/LangChain/managed-deepagents-sdk                                                                                       │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

* LangSmith tracing enabled - Run failed
Tip: ask for a docs change, or use /exit when you are done.

* Error 400 Invalid value: 'file'. Supported values are: 'text', 'refusal', 'image_url', and 'input_audio'.
# Credential Diagnostics
  Raw secret values are intentionally not printed.

  OPENWIKI_PROVIDER source=process.env over ~/.openwiki/.env
  length=6 preview="openai"
  warnings=none

  BASETEN_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  FIREWORKS_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  OPENAI_API_KEY source=process.env
  length=51 preview="lsv2_s...b552"
  warnings=none

  ANTHROPIC_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  OPENROUTER_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  OPENWIKI_MODEL_ID source=process.env over ~/.openwiki/.env
  length=7 preview="gpt-5.5"
  warnings=none

  LANGSMITH_API_KEY source=process.env over ~/.openwiki/.env
  length=51 preview="lsv2_p...3fc9"
  warnings=none
```

YanivBohbot:
i have the same problem

MRL-00:
Yep - same problem here also

[...]

PR #28, opened 2026-07-02
feat: add Ollama provider and --init CLI flags

### LINKED ISSUE #26: Support custom OpenAi-Compatible providers
I would love to use this at my job but I can’t without being able to specify an openai compatible base url

rohitvipin-inapp:
This PR adds Ollama as a supported inference provider and introduces non-interactive flags for `openwiki --init`.

### Motivation

- Closes #26 — adds support for custom OpenAI-compatible providers via `--provider`, `--base-url`, `--baseurl`, and `--model` flags.
- Addresses #23 — `--init` can now reconfigure provider, model, and API credentials without immediately calling the API when flags are provided.

### Changes

- Add `@langchain/ollama` dependency
- Add Ollama provider config, env keys, and onboarding flow
- Support custom base URL collection for Ollama and OpenAI
- Add `--provider`, `--base-url`, `--baseurl`, and `--model` CLI arguments to `openwiki --init`
- Update README to document new providers and flags

### Verification

- `pnpm install --frozen-lockfile` ✅
- `pnpm run build` ✅
- `pnpm run pack:check` ✅

### Related issues

- Closes #26
- Related to #23
- May help with #21 (init-time failures due to forced API calls)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -51,3 +51,27 @@ jobs:
 
       - name: Check lint
         run: pnpm run lint:check
+
+  pack:
+    name: Pack
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          run_install: false
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build and pack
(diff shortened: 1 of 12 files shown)


rajpratham1 (changes requested):
Thanks for the feature—adding Ollama support and non-interactive init options is a nice improvement. I noticed one issue in runNonInteractiveInit(): after successfully saving credentials, the summary prints the base URL using process.env[baseUrlEnvKey]. If the user supplied --base-url, that value isn't guaranteed to be in process.env on this code path, so the confirmation output can be incorrect or empty even though the configuration was saved successfully. It would be better to print the resolved base URL that was passed to completeCredentialSetup() (or the resolved local variable) instead.

bracesproul:

Issue #29, opened 2026-07-02
Add support for gemini


oshai:
Will it make sense to also support gemini or the documentation structure differ?

dcorto:
Any update on this? i pay for gemini AI and it will be great for me use it 😂🔝


Thanks a lot

oshai:
I saw google published okf: https://cloud.google.com/blog/products/data-analytics/how-the-open-knowledge-format-can-improve-data-sharing

How will it work with this project?

victor-mico:
Gemini works today via the `openai-compatible` provider — no code changes needed, just a couple of gotchas that took some digging to isolate:

**Setup**

- Provider: `openai-compatible`
- `OPENAI_COMPATIBLE_BASE_URL`: `https://generativelanguage.googleapis.com/v1beta/openai/`
- `OPENAI_COMPATIBLE_API_KEY`: your Gemini API key (from Google AI Studio)
- `OPENWIKI_MODEL_ID`: a **Gemini 2.5** model — `gemini-2.5-flash`, `gemini-2.5-flash-lite`, or `gemini-2.5-pro`

**Gemini 3.x doesn't work (yet)**

Any Gemini 3.x model (`gemini-3.5-flash`, `gemini-3.1-*`, `gemini-3-flash-preview`, and the `gemini-*-latest` aliases, which currently resolve to 3.x) fails partway through a run with:

400 status code (no body)

The real error (visible with `OPENAI_LOG=debug`) is:

Function call is missing a thought_signature in functionCall parts. This is required
for tools to work correctly...

Gemini 3.x's OpenAI-compat endpoint requires every tool-call message to echo back a `thought_signature` blob from the model's prior turn. LangChain's generic `ChatOpenAI` client (which OpenWiki uses) doesn't know about this Gemini-specific field, so it gets dropped when reconstructing conversation history for the next turn — and since OpenWiki's agent loop isut, this breaks every real run on any 3.x model. Confirmed viadirect curl reproduction against the API, independent of OpenWiki.

Gemini 2.5 models don't have this requirement and work end-to-pporting Gemini 3.x properly would need OpenWiki (orlangchain's OpenAI-compat layer) to capture and replay `extra_content.google.thought_signature` on tool-call messages.

dcorto:
> Gemini works today via the `openai-compatible` provider — no code changes needed, just a couple of gotchas that took some digging to isolate:
> 
> **Setup**
> 
> - Provider: `openai-compatible`
[...]

PR #32, opened 2026-07-02
feat: add reconfigure option, support Gemini and custom providers, and resolve role content error

STiFLeR7:
This pull request addresses several key defects and enhancements:

1. **Reconfigure API key credentials (Issue #23)**:
   - Adds a dedicated `--config` flag to unconditionally run the interactive credentials setup and exit cleanly.
   - Updates `InitSetup` with a `reconfigure` mode that displays current settings and allows users to press **Enter** to keep existing configuration values (instead of requiring re-entry or resetting them).

2. **Support for Gemini (Issue #29)**:
   - Integrates the `gemini` provider using the OpenAI-compatible endpoint (`https://generativelanguage.googleapis.com/v1beta/openai`) and `GEMINI_API_KEY`, pre-populating Gemini flash and pro model presets.

3. **Custom OpenAI-compatible base URL (Issue #26)**:
   - Configures a `custom` provider allowing users to configure any custom OpenAI-compatible API base URL (`OPENWIKI_CUSTOM_BASE_URL`) and custom API key (`OPENWIKI_CUSTOM_API_KEY`).

4. **Invalid value: 'file' error (Issue #21)**:
   - Implements a custom `SanitizedChatOpenAI` wrapper that filters and sanitizes message content blocks before sending them to the API. If the model is not a native Responses-API model, it dynamically replaces `{ type: 'file' }` blocks with a text-based block containing file metadata, preventing completions from crashing when loading binary-read tool results from the SQLite checkpointer.

5. **Environment Precedence Warnings**:
   - Warns users in the diagnostics display if a saved key is shadowed by a shell-exported environment variable (`process.env over ~/.openwiki/.env`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "scripts": {
     "openwiki": "node dist/cli.js",
     "build": "tsc -p tsconfig.json",
-    "clean": "rm -rf dist",
+    "clean": "node -e \"require('fs').rmSync('dist', { recursive: true, force: true })\"",
     "dev": "tsx src/cli.tsx",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -2,6 +2,7 @@ import { createHash } from "node:crypto";
 import { chmod, mkdir } from "node:fs/promises";
 import path from "node:path";
 import { ChatAnthropic } from "@langchain/anthropic";
+import { BaseMessage } from "@langchain/core/messages";
 import { SqliteSaver } from "@langchain/langgraph-checkpoint-sqlite";
 import { ChatOpenAI } from "@langchain/openai";
 import { ChatOpenRouter } from "@langchain/openrouter";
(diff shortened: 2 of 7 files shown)


rajpratham1 (changes requested):
The new SanitizedChatOpenAI only overrides _generate, _streamChatModelEvents, and _streamResponseChunks, but ChatOpenAI has other execution paths (e.g. batching or future internal methods) that can bypass these overrides. That means the "file → text" sanitization isn't guaranteed to be applied consistently, and the original role/content error can still reappear depending on which LangChain API path is used. Consider applying the sanitization at a single shared boundary (before invoking the model) rather than relying on overriding several transport methods. That keeps the behavior consistent regardless of how the underlying client evolves.

STiFLeR7:
I have refactored the solution to use an ES6 Proxy wrapper (wrapModelWithSanitizer) at the model creation boundary in src/agent/index.ts, ensuring that the message sanitization is consistently applied across all LangChain execution methods. All builds and tests are passing successfully.

colifran:
Hi! Thanks for the PR! This is superseded by: https://github.com/langchain-ai/openwiki/issues/23. We also recently added support for gemini as a provider. If there are additional parts of this PR please open PRs for each as per the contributing guidelines: https://github.com/langchain-ai/openwiki/blob/main/CONTRIBUTING.md

PR #35, opened 2026-07-02
fix: support OPENAI_BASE_URL and api-key header for Azure APIM-style gateways

romiluz13:
## Problem

Two bugs when using OpenWiki with OpenAI-compatible gateways (e.g. Azure APIM, Grove):

**1. Wrong auth header (401)**
These gateways use an `api-key` request header instead of `Authorization: Bearer`. Every request failed with 401 because the OpenAI SDK sends the wrong header.

Additionally, `OPENAI_BASE_URL` was in the `deprecatedEnvKeys` list in `src/env.ts`, so even if a user set it in `~/.openwiki/.env` it was silently dropped on load.

**2. File content blocks rejected (400)**
DeepAgents reads repo files and sends them as `{type:'file', mimeType, data}` content blocks in the message history. Third-party gateways (e.g. GLM-5.2 via Azure APIM) only accept plain string message content and reject these with:
> `Input should be a valid string, field: messages[N].content.str, value: [{'type': 'file', ...}]`

## Fix

**`src/constants.ts`** — export `OPENAI_BASE_URL_ENV_KEY` constant.

**`src/env.ts`**
- Remove `OPENAI_BASE_URL` from `deprecatedEnvKeys` so it's no longer stripped on load/save.
- Add it to `managedEnvKeys` so it's persisted and included in diagnostics.

**`src/agent/index.ts`** — two changes in `createModel()`:
- Read `OPENAI_BASE_URL` from the environment; pass `api-key` as a `defaultHeaders` entry for Azure APIM-style auth.
- When `OPENAI_BASE_URL` is set, use `TextOnlyChatOpenAI` — a thin `ChatOpenAI` subclass that intercepts `_generate` and `_streamResponseChunks` to decode base64 file blocks into `{type:'text'}` blocks before they reach the API. Standard providers are unaffected.

## Usage

```sh
export OPENWIKI_PROVIDER=openai
export OPENAI_API_KEY="<your-gateway-key>"
export OPENAI_BASE_URL="https://your-gateway.azure-api.net/openai/v1"
export OPENWIKI_MODEL_ID="your-model-id"
openwiki --init
```

Or persist once in `~/.openwiki/.env`:
```
OPENWIKI_PROVIDER="openai"
OPENAI_API_KEY="<your-gateway-key>"
OPENAI_BASE_URL="https://your-gateway.azure-api.net/openai/v1"
OPENWIKI_MODEL_ID="your-model-id"
```

## Test

Verified end-to-end against a live Azure APIM gateway (Grove) with `FW-GLM-5.2` on a real Python repo — previously 401 on hello-world, then 400 on file blocks during `--init`; both now work correctly.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


romiluz13:
Closing for now — the approach works but needs more testing before it's ready for upstream.

rajpratham1 (changes requested):
OPENAI_BASE_URL is added as a supported configuration, but it's still listed in deprecatedEnvKeys. That means users who intentionally configure a custom gateway via OPENAI_BASE_URL will continue to receive deprecation warnings even though this PR officially introduces it as the supported mechanism. Please remove OPENAI_BASE_URL from deprecatedEnvKeys (or update the deprecation logic) so the diagnostics match the new behavior.

colifran:
Thanks for the PR! The second issue was fixed upstream in deepagentsjs. The first issue was superseded by #328.

PR #40, opened 2026-07-02
feat: add Amazon Bedrock provider with API key (bearer token) auth

lyle-kargo:
## Summary

Adds **Amazon Bedrock** as a selectable inference provider, per the README's invitation to open a PR for new providers.

- Authenticates with a [Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (`AWS_BEARER_TOKEN_BEDROCK`) — the same single-env-var pattern as the existing providers, no SigV4 credential plumbing needed.
- Uses `ChatBedrockConverse` from `@langchain/aws`, which natively supports bearer-token auth.
- Region defaults to `us-east-1`, overridable via `AWS_REGION`.
- Preset models (Claude Haiku 4.5, Claude Sonnet 5, Claude Opus 4.8) use `global.` cross-region inference profile IDs so they work from any region with capacity; custom model IDs work as with other providers (the existing `isValidModelId` regex already accepts Bedrock's dotted/colon IDs).

## Changes

- `src/constants.ts` — `bedrock` provider entry, env-key constants, model presets
- `src/agent/index.ts` — `ChatBedrockConverse` branch in `createModel()`, debug env list
- `src/env.ts` — managed env keys + credential diagnostics for the new vars
- `src/cli.tsx` — provider validation message, secret redaction list
- `README.md` — Bedrock setup notes
- `package.json` — adds `@langchain/aws`

## Testing

- `pnpm run build` and `pnpm run lint:check` pass
- Verified `ChatBedrockConverse` constructs with a bearer token + region
- Exercised the interactive provider/credential setup flow with the new provider


adebiolley4G (approved):
As a follow-up, access to Bedrock models should also be enable without API key, using local credentials.

rajpratham1 (changes requested):
AWS_REGION is introduced as a supported configuration (BEDROCK_REGION_ENV_KEY) and documented in the README, but it is never surfaced by the environment diagnostics. Unlike the other provider-specific settings, users won't be able to see whether their region override was detected or where it came from, making troubleshooting harder. Please add createCredentialDiagnostic(BEDROCK_REGION_ENV_KEY, fileEnv) alongside the Bedrock API key so the diagnostics stay consistent with the documented configuration.

lyle-kargo:
Good catch @rajpratham1 — fixed in b83b028. `AWS_REGION` now appears in the credential diagnostics alongside the Bedrock API key. Since the region isn't a secret, it displays in plaintext like the provider/model entries rather than masked:

```
AWS_BEARER_TOKEN_BEDROCK | source: ~/.openwiki/.env | preview: "bedroc...PTE="
AWS_REGION | source: process.env | preview: "eu-central-1"
```

@adebiolley4G agreed on local-credential (SigV4) support — `ChatBedrockConverse` falls back to the default AWS credential chain natively, so it's a small change. Planning it as a follow-up PR alongside dynamic model listing to keep this one scoped to API-key auth.

lyle-kargo:
@rajpratham1 @adebiolley4G I noticed there is another PR doing roughly the same thing, should I close this?

SiddheshB-UBI:
Ran this against a real multi-turn tool-calling session (OpenWiki's own `--init` doc-generation flow, driving many `write_file`/`task` tool calls across turns against `ChatBedrockConverse` from `@langchain/aws`) and hit three issues worth fixing before this merges, independent of which credential-auth approach (IAM role/profile vs. bearer token) wins:

1. **Default `maxTokens` (4096) silently truncates large tool-call arguments.** OpenWiki's doc-generation loop makes `write_file` tool calls with full file contents as arguments — these routinely exceed 4096 tokens, and the truncated response either fails Zod schema validation (crashing the whole agent run, since `DynamicStructuredTool.call()` errors aren't caught by deepagents' tool wrapper) or silently drops file content. Needs an explicit `maxTokens` override in the `ChatBedrockConverse` constructor — we used `65536` against a Claude model with a 128000 max, and that fixed it.

2. **Default stream idle timeout (60s) is too short.** Hit `Bedrock Converse stream timed out after 60000 ms without receiving a chunk.` on a cold-ish response. `@langchain/aws`'s `stream_timeout.cjs` defaults `DEFAULT_STREAM_IDLE_TIMEOUT` to `6e4` — passing `streamIdleTimeout: 300000` (5 min) resolved it for us.

3. **Extended-thinking crash on multi-turn tool-call replay.** With a thinking-capable Claude model and thinking left enabled, replaying message history back to Bedrock Converse on a later turn throws:
   ```
   Error: 1 validation error detected: Value at 'messages.3.member.content.1.member.reasoningContent.reasoningText.text' failed to satisfy constraint: Member must not be null
   ```
   Reproduces with both `thinking.type: "enabled"` (`budget_tokens`) and `thinking.type: "adaptive"` (`output_config.effort`), and persists even with `disableStreaming: true`. We weren't able to root-cause this inside `@langchain/aws`'s history-replay/marshalling path — the only reliable workaround was disabling thinking entirely via `additionalModelRequestFields: { thinking: { type: "disabled" } }`. Since OpenWiki's agent loop is inherently multi-turn tool-calling, a Bedrock provider probably needs to either ship with thinking disabled by default or strip/repair `reasoningContent` blocks before replaying history, until this is properly root-caused (may be worth a separate issue against `@langchain/aws` itself).

Happy to help verify whichever of #96 / #40 moves forward once these are folded in.

lyle-kargo:
@SiddheshB-UBI thank you for running this against a real doc-generation session — all three are addressed in 45a5fd0:

1. **`maxTokens: 64000`** — set explicitly on the `ChatBedrockConverse` constructor so large `write_file` tool-call arguments no longer truncate at the SDK's 4096 default. Went with 64000 rather than 65536 since it's safely within every preset model's output ceiling (Haiku 4.5 caps at 64K).

2. **`streamIdleTimeout: 300000`** — 5-minute idle window per your finding, replacing the SDK's 60s default.

3. **Thinking disabled via `additionalModelRequestFields`** — with one refinement: the override is only sent for Claude models where thinking is *optional*. Models with always-on thinking (Sonnet 5 and later) reject an explicit `thinking: {type: "disabled"}` outright, so for those (and non-Claude model IDs) no override is sent. Agreed this wants a proper root-cause issue against `@langchain/aws`'s reasoningContent replay path — happy to file it.

Would definitely appreciate a re-run of your `--init` session against the updated branch if you have the cycles.

SiddheshB-UBI:
Re-ran our `--init`/`--update` doc-generation session against `45a5fd0` — thanks for the quick turnaround. Two of three hold up, one doesn't quite for the model this PR ships as its own preset:

1. **`maxTokens: 64000`** — confirmed fixed. Ran a full multi-turn session with several `write_file`-style tool calls and saw no truncation or schema-validation failures.
2. **`streamIdleTimeout: 300000`** — confirmed fixed. No idle-timeout errors across the run.
3. **Extended-thinking crash — not fixed for `global.anthropic.claude-sonnet-5`**, which is the exact Sonnet preset both this PR and #96 ship. `bedrockThinkingOverride()` skips sending the disable override for any model matching `/sonnet-5|opus-5|haiku-5|fable|mythos/`, on the assumption those always-on-thinking models reject an explicit `thinking: {type: "disabled"}`. Running against that preset unmodified reproduced our original crash verbatim on the second message:
   ```
   1 validation error detected: Value at 'messages.2.member.content.1.member.reasoningContent.reasoningText.text' failed to satisfy constraint: Member must not be null
   ```
   To isolate the cause, I forced the override through for Sonnet 5 anyway (bypassing the skip-list) and reran: Bedrock accepted `additionalModelRequestFields: { thinking: { type: "disabled" } }` without complaint, and the same doc-update session completed cleanly end-to-end with no crash. So at least for `global.anthropic.claude-sonnet-5` via Converse, the premise that it "rejects an explicit disable" doesn't hold up in practice — the skip is actively leaving the crash in place for the preset model most people will pick first.

Might be worth either dropping `sonnet-5` from the skip regex, or confirming against a real Bedrock account which models actually reject the override vs. which ones the assumption was carried over from without testing.

Happy to share the exact repro repo/session if useful.

(Note on methodology: to test this against our own AWS SSO/profile credentials rather than a Bedrock API key, I ran a locally-patched build that bypasses the `AWS_BEARER_TOKEN_BEDROCK` requirement and lets `ChatBedrockConverse` fall through to the default AWS credential chain — that bypass is local-only and unrelated to the three fixes under test here.)

SiddheshB-UBI:
@lyle-kargo Follow-up with a bit more evidence on the skip-list, in case it's useful before this merges.

I ran the same doc-update session across every model the skip regex (`/sonnet-5|opus-5|haiku-5|fable|mythos/`) actually references, checking whichever are real Bedrock models today:

| Model | Bedrock ID | Regex skips override? | Result |
|---|---|---|---|
| Sonnet 5 | `claude-sonnet-5` | Yes | Crashes with the `reasoningContent` error (as reported above); forcing the override through fixes it |
| Fable 5 | `claude-fable-5` | Yes | Couldn't test — this account gets `data retention mode 'default' is not available for this model` before ever reaching the Converse call, unrelated to this PR |
| Opus 4.8 | `claude-opus-4-8` | No (doesn't match `opus-5`) | Clean run — override sent, Bedrock accepts it, no crash |
| Haiku 4.5 | `claude-haiku-4-5-...` | No (doesn't match `haiku-5`) | Clean run — override sent, Bedrock accepts it, no crash |

Worth noting `opus-5` and `haiku-5` don't correspond to any model currently in Bedrock's Anthropic catalog — the real IDs are `opus-4-8` and `haiku-4-5`, neither of which matches those substrings, so those two regex branches are dead code today. `mythos` isn't in the catalog either.

So everywhere the code actually sends the disable override (Opus 4.8, Haiku 4.5), Bedrock accepts it fine and the run is clean. The only failure is where the code withholds it — Sonnet 5. That's consistent with the disable being safe to send universally, and the skip-list itself being the bug rather than a necessary accommodation for "always-on-thinking" models. Given I couldn't get a case where Bedrock actually rejected the override, I'd lean toward just dropping the skip-list and always sending `thinking: {type: "disabled"}` for Claude models, unless someone has a repro of a model that does reject it.

lyle-kargo:
@SiddheshB-UBI thanks for isolating that with the forced-override rerun — that's exactly the evidence needed. You're right: the skip-list premise was carried over from the first-party Anthropic API (where always-on-thinking models 400 on an explicit disable) without verifying Bedrock Converse behaves the same. Since you've confirmed `global.anthropic.claude-sonnet-5` accepts the disable via Converse and completes the session cleanly with it, 1345cb2 narrows the skip regex to Fable/Mythos-class models only — Sonnet 5 (and any other Claude model) now gets the thinking-disable override.

Also rebased onto latest main to pick up the engineering-hygiene refactor (#141): the Bedrock env keys now flow through the new `MANAGED_ENV_KEYS` single source of truth in `src/env.ts`, and `AWS_BEARER_TOKEN_BEDROCK` is added to the extracted `src/diagnostics.ts` redaction list. All 86 tests pass.

If you get a chance to rerun your session against `1345cb2` with the stock preset, that would close the loop.

SiddheshB-UBI:
@lyle-kargo Confirmed — reran the stock preset (`global.anthropic.claude-sonnet-5`, no overrides on my end) against `1345cb2`. Clean run end-to-end, no `reasoningContent` error, exit code 0. That closes the loop on all three original issues (`maxTokens`, `streamIdleTimeout`, thinking-disable). Thanks for the fast turnaround on this.

colifran:
thanks for the pr @SiddheshB-UBI! sorry for getting to this late. could you rebase your changes?

colifran:
closing in favor of https://github.com/langchain-ai/openwiki/pull/185

PR #48, opened 2026-07-02
feat: Make OpenWiki general purpose

bracesproul:


rajpratham1 (changes requested):
The provider selection order and default provider have changed from openrouter to openai (OPENAI_API_KEY → OPENROUTER_API_KEY → ANTHROPIC_API_KEY → ...). This is a behavior change that can affect existing users who have multiple provider credentials configured. Can you confirm this change is intentional and covered by tests? If it's intended, it should also be called out prominently in the migration notes/README since upgrading could silently switch users to a different provider.

Issue #49, opened 2026-07-03
Summarization offloads write conversation_history/ into the target repository


blue-az:
When the summarization middleware triggers in `deepagents`, it offloads the conversation history to a `conversation_history/` directory created in the target repository root.

This pollutes the workspace being documented: the directory can be swept into version control by accident, and subsequent OpenWiki update runs (or other agents working in the repo) can discover and read these history files, bloating context with stale transcript content.

Observed while running `--init`/`--update` cycles against a mid-size repository on the OpenRouter path with summarization triggering at a configured input-token cap.

**Suggested fixes** (either would resolve it):
1. Redirect the offload location outside the target repository — e.g. `~/.openwiki/<repo-hash>/conversation_history/` alongside the existing credential storage, or a temp path.
2. At minimum, add `conversation_history/` to the prompt-level discovery/read exclusions so documentation runs never inspect it.

Related: I'm submitting a small opt-in hardening PR for the OpenRouter path (early-summarization token cap, subagent/fallback kill switches) — this issue is the remaining workspace-pollution item discovered during that testing.

PrinceThummar011:
Hi! I'd like to take this issue. I'll trace how conversation_history/ is currently created, implement a clean fix that aligns with the project's architecture, add tests if needed, and open a PR once it's ready.  @christian-bromann

blue-az:
Hi @PrinceThummar011 — thanks for the nudge on this one! I actually filed #49 while wrapping up a related PR (#312) and deliberately kept them separate rather than scope-creep that PR. I'd been holding off circling back while the surrounding backend was going through the #48 rewrite, and it sat longer than I meant it to.

Your comment prompted me to finally take a pass — I've got a fix implemented and tested (redirects `/conversation_history` writes to `~/.openwiki/conversation_history` instead of the target repo, with backend changes + test coverage). Happy to open a PR with it, or if you'd rather run with your own approach, just say the word and I'll hold off. Didn't mean to step on your offer to take this one!

PrinceThummar011:
Thanks for the update! Please go ahead and open the PR since you've already implemented and tested it. I appreciate you checking in. I'll keep an eye out for another issue to contribute to and learn from your implementation.

colifran:
Closed with #504

PR #50, opened 2026-07-03
Harden OpenRouter path with opt-in transcript and routing controls

blue-az:
## Problem

On the OpenRouter path with constrained/budget models, documentation runs fail with HTTP 500 errors before summarization ever triggers: the summarization middleware's default trigger (~170k tokens) sits far above the payload size where OpenRouter requests actually start failing (roughly 270–340KB, ~12–25k tokens depending on the model). Two secondary issues compound this on longer runs: generated `openwiki.*` variant directories (from testing/validation runs) leak into discovery and get re-read, and subagent fan-out multiplies transcript growth.

## Changes

All hardening is **strictly opt-in** — with no environment variables set, behavior is unchanged for every provider:

- `OPENWIKI_OPENROUTER_MAX_INPUT_TOKENS` — when set, attaches a model profile with this `maxInputTokens` to OpenRouter model instances (via `Object.defineProperty`, so frozen/non-extensible instances don't crash), which lets the summarization middleware trigger early enough to stay under the payload limit. Suggested value for constrained models: `15000`.
- `OPENWIKI_DISABLE_SUBAGENTS=1` — disables subagent task delegation for runs where transcript growth is the binding constraint.
- `OPENWIKI_DISABLE_MODEL_FALLBACK=1` — disables OpenRouter fallback routing for deterministic single-model runs.
- Prompt discipline: discovery/read exclusions for `openwiki.*` variant directories so testing/validation copies of the wiki are never inspected or documented.
- Docs: the new options are documented in `openwiki/operations/credentials-and-updates.md`.

## Verification

- `pnpm run build`, `pnpm run lint:check`, and `pnpm run format:check` all pass.
- Live-validated end-to-end on the OpenRouter path with the cap set: summarization prunes at the configured threshold and full `--init`/`--update` runs complete with no 500s (previously reproducible failures).

## Related

The remaining workspace-pollution item found during this testing is filed separately as #49 (`conversation_history/` offloads written into the target repository).

### DIFF
--- a/openwiki/operations/credentials-and-updates.md
+++ b/openwiki/operations/credentials-and-updates.md
@@ -20,6 +20,10 @@ The file stores provider configuration and API keys:
 - `OPENWIKI_MODEL_ID` — the default model ID
 - Provider API keys: `OPENROUTER_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `BASETEN_API_KEY`, `FIREWORKS_API_KEY`
 - Optional LangSmith settings: `LANGSMITH_API_KEY`, `LANGCHAIN_PROJECT`, `LANGCHAIN_TRACING_V2`
+- Run-hardening options:
+  - `OPENWIKI_DISABLE_MODEL_FALLBACK` — set to `1` to disable OpenRouter fallback routing
+  - `OPENWIKI_DISABLE_SUBAGENTS` — set to `1` to disable subagent task delegation entirely
+  - `OPENWIKI_OPENROUTER_MAX_INPUT_TOKENS` — set a custom input token cap (suggested: `15000` for constrained models) for OpenRouter models to force early summarization and prevent HTTP 500 payload errors on large transcripts
 
 The loader merges those values into `process.env`, while preferring existing process-level values over file values. Deprecated keys (`OPENAI_BASE_URL`, `OPENAI_ORG_ID`, `OPENAI_PROJECT`) are skipped on load and removed on save.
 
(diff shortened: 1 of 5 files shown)


rajpratham1 (changes requested):
configureDeepAgentHarness() only calls registerHarnessProfile() when OPENWIKI_DISABLE_SUBAGENTS === "1", but there are no corresponding tests verifying that the task tool is actually disabled. Since this changes runtime behavior by removing subagent delegation, please add an integration or unit test that confirms the task tool is excluded and generalPurposeSubagent.enabled is false when the environment variable is set, while leaving the default behavior unchanged. This helps prevent future regressions.

blue-az:
Good call — added `node:test` coverage in 6e17267 (no new deps; `tsx` was already dev-installed). The disabled profile now has an asserted shape (`task` excluded, `generalPurposeSubagent.enabled` false), and the middleware test invokes `wrapModelCall` to confirm `task` is filtered out of the model call while default behavior stays untouched (no middleware, no tools removed when the flag is unset). I also pulled the tool name into one shared constant so the profile and middleware can't drift out of sync. Thanks for the review.

blue-az:
Pushed the requested test coverage in 6e17267 — could you re-review when you get a chance?

blue-az:
Closing in favor of #312 — main has moved far enough (connector tools, output modes, ChatGPT OAuth) that the same three opt-in controls needed a clean redo rather than a conflict resolution. Thanks for the review here; the test-coverage ask carried over.

PR #53, opened 2026-07-03
feat: add DeepSeek as a supported provider

poliakarmai:
## Summary

Adds DeepSeek as a first-class provider in OpenWiki. DeepSeek offers competitive pricing and strong coding capabilities through an OpenAI-compatible API, making it a natural fit for documentation generation.

## Changes

- `src/constants.ts` — +12 lines
  - Added `DEEPSEEK_API_KEY` env var constant
  - Added `"deepseek"` to `OpenWikiProvider` union type
  - Added `"deepseek"` to `SELECTABLE_OPENWIKI_PROVIDERS`
  - Added provider config with base URL `https://api.deepseek.com/v1`, API key env var, and model options (`deepseek-chat`, `deepseek-reasoner`)

## How it works

DeepSeek provides an OpenAI-compatible API, so it works through the existing `ChatOpenAI` codepath — no model creation changes needed. The provider config only sets the base URL and API key reference.

## Usage

```bash
# Via environment variables
OPENWIKI_PROVIDER=deepseek DEEPSEEK_API_KEY=sk-... openwiki --init

# Or set in ~/.openwiki/.env
OPENWIKI_PROVIDER=deepseek
DEEPSEEK_API_KEY=sk-...
```

## Note

DeepSeek's API doesn't support the `file` content type used by some LangChain tool messages. This is a LangChain-level compatibility issue tracked separately — it affects all OpenAI-compatible providers that lack `file` type support, not just DeepSeek. In practice, this may cause issues with `deepagents` file operations and should be addressed in a follow-up.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -4,6 +4,7 @@ export const BASETEN_API_KEY_ENV_KEY = "BASETEN_API_KEY";
 export const FIREWORKS_API_KEY_ENV_KEY = "FIREWORKS_API_KEY";
 export const OPENAI_API_KEY_ENV_KEY = "OPENAI_API_KEY";
 export const ANTHROPIC_API_KEY_ENV_KEY = "ANTHROPIC_API_KEY";
+export const DEEPSEEK_API_KEY_ENV_KEY = "DEEPSEEK_API_KEY";
 export const OPENROUTER_API_KEY_ENV_KEY = "OPENROUTER_API_KEY";
 export const OPENWIKI_PROVIDER_ENV_KEY = "OPENWIKI_PROVIDER";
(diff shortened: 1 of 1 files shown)


rajpratham1 (changes requested):
This change only updates constants.ts by adding the deepseek provider configuration, but I don't see the corresponding runtime integration. Unless another commit adds it, createModel() still won't know how to instantiate a DeepSeek client, env.ts won't surface DEEPSEEK_API_KEY, and the CLI/README won't expose the new provider. As-is, users can select deepseek, but it won't function end-to-end. Please update the remaining provider integration points or split this into a preparatory refactor if that's the intent.

bracesproul:

PR #55, opened 2026-07-03
feat: Add Anthropic on Azure AI Foundry provider (with HTTP proxy support)

### LINKED ISSUE #33: Anthropic 400 INVALID_TOOL_RESULTS: files with unmapped/no extension (Makefile, Dockerfile, uv.lock) are sent as base64 documents
## Summary

`openwiki --init` / `--update` reliably crashes mid-run with Anthropic provider on any real-world repo:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.56.content.2.tool_result.content.0.document.source.base64.media_type: Input should be 'application/pdf'"}}

```

## Root cause

The read-tool MIME detection in the bundled `deepagents` dist (`langsmith-*.js` / `langsmith-*.cjs`) is extension-map based:

```js
function getMimeType(filePath) {
}
function isTextMimeType(mimeType) {
    return mimeType.startsWith("text/") || mimeType === "application/json" || mimeType === "application/javascript" || mimeType === "image/svg+xml";
}
```

Two problems:

1. **Any file whose extension is not in `MIME_TYPES` falls back to `application/octet-stream`** and is treated as binary → base64 `document` block. This includes perfectly ordinary text files every repo has: `Makefile`, `Dockerfile`, `Caddyfile`, `LICENSE` (Node `extname()` returns `""` for extensionless files — the `.makefile`/`.dockerfile` map entries never match them), plus `uv.lock`, `*.mermaid`, `*.mako`, `.env.example`, etc. So the agent crashes as soon as it reads a Makefile.
2. **Actual binary files (.png, .xlsx, .docx) are also sent as `document` blocks**, but the Anthropic Messages API only accepts `application/pdf` as a document media type → same 400.

The OpenAI provider hits the equivalent wall (`Error 400 Invalid value: 'file'`) — likely the same root cause as #21.

Because the whole agent run dies on the first such file, `--init` cannot complete on most non-trivial repos.

## Repro

Python 3.13 / FastAPI monorepo (~1500 files) containing `Makefile`, `docker/*.Dockerfile`, `uv.lock`, `docs/*.xlsx`, `docs/*.png`, `static/media/*.png`:

```
npm install -g openwiki   # 0.0.1
openwiki --init           # provider: anthropic, model: claude-sonnet-5
# → crashes with the 400 above, wiki partially generated
```

## Workaround we used

Patch the fallback in the installed bundle so unmapped extensions are treated as text:

```bash
B="$(npm root -g)/openwiki/node_modules/deepagents/dist"
sed -i.bak 's#(filePath).toLocaleLowerCase()] || "application/octet-stream"#(filePath).toLocaleLowerCase()] || "text/plain"#' "$B"/langsmith-*.js "$B"/langsmith-*.cjs
```

…and move real binaries (`*.png`, `*.xlsx`, `.coverage`) out of the tree for the duration of the run. After that `--init`/`--update` complete fine.

## Suggested fix

1. Fall back to content sniffing (or `text/plain`) for unknown extensions instead of `application/octet-stream`; match extensionless well-known names (`Makefile`, `Dockerfile`, `LICENSE`) as text.
2. For genuinely binary files with media types the provider cannot accept, return a tool **error string** ("binary file skipped") instead of a base64 document block — one unreadable file should not kill the whole documentation run.

Environment: openwiki 0.0.1, node v22.19.0, macOS (darwin 25.5.0), provider anthropic / claude-sonnet-5.

uesleilima:
## Summary

Adds **Anthropic on Azure AI Foundry** as an inference provider (id: `anthropic-foundry`), plus global outbound **HTTP proxy** support that was required to make it work in a corporate environment.

> **Update (rebased on `main`):** This branch has been merged up to the latest `upstream/main`. The Anthropic **request sanitizer** that this PR originally shipped has been **removed** — the underlying media-type crash was fixed upstream in [deepagentsjs#656](https://github.com/langchain-ai/deepagentsjs/pull/656) (unknown/extension-less files now default to `text/plain` instead of being sent as base64 `document` blocks), which landed via [openwiki#221](https://github.com/langchain-ai/openwiki/pull/221) and is now pinned here through `deepagents@^1.10.8`. `main` carries no sanitizer of its own, so keeping ours would have been redundant. See the ["Removed: request sanitizer"](#removed-request-sanitizer) note below.

### 1. New provider: `anthropic-foundry`

Azure AI Foundry serves Claude models through an Anthropic-_compatible_ endpoint
(`POST https://<resource>.services.ai.azure.com/anthropic/v1/messages`, using the
`x-api-key` + `anthropic-version` headers). Because the wire protocol is Anthropic's,
the cleanest integration is to **reuse `ChatAnthropic`** with a custom base URL rather
than adding a new client type. The Anthropic SDK appends `/v1/messages`, so the
configured base URL is `https://<resource>.services.ai.azure.com/anthropic`, passed via
`anthropicApiUrl`.

- Added to the `OpenWikiProvider` union, `SELECTABLE_OPENWIKI_PROVIDERS`, and
  `PROVIDER_CONFIGS`, and a `createModel()` branch in `src/agent/index.ts`.
- Introduced an optional `baseUrlEnvKey` field on `ProviderConfig` because, unlike the
  other providers whose base URLs are fixed/hardcoded, a Foundry endpoint is
  **per-resource** and must come from the environment.

**Why the id is `anthropic-foundry` (not `azurefoundry`):** Azure AI Foundry hosts many
model families (OpenAI, etc.). We only support its _Anthropic_ surface here, so the id
reflects that scope. The display label is "Anthropic on Azure AI Foundry" for the same
reason. The hyphenated id matches other multi-word providers such as `openai-compatible`
and `openai-chatgpt`.

**Why these exact env var names** — `ANTHROPIC_FOUNDRY_API_KEY`,
`ANTHROPIC_FOUNDRY_BASE_URL`, `ANTHROPIC_FOUNDRY_RESOURCE`: these are the **official
environment variables used by the Foundry-aware Anthropic clients and Claude Code**
(per Anthropic's Foundry docs and MS Learn). Reusing the same names means anyone who
already configured Foundry works out of the box, with no new convention to learn. We
support `ANTHROPIC_FOUNDRY_RESOURCE` as an alternative to the base URL and build
`https://<resource>.services.ai.azure.com/anthropic` from it, mirroring the official
SDK behavior. The model id for this provider is the Foundry **deployment name**; it
shares the same preset model list as the `anthropic` provider since they're the same
models.

Onboarding (`src/credentials.tsx`) gained a base-URL step shown only for this provider
(accepts either a full base URL or a bare resource name), and `src/env.ts` persists /
diagnoses the new keys.

### 2. Global HTTP proxy support

The Foundry endpoint in our environment is only reachable through a corporate web proxy.
**Node's built-in `fetch` does not honor `HTTP(S)_PROXY` env vars**, and the repo had no
proxy handling, so requests would silently bypass the proxy and fail. We now read a proxy
URL (precedence: `OPENWIKI_PROXY` > `HTTPS_PROXY` > `HTTP_PROXY` > `ALL_PROXY`, plus
lowercase variants) and install a global `undici` `ProxyAgent` dispatcher.

- **Why `undici`:** it's the implementation behind Node's global `fetch`;
  `setGlobalDispatcher(new ProxyAgent(url))` is the supported way to route `fetch`
  through a proxy. It's added as an explicit dependency because the module isn't
  importable from Node core.
- **Why global (all providers):** proxy reachability is an environment property, not a
  provider one — if you're behind the proxy, every outbound call needs it.
- A bare `host:port` (like the one in our internal proxy) is normalized to
  `http://host:port`.

### <a name="removed-request-sanitizer"></a>Removed: request sanitizer

This PR originally added an outbound `fetch` wrapper that rewrote Anthropic `document`
blocks to work around:

> `messages.N.content.*.tool_result.content.0.document.source.base64.media_type: Input should be 'application/pdf'`

That 400 (tracked in [#33](https://github.com/langchain-ai/openwiki/issues/33)) was
caused by deepagents' `read_file` mapping unknown/extension-less files (e.g. `Makefile`,
`Dockerfile`, `uv.lock`, `.tf`) to `application/octet-stream` and sending them as base64
`document` blocks. It is now **fixed upstream** in
[deepagentsjs#656](https://github.com/langchain-ai/deepagentsjs/pull/656) (unknown
extensions default to `text/plain`), shipped via
[openwiki#221](https://github.com/langchain-ai/openwiki/pull/221) and pinned here through
`deepagents@^1.10.8` after merging `main`. Since `main` relies solely on that fix and
carries no sanitizer, the custom implementation has been removed to stay aligned with
upstream.

_Residual edge case:_ genuinely binary files with mapped office extensions (`.ppt`,
`.pptx`) can still reach the API as non-PDF documents; upstream accepts this and closed
#33 on the deepagents fix alone.

## Files

- `src/constants.ts` — provider registration, env-key constants, base-URL/proxy resolution helpers.
- `src/agent/index.ts` — `createModel` branch for `anthropic-foundry` + global proxy install.
- `src/env.ts` — managed keys + credential diagnostics for the new config.
- `src/credentials.tsx` — onboarding base-URL step.
- `src/cli.tsx` — provider error string + diagnostic redaction of the new API key.
- `package.json` / `pnpm-lock.yaml` — `undici` dependency; `deepagents` bumped to `^1.10.8` via the `main` merge.
- `README.md`, `openwiki/quickstart.md`, `openwiki/operations/credentials-and-updates.md` — docs.

## Test plan

- `tsc` build, `eslint`, and `prettier --check` pass.
- Full test suite passes (228/228).
- `ChatAnthropic` receives the Foundry base URL via `anthropicApiUrl`.
- Helper unit checks: base-URL normalization (strips `/v1/messages`), resource→URL derivation, proxy precedence + bare-host normalization.
- Manual end-to-end run against a target repo through the corporate proxy (init/update) succeeds; the previously-failing document request now passes on `deepagents@1.10.8` without any sanitizer.


rajpratham1 (approved):
This is a substantial feature that is internally consistent: it adds a new anthropicfoundry provider, endpoint configuration, proxy support, credential flow, request sanitization, and documentation.

What looks good
✅ Provider is integrated across runtime, CLI, credentials, env handling, and docs.
✅ Request sanitization for Anthropic document uploads addresses a known compatibility issue.
✅ Proxy support is useful for enterprise deployments.
✅ Good documentation and onboarding updates.
✅ Reuses Anthropic model configuration instead of duplicating it.
One thing I'd request before merging

The proxy installation is global:

setGlobalDispatcher(new ProxyAgent(proxyUrl));

Once installed it cannot be changed during the process lifetime because of:

let globalProxyInstalled = false;

If OPENWIKI_PROXY changes (or multiple OpenWiki runs occur with different proxy settings in the same process), the dispatcher won't update.

Consider either:

documenting that the proxy is fixed for the process lifetime, or
reinstalling the dispatcher when the configured proxy changes.
Verdict

Approve. The feature is well integrated and the remaining concern is a minor lifecycle/design consideration rather than a blocker.

uesleilima:
hi @rajpratham1 I had a few merge conflicts to address, could you review it again? thanks!

chrisjohnson00:
```
| Error                                                                                                                                                                                                                                  
│ ANTHROPIC_FOUNDRY_API_KEY is required.
```

My company does not use foundry api keys, it is all Azure RBAC based.  Would love to see that supported here too - if possible in this PR.

uesleilima:
Closing this as it clearly didn't get much traction with the maintainer.

PR #57, opened 2026-07-03
fix(update): skip agent run when no repository changes are detected

### LINKED ISSUE #44: --update with zero diff runs a full agent pass instead of a no-op
## Summary

`openwiki --update` on a repo with **no commits and no working-tree changes** since the last successful update still launches the full DeepAgents tool loop and can take significantly longer than an update with a real diff present — in our test, ~6x longer (759s vs 117s) — despite having strictly less work to do. The run isn't wasted (it found and fixed a genuine content gap / hallucination on its own initiative), but there's no way to get a cheap "nothing to do, exiting" result, and no bound on how much work the agent decides to do on a technically-unnecessary run.

## Repro

```sh
openwiki --update -p "Update the documentation"   # run 1: real diff present -> 117s, correctly
                                                  #   scoped to 1 changed file
openwiki --update -p "Update the documentation"   # run 2: zero new commits, zero working-tree
                                                  #   changes -> 759s, rewrote both existing
                                                  #   pages + patched CLAUDE.md
```

`.last-update.json`'s `gitHead` was correctly unchanged between the two snapshots (confirming git-level diff detection itself works) — the issue is that "no diff" never short-circuits the run.

## Root cause (from source, openwiki 0.0.1)

- `dist/agent/utils.js:131-171` (`createGitSummary`) computes `git log <lastGitHead>..HEAD` and `git diff --name-status HEAD`; on a zero-diff repo both are empty. This is the **only** place "nothing changed" is computed anywhere in the codebase — and it's only ever assembled into a prompt string (`context.gitSummary`), never compared/branched on in code.
- `dist/agent/index.js:70-141` (`runOpenWikiAgentCore`) always proceeds to create the model, open the agent, and start the full `agent.stream(...)` tool loop, regardless of what `createGitSummary` found.
- The only "this might be a no-op" logic anywhere is prompt text (`dist/agent/prompt.js:166`): *"Updates may be a no-op. If there are no relevant ... changes ... do not edit files."* — advisory, not enforced. The model is free to (and in our test, did) decide the existing wiki has other issues worth fixing and proceed to a full self-directed quality pass, with no turn/token/cost cap distinct from prompt guidance.
- There is a `--dry-run` flag in the parser (`dist/commands.js:16-26`) but it's gated behind `isDevelopmentMode()` (`NODE_ENV=development` or `OPENWIKI_DEV=1`), hidden from public `--help`, and even when enabled only prints a static execution-plan summary — it doesn't run `createGitSummary` or report whether an update would be a no-op.

## Suggested fix

1. Promote a real no-op short-circuit into code: if `createGitSummary`'s `git log`/`git diff` evidence for an `update` command is empty (and `openwiki/` content hash — see `createOpenWikiContentSnapshot`, `utils.js:41-46` — matches the last recorded state), skip invoking the agent entirely and print "documentation already up to date" — no model call at all.
2. Alternatively/additionally, surface `--dry-run` (or a new `--check`) as a public, non-dev flag that runs `createGitSummary` and reports whether an update has anything to do, without opening the agent stream — so callers (cron/CI) can probe cheaply before paying for a full run.
3. If a full audit pass on a zero-diff repo is intentionally allowed (as a "periodic quality check" feature), it should be opt-in and/or budget-capped, not the unconditional default behavior of a plain `--update` invocation.

## Economics note

No token/cost telemetry is printed by the CLI at any point, so this cost can currently only be estimated via wall-clock time or the provider dashboard — worth keeping in mind alongside a fix here, since a nightly `--update` cron across N repos multiplies an already-invisible cost.

## Environment

openwiki 0.0.1, npm global install, macOS (darwin 25.3.0), provider OpenRouter, model `z-ai/glm-5.2`.

zan22ye:
## Summary

- Add an update no-op preflight before provider/model/agent initialization.
- Skip `openwiki --update` when the last recorded OpenWiki `gitHead` matches the current HEAD and the worktree has no meaningful changes.
- Add regression tests for no-op detection and dirty worktree behavior.

Fixes #44.

## Testing

- `./node_modules/.bin/tsx --test test/**/*.test.ts`
- `./node_modules/.bin/tsc -p tsconfig.json`
- `./node_modules/.bin/eslint .`
- `./node_modules/.bin/prettier --check package.json src/agent/index.ts src/agent/types.ts src/agent/utils.ts test/update-noop.test.ts`
- `git diff --check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "prebuild": "pnpm run clean",
     "prepack": "pnpm run build",
     "start": "node dist/cli.js",
-    "test": "tsx --test test/**/*.test.ts"
+    "test": "vitest run"
   },
   "dependencies": {
     "@langchain/anthropic": "^1.5.1",
@@ -58,7 +58,8 @@
     "prettier": "^3.8.4",
     "tsx": "^4.19.2",
     "typescript": "^5.7.2",
-    "typescript-eslint": "^8.62.0"
+    "typescript-eslint": "^8.62.0",
+    "vitest": "^4.1.9"
   },
   "pnpm": {
     "onlyBuiltDependencies": [
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "lint:check": "eslint .",
     "prebuild": "pnpm run clean",
     "prepack": "pnpm run build",
-    "start": "node dist/cli.js"
+    "start": "node dist/cli.js",
+    "test": "tsx --test test/**/*.test.ts"
   },
   "dependencies": {
     "@langchain/anthropic": "^1.5.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -63,6 +63,9 @@ importers:
       typescript-eslint:
         specifier: ^8.62.0
         version: 8.62.0(eslint@10.5.0)(typescript@5.9.3)
+      vitest:
+        specifier: ^4.1.9
+        version: 4.1.9(@types/node@22.19.21)(vite@8.1.3(@types/node@22.19.21)(esbuild@0.28.1)(tsx@4.22.4)(yaml@2.9.0))
 
 packages:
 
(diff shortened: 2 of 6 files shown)


rajpratham1 (approved):
This is a useful optimization with appropriate tests.

Pros
✅ Avoids unnecessary LLM/agent runs when nothing changed.
✅ Checks both:
Git HEAD unchanged.
Working tree clean (excluding .last-update.json).
✅ Returns a clear skipped flag.
✅ Emits informative debug/event messages.
✅ Includes tests for both skip and non-skip cases.

Currently only .last-update.json is ignored:

isUpdateMetadataStatusLine(...)

If future OpenWiki metadata files are added (cache, timestamps, diagnostics, etc.), this helper will need updating. It may be more maintainable to ignore the entire openwiki/.last-update.json via a configurable allowlist or metadata-path helper.

Verdict

Approve. The optimization is safe, improves performance, and is covered by tests. The only suggestion is future-proofing the metadata-ignore logic.

bracesproul (approved):
thank you! i pushed a commit updating the test to use vitest, and one small logic fix in the update checker. the last commit might be an openwiki docs update, so it checks that the last commit didnt just update openwiki files

bracesproul:
Do you have a twitter I can mention if I tweet about this improvement?

zan22ye:
> Do you have a twitter I can mention if I tweet about this improvement?

Thanks so much! I really appreciate the review and the follow-up improvements. You can mention me at @hudushy if you tweet about this improvement — https://x.com/hudushy

PR #58, opened 2026-07-03
feat: add DeepSeek as a first-class provider

gariwat:
## Summary

DeepSeek exposes an OpenAI-compatible API at `https://api.deepseek.com/v1`. This PR adds it as a first-class provider in OpenWiki, alongside the existing `openai`, `anthropic`, `baseten`, `fireworks`, and `openrouter` providers.

It also patches a DeepSeek-API limitation: the chat completions endpoint only accepts `text` content blocks and rejects every other variant (file, image, …) with a 400. The deepagents agent loop emits non-text blocks on subsequent turns, which made `--init`/`--update` unusable against DeepSeek. A targeted, DeepSeek-only fetch rewriter flattens array content into a single text string before sending.

## Changes

| File | Change |
|---|---|
| `src/constants.ts` | Add `DEEPSEEK_API_KEY_ENV_KEY` + `DEEPSEEK_BASE_URL`; extend `OpenWikiProvider` and `SELECTABLE_OPENWIKI_PROVIDERS` with `"deepseek"`; add `deepseek` entry to `PROVIDER_CONFIGS` exposing `deepseek-chat` (V3) and `deepseek-reasoner` (R1). |
| `src/env.ts` | Add `DEEPSEEK_API_KEY` to `managedEnvKeys` and the credential diagnostics report. |
| `src/agent/index.ts` | (1) Add an explicit `provider === "deepseek"` branch in `createModel` that uses `ChatOpenAI` with `DEEPSEEK_BASE_URL` + `DEEPSEEK_API_KEY`. (2) Install a one-time global fetch rewriter for `api.deepseek.com` requests that flattens array content into text so DeepSeek's spec-compliant endpoint accepts the payload. |

98 lines added across 2 commits, no deletions. `dist/` and `node_modules/` are gitignored.

## Why an explicit `provider === "deepseek"` branch?

The generic `ChatOpenAI` fallthrough in `createModel` would handle `deepseek` automatically (it reads `providerConfig.baseURL` and `getProviderApiKeyEnvKey(provider)`). I added an explicit branch anyway because:
1. It makes intent self-documenting — future readers see DeepSeek is a first-class target, not a config-only afterthought.
2. It isolates future provider-specific tweaks (headers, model defaults, response shaping) from the OpenAI path.

## Operator setup

No secrets in this PR. To run OpenWiki against DeepSeek:

`~/.openwiki/.env`:
```env
OPENWIKI_PROVIDER="deepseek"
OPENWIKI_MODEL_ID="deepseek-chat"
DEEPSEEK_API_KEY="<your deepseek key>"
```

Or in shell:
```bash
export OPENWIKI_PROVIDER=deepseek
export OPENWIKI_MODEL_ID=deepseek-chat
export DEEPSEEK_API_KEY='<your deepseek key>'
openwiki --init
```

Default model is `deepseek-chat`. Set `OPENWIKI_MODEL_ID=deepseek-reasoner` to switch to R1. LangSmith tracing remains optional via `LANGSMITH_API_KEY`.

## Verification

- `pnpm install --frozen-lockfile` ✅
- `pnpm run build` (tsc, strict) ✅
- `pnpm run lint:check` ✅
- `pnpm run format:check` ✅
- `--print` smoke test against `deepseek-chat` via the DeepSeek API using `DEEPSEEK_API_KEY` from env: ✅
- Full `--init` end-to-end against `langchain-ai/openwiki#55` (a small TypeScript repo): agent loop completed without `400 unknown variant 'file'`, generated 8 markdown pages + AGENTS.md. Key sourced from env, not committed.

## Notes for reviewer

- `OPENAI_BASE_URL` is in `deprecatedEnvKeys` and is intentionally skipped on env load — the existing code path does not support overriding `openai` base URL via env. Adding a first-class provider is the supported way to add a new OpenAI-compatible endpoint.
- I did not touch any existing provider config, credential handling, or onboarding UI flow. `SELECTABLE_OPENWIKI_PROVIDERS` and `getProviderModelOptions` are data-driven, so `src/credentials.tsx` automatically surfaces the new provider and its two models with no further edits.
- The fetch rewriter is gated on `Symbol.for("openwiki.deepseekFetchRewriterInstalled")` and only rewrites requests whose URL contains `api.deepseek.com`. Other providers see the original `globalThis.fetch` and are unaffected.
- The rewriter intentionally drops non-text content (file, image, etc.) by replacing each block with `[omitted <type> block]`. DeepSeek cannot consume those blocks anyway; this avoids the 400. If we ever need to forward tool/file payloads to DeepSeek, that would be a follow-up.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -387,6 +387,7 @@ async function createModel(provider: OpenWikiProvider, modelId: string) {
   }
 
   if (provider === "deepseek") {
+    installDeepSeekFetchRewriter();
     return new ChatOpenAI({
       apiKey: process.env[DEEPSEEK_API_KEY_ENV_KEY],
       configuration: {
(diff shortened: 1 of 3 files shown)


rajpratham1 (changes requested):
The provider integration is mostly complete, but I wouldn't approve it yet because a major feature is added without any tests.

What's good
✅ Adds deepseek as a first-class provider.
✅ Registers API key, provider config, and models.
✅ Updates environment diagnostics.
✅ Adds a fetch rewriter to convert multipart content into text, matching DeepSeek's API limitations.
Blocking issue

No tests are included.

This PR introduces ~100 lines of runtime logic, including:

global fetch monkey-patching,
JSON request rewriting,
provider-specific model creation,
message transformation logic.

None of these behaviors are covered.

At minimum I'd expect tests for:

DeepSeek provider selection.
rewriteMessageContentForDeepSeek().
Non-text blocks becoming [omitted ... block].
Text blocks preserved.
Fetch interceptor only affecting api.deepseek.com.
Non-DeepSeek providers remaining unchanged.
Minor observations
The global fetch patch is intentionally installed once via Symbol.for, which is reasonable, but tests should verify it doesn't double-wrap fetch.
Consider returning a cleanup function similar to the OpenRouter debug interceptor to improve testability.

bracesproul:

PR #60, opened 2026-07-03
feat: support excluding folders from documentation via .openwikiignore

mhassan1219:
## Summary

Adds a way to hide or skip folders from documentation. Repositories can create a `.openwikiignore` file at the repo root listing paths or glob patterns — one pattern per line, blank lines and `#` comments ignored. Negation patterns (`!`) are not supported and are rejected with a clear error rather than misread as exclusions:

```
# Internal tooling that should not be documented
scripts/internal
vendor/
**/legacy/**
```

The CLI reads the file when building run context and injects an "Exclusions discipline" section into the agent's system prompt, alongside the existing run/subagent/git discipline sections. The agent is instructed to:

- never `ls` / `read` / `glob` / `grep` excluded paths, run shell commands against them, or send subagents into them
- never document excluded paths or mention their contents, structure, or purpose in `openwiki/` or top-level agent instruction files
- ignore excluded paths when they show up in git change summaries — they never require documentation updates
- remove existing wiki content covering newly excluded paths during `--update` runs
- document surrounding behavior without describing an excluded path if one seems essential to explain the repo

## Why a file instead of a CLI flag

Exclusions need to persist across `--update` runs and the scheduled GitHub Action (`examples/openwiki-update.yml`); a repo-committed file gives every future run (and every contributor) the same exclusions with no extra CLI surface. The naming follows the `.gitignore` / `.dockerignore` convention.

## Implementation notes

- `readIgnoredPaths` in `src/agent/utils.ts` validates the file loudly instead of silently dropping entries: negation patterns, control characters, lines over 200 chars, and more than 200 patterns each fail with an error naming the offending line and the fix. An owner's exclusion list is honored in full or the run stops — it is never partially applied.
- Entries are rendered JSON-quoted in the prompt so they read as data rather than instruction text.
- Exclusions flow through `RunContext` and apply to init, update, and chat runs (so chat answers don't leak hidden folders either).
- With no `.openwikiignore` present, the system prompt is byte-identical to before.
- `--debug` reports `ignore.patterns=<count>`.

## Testing

- `pnpm run build`, `pnpm run lint:check`, and prettier all pass.
- Verified against the built `dist/` output: parsing (comments, blanks, dedupe, trimming, missing file), all four rejection paths (negation, over-long line, control characters, pattern-count cap), and prompt rendering (quoted entries and git-summary rule present with patterns; absent and byte-identical without).

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,19 @@ On the first interactive run, OpenWiki will have you configure your inference pr
 
 These configuration options and secrets will be saved to `~/.openwiki/.env` on your local machine.
 
+## Excluding Folders From Documentation
+
+To keep parts of your repository out of the wiki, add a `.openwikiignore` file to the repository root. Each line is a path or glob pattern relative to the repository root. Blank lines and lines starting with `#` are ignored. Negation patterns (lines starting with `!`) are not supported.
+
+```
+# Internal tooling that should not be documented
+scripts/internal
+vendor/
+**/legacy/**
+```
+
+OpenWiki will not explore or document excluded paths, and update runs remove existing wiki content that covers newly excluded paths.
+
 ## Customizing
 
 OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -159,6 +159,7 @@ async function runOpenWikiAgentCore(
 ): Promise<OpenWikiRunResult> {
   const context = await createRunContext(command, cwd);
   emitDebug(options, "context=created");
+  emitDebug(options, `ignore.patterns=${context.ignoredPaths.length}`);
   const openWikiSnapshotBefore =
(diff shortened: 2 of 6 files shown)


rajpratham1 (changes requested):
The idea is valuable, but the implementation only changes the prompt. It does not actually prevent the agent from reading excluded files.

Blocking issue

The PR promises:

"OpenWiki will not explore or document excluded paths"

However, it only:

reads .openwikiignore,
passes patterns into the system prompt,
instructs the LLM not to inspect those paths.

The agent can still:

call tools on ignored directories,
grep/search ignored files,
read them accidentally,
have subagents inspect them.

This is policy-by-prompt rather than enforcement.

If exclusion is a feature, the filesystem/tool layer should honor it.

Missing tests

There are no tests for:

parsing .openwikiignore
duplicate removal
comments/blank lines
invalid ! patterns
overlong patterns
missing file
prompt generation with ignored paths

Given ~70 lines of parsing logic, unit tests are expected.

Minor observations
✅ Deduplicating patterns is nice.
✅ Good validation/error messages.
✅ Reasonable limits (200 patterns, 200 chars).
✅ README is clear.

colifran:
Hi! Thanks for your work on this. Adding support for a .openwikiignore is something we're planning to handle with the internal team.

PR #66, opened 2026-07-03
fix: sanitize non-PDF binary file blocks before they reach Anthropic

### LINKED ISSUE #65: read_file crashes with Anthropic 400 on extensionless files (Dockerfile, LICENSE, CHANGELOG, ...)
## Description

`openwiki --init` / `--update` crashes with an Anthropic API 400 error whenever
the agent (or the auto-added general-purpose subagent) reads certain repo files
via deepagents' `read_file` tool:

    Error 400 {"type":"error","error":{"type":"invalid_request_error",
    "message":"messages.N.content.M.tool_result.content.0.document.source.base64.media_type:
    Input should be 'application/pdf'"}}

## Root cause

deepagents' `read_file` tool determines MIME type purely from file extension
(the `MIME_TYPES` lookup in the memory backend utils), defaulting anything
unrecognized to `application/octet-stream`. Files with **no extension at all**
— `Dockerfile`, `LICENSE`, `CHANGELOG`, `CODEOWNERS`, etc. — fall into this
default and get returned as a generic:

    { type: "file", mimeType: "application/octet-stream", data: <base64> }

content block. When forwarded to Anthropic, this gets serialized as a
"document" content block, but Anthropic's API only accepts
`application/pdf` for base64 document blocks — so any non-PDF binary `file`
block causes the whole run to fail.

The same class of bug also affects `{ type: "image"|"audio"|"video" }` blocks
read_file returns for other binary files: Anthropic only supports a handful
of raster image formats (not HEIC/HEIF, which read_file happily returns as
`image/heic`), and has no audio or video content block type at all.

## Repro

1. `openwiki --init` in any repo containing an extensionless file
   (e.g. a repo with a `Dockerfile`, `LICENSE`, or `CHANGELOG` at the root).
2. The agent reads that file during docs generation.
3. Run crashes with the 400 above.

## Suggested fix

I worked around this in OpenWiki itself with a `wrapToolCall` middleware that
rewrites unsupported binary blocks (file/image/audio/video with a mimeType
Anthropic doesn't accept) into a text placeholder before they reach the
model — PR incoming.

A more complete upstream fix likely belongs in `deepagents`' `read_file` tool:
either treat unknown/extensionless files as text by default (or sniff
content), or avoid mapping unsupported binary blocks to a provider-specific
content type the provider will reject.

## Environment

- openwiki 0.0.1
- provider: anthropic, model: claude-sonnet-5

charlesprost:
## Summary

Fixes #65.

`read_file` (from `deepagents`) returns unrecognized or extensionless files
(`Dockerfile`, `LICENSE`, `CHANGELOG`, `CODEOWNERS`, ...) as a generic
`{ type: "file", mimeType: "application/octet-stream", data }` content block.
When that reaches Anthropic it's sent as a "document" content block, which the
API rejects with a 400 since document blocks only accept `application/pdf` —
crashing the whole run.

**First commit** adds a `wrapToolCall` middleware that rewrites any such
unsupported binary file block into a text placeholder before it reaches the
model, and attaches it to both the main agent and the auto-added
general-purpose subagent (which runs its own middleware stack and doesn't
inherit the main agent's `middleware` — confirmed by reproducing the crash
from that subagent specifically after fixing the main agent alone).

**Second commit** hardens that fix after a closer look at deepagents' other
binary content types and this repo's other model providers, so the fix
doesn't regress anyone else:
- Broadens the sanitizer to also cover `image`/`audio`/`video` blocks
  (e.g. HEIC/HEIF images, or any audio/video file at all — Anthropic doesn't
  support those either).
- Fixes the redeclared general-purpose subagent's harness-profile lookup to
  key off the same LangChain-class-based provider hint deepagents uses
  internally, instead of this repo's own provider label (baseten/fireworks
  are really `ChatOpenAI`; openrouter's `ChatOpenRouter` isn't recognized at
  all), so it resolves the same profile the main agent does.
- Fixes that same lookup to handle model ids that contain a colon themselves
  (a common OpenRouter convention, e.g. `...:free`), matching deepagents'
  own fallback behavior.
- Threads the main agent's `tools` through to the redeclared subagent, since
  once redeclared it no longer inherits them automatically.

## Test plan

- [x] `pnpm build` (tsc) passes with no errors, for both commits individually
- [x] `pnpm lint` passes with no errors
- [x] Reproduced the original crash on a repo containing `Dockerfile`,
      `LICENSE`, `CHANGELOG`, `CODEOWNERS` at the root
- [x] Confirmed `openwiki --init` completes successfully on that repo after
      the fix, including subagent-delegated reads
- [x] Verified the corrected harness-profile resolution against the real
      `deepagents` package output for anthropic/openai/baseten/fireworks/
      openrouter model specs, including colon-containing model ids

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -3,6 +3,7 @@ import { chmod, mkdir } from "node:fs/promises";
 import path from "node:path";
 import { ChatAnthropic } from "@langchain/anthropic";
 import { ToolMessage } from "@langchain/core/messages";
+import type { StructuredTool } from "@langchain/core/tools";
 import { SqliteSaver } from "@langchain/langgraph-checkpoint-sqlite";
 import { ChatOpenAI } from "@langchain/openai";
 import { ChatOpenRouter } from "@langchain/openrouter";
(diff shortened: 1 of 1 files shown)


charlesprost:
Closing — recreated from the scality/openwiki fork instead of my personal fork. See the replacement PR.

PR #68, opened 2026-07-03
feat: add DeepSeek as a first-class provider

himanshu231204:
## Summary

DeepSeek exposes an OpenAI-compatible API at https://api.deepseek.com/v1. This PR adds it as a first-class provider in OpenWiki, alongside the existing openai, anthropic, baseten, fireworks, and openrouter providers.

It also patches a DeepSeek-API limitation: the chat completions endpoint only accepts text content blocks and rejects every other variant (file, image, …) with a 400. The deepagents agent loop emits non-text blocks on subsequent turns, which made --init/--update unusable against DeepSeek. A targeted, DeepSeek-only fetch rewriter flattens array content into a single text string before sending.

## Changes

| File | Change |
|------|--------|
| src/constants.ts | Add `DEEPSEEK_API_KEY_ENV_KEY` + `DEEPSEEK_BASE_URL`; extend `OpenWikiProvider` and `SELECTABLE_OPENWIKI_PROVIDERS` with `"deepseek"`; add `deepseek` entry to `PROVIDER_CONFIGS` exposing `deepseek-chat` (V3) and `deepseek-reasoner` (R1). |
| src/env.ts | Add `DEEPSEEK_API_KEY` to `managedEnvKeys` and the credential diagnostics report. |
| src/agent/index.ts | (1) Add an explicit `provider === "deepseek"` branch in `createModel` that uses `ChatOpenAI` with `DEEPSEEK_BASE_URL` + `DEEPSEEK_API_KEY`. (2) Install a one-time global fetch rewriter for `api.deepseek.com` requests that flattens array content into text so DeepSeek's spec-compliant endpoint accepts the payload. |

**91 lines added across 3 files, no deletions.**

## Why an explicit `provider === "deepseek"` branch?

The generic `ChatOpenAI` fallthrough in `createModel` would handle deepseek automatically (it reads `providerConfig.baseURL` and `getProviderApiKeyEnvKey(provider)`). I added an explicit branch anyway because:

- It makes intent self-documenting — future readers see DeepSeek is a first-class target, not a config-only afterthought.
- It isolates future provider-specific tweaks (headers, model defaults, response shaping) from the OpenAI path.

## Operator setup

No secrets in this PR. To run OpenWiki against DeepSeek:

**~/.openwiki/.env:**

```
OPENWIKI_PROVIDER="deepseek"
OPENWIKI_MODEL_ID="deepseek-chat"
DEEPSEEK_API_KEY="<your deepseek key>"
```

**Or in shell:**

```bash
export OPENWIKI_PROVIDER=deepseek
export OPENWIKI_MODEL_ID=deepseek-chat
export DEEPSEEK_API_KEY='<your deepseek key>'
openwiki --init
```

Default model is `deepseek-chat`. Set `OPENWIKI_MODEL_ID=deepseek-reasoner` to switch to R1. LangSmith tracing remains optional via `LANGSMITH_API_KEY`.

## Verification

- ✅ `pnpm install --frozen-lockfile`
- ✅ `pnpm run build` (tsc, strict)
- ✅ `pnpm run format:check`
- ⚠️ `pnpm run lint:check` - Pre-existing lint errors in `.opencode/` directory (not from this PR)

## Notes for reviewer

- `OPENAI_BASE_URL` is in `deprecatedEnvKeys` and is intentionally skipped on env load — the existing code path does not support overriding openai base URL via env. Adding a first-class provider is the supported way to add a new OpenAI-compatible endpoint.
- I did not touch any existing provider config, credential handling, or onboarding UI flow. `SELECTABLE_OPENWIKI_PROVIDERS` and `getProviderModelOptions` are data-driven, so `src/credentials.tsx` automatically surfaces the new provider and its two models with no further edits.
- The fetch rewriter is gated on a module-level boolean and only rewrites requests whose URL contains `api.deepseek.com`. Other providers see the original `globalThis.fetch` and are unaffected.
- The rewriter intentionally drops non-text content (file, image, etc.) by replacing each block with `[omitted <type> block]`. DeepSeek cannot consume those blocks anyway; this avoids the 400. If we ever need to forward tool/file payloads to DeepSeek, that would be a follow-up.

Closes #58


himanshu231204:
Hi @langchain-ai/core-team! 👋

This PR adds DeepSeek as a first-class provider (closes #58). It includes:

- DeepSeek Chat (V3) + Reasoner (R1) models
- Credential diagnostics support
- Fetch rewriter to handle DeepSeek's text-only API spec

The build passes locally (pnpm run build ✅). Could a maintainer please approve the workflow runs so CI can verify?

Happy to address any feedback! 🙏

rajpratham1 (changes requested):
Thanks for adding DeepSeek support! I think this should be Request Changes before merging. The provider configuration itself looks straightforward, but this PR also installs a global fetch wrapper that rewrites request bodies for any request targeting api.deepseek.com. That changes process-wide behavior and could have unintended side effects for other code using fetch. I'd prefer to see this implemented closer to the model/client layer (or otherwise scoped more narrowly), along with tests covering the request rewriting behavior and ensuring it only affects the intended requests.

himanshu231204:
@zan22ye Could you please review this PR? The merge conflict in src/agent/index.ts has been resolved.

himanshu231204 (commented):
Could you please review this PR?"

himanshu231204:
@rajpratham1 Thanks for the review feedback! I've addressed both concerns:

**1. Scoped fetch rewriter** — Replaced the global globalThis.fetch wrapper with createDeepSeekFetchRewriter() that returns a scoped fetch function. It's now passed via configuration.fetch to the ChatOpenAI constructor, so it only affects requests from that specific model instance — no process-wide side effects.

**2. Test coverage** — Added 13 unit tests in 	est/deepseek-rewriter.test.ts covering:
- String content passthrough
- Array content flattening (text blocks)
- Non-text block replacement with [omitted <type> block] placeholders
- Mixed content across messages
- Error passthrough (invalid JSON body)
- Empty/missing messages
- No body requests
- Request property preservation (method, headers)
- No mutation of original body objects
- Instance isolation (independent original fetch per rewriter)

All 19 tests pass (
px vitest run), build passes (pnpm run build). Ready for re-review!

himanshu231204:
@bracesproul Could you please review this PR? It adds DeepSeek as a first-class provider with scoped fetch rewriting and full test coverage. Thanks!

zan22ye:
It looks like the underlying request may already be covered. As noted by a maintainer in [#58](https://github.com/langchain-ai/openwiki/pull/58#issuecomment-4887957018), [#110](https://github.com/langchain-ai/openwiki/pull/110) already adds support for OpenAI-compatible APIs, and they considered that sufficient to resolve the earlier request. Also, I’m only a contributor on this repo, so I don’t have permission to merge or close PRs myself.

himanshu231204:
@zan22ye Thanks for the context! You're right that #110's openai-compatible provider can technically reach DeepSeek's API. However, this PR still adds value as a first-class provider for two key reasons:

1. **DeepSeek-specific fetch rewriter** — DeepSeek's chat completions endpoint rejects non-text content blocks (file, image, etc.) with a 400 error. The deepagents agent loop emits these on subsequent turns, which makes -init/--update unusable against DeepSeek without this patch. The generic openai-compatible provider doesn't handle this — users would hit silent failures on multi-turn conversations.

2. **Better DX** — With a dedicated provider, users just need:
   `
   OPENWIKI_PROVIDER=deepseek
   DEEPSEEK_API_KEY=sk-...
   `
   No need to know the base URL or look up model IDs. The curated model list (deepseek-chat, deepseek-reasoner) is also documented in onboarding.

The openai-compatible provider is great for arbitrary gateways, but for a well-known provider with known API quirks, a first-class entry with built-in workarounds is the better UX.

colifran:
Closing - OpenAI-compatible APIs are already supported.

PR #80, opened 2026-07-04
feat: support alternative base URLs for the Anthropic provider

ppsplus-bradh:
## Summary

The Anthropic provider previously always talked to the default Anthropic API endpoint. Unlike the OpenAI-compatible providers (which thread a base URL into `ChatOpenAI`), the Anthropic branch in `createModel` constructed `ChatAnthropic` with only an API key, so there was no way to point it at an alternative, Anthropic-compatible base URL (e.g. a self-hosted or proxied gateway).

This PR adds support for an alternative base URL for the Anthropic provider via a new `ANTHROPIC_BASE_URL` environment variable, and generalizes base-URL resolution so it's easy to extend to other providers.

## Changes

- **`src/constants.ts`**: add an optional `baseUrlEnvKey` to `ProviderConfig`, wire it up for the anthropic provider (`ANTHROPIC_BASE_URL`), and add a `resolveProviderBaseUrl()` helper that prefers a provider's alternative base URL env var over the built-in default (falling back to the SDK default when neither is set).
- **`src/agent/index.ts`**: pass the resolved base URL to `ChatAnthropic` as `anthropicApiUrl`, route the OpenAI-compatible branch through the same helper, and surface the resolved base URL in debug output.
- **`src/env.ts`**: persist `ANTHROPIC_BASE_URL` in `~/.openwiki/.env` handling and include it in credential diagnostics, treating the base URL as a non-secret value (shown in full rather than masked).
- **Docs**: document alternative base URLs in the README and the OpenWiki docs (`cli/usage`, `agent/workflow`, `operations/credentials-and-updates`).

## Usage

```bash
OPENWIKI_PROVIDER=anthropic
ANTHROPIC_API_KEY=your-key
ANTHROPIC_BASE_URL=https://your-gateway.example.com/anthropic
```

When `ANTHROPIC_BASE_URL` is unset, behavior is unchanged (the default Anthropic API endpoint is used).

## Testing

- `pnpm build`, `pnpm lint:check`, and `pnpm format:check` pass.
- Verified end-to-end against a local build: with `ANTHROPIC_BASE_URL` pointed at a local HTTP server, the Anthropic client issued its `POST /v1/messages` request to the alternative base URL instead of the default API endpoint. With the variable unset, requests continue to go to the default endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,4 +72,18 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
 
+### Alternative base URLs
+
+To route the Anthropic provider at an alternative, Anthropic-compatible endpoint
+(for example a self-hosted or proxied gateway) instead of the default API, set
+`ANTHROPIC_BASE_URL` alongside `ANTHROPIC_API_KEY`:
+
+```bash
+OPENWIKI_PROVIDER=anthropic
+ANTHROPIC_API_KEY=your-key
+ANTHROPIC_BASE_URL=https://your-gateway.example.com/anthropic
+```
+
+The base URL can be set in your environment or stored in `~/.openwiki/.env`.
+
 If there's an inference provider or model you'd like to see added, please open a PR!
--- a/openwiki/agent/workflow.md
+++ b/openwiki/agent/workflow.md
@@ -23,9 +23,11 @@ Chat runs skip metadata writes entirely.
 
 `createModel()` in `src/agent/index.ts` branches by provider:
 
-- **anthropic**: `new ChatAnthropic(modelId, { apiKey })` — uses `@langchain/anthropic` directly.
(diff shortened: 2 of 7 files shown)


rajpratham1 (approved):
This is a well-scoped feature that cleanly adds support for custom Anthropic-compatible gateways without changing default behavior.

What's good
✅ Adds ANTHROPIC_BASE_URL support while preserving the default Anthropic endpoint.
✅ Centralizes base URL resolution via resolveProviderBaseUrl() instead of duplicating logic.
✅ Updates documentation, environment handling, diagnostics, and runtime consistently.
✅ Keeps the implementation extensible for future providers that may also support configurable base URLs.
Minor suggestions (non-blocking)
Validate the URL
Trimmed strings are accepted, but malformed values (e.g. abc) will only fail at runtime.
Consider validating with new URL() before using or saving the value.
Generalize diagnostics
isNonSecretDiagnosticKey() currently special-cases ANTHROPIC_BASE_URL.
If more providers gain configurable base URLs later, a generic mechanism based on baseUrlEnvKey would scale better.
Add tests
Add regression tests for:
resolveProviderBaseUrl() returns override when set.
Falls back to default when unset.
ChatAnthropic receives anthropicApiUrl.
Empty/whitespace ANTHROPIC_BASE_URL falls back to the default endpoint.

bracesproul:
closing as i think you already solved for this in #110

PR #82, opened 2026-07-04
Fix: Strip provider prefix from Anthropic model ID to prevent 404 (Fi…

### LINKED ISSUE #71: Predefined Anthropic model IDs are prefixed (anthropic/claude-sonnet-5) but passed verbatim to ChatAnthropic → 404 on first run
## Summary

Selecting provider **Anthropic** with the predefined "Claude Sonnet" option stores `OPENWIKI_MODEL_ID=anthropic/claude-sonnet-5`, and the very first API call fails:

```
404 {"type":"error","error":{"type":"not_found_error","message":"model: anthropic/claude-sonnet-5"}}
```

The Anthropic API expects the bare model id (`claude-sonnet-5`); the `anthropic/` prefix is never stripped. Out-of-the-box Anthropic setup is broken for both predefined options.

## Root cause (openwiki 0.0.1)

- `dist/constants.js` (~lines 66-67): predefined Anthropic options are provider-prefixed —
  `{ id: "anthropic/claude-opus-4.8", label: "Claude Opus" }`, `{ id: "anthropic/claude-sonnet-5", label: "Claude Sonnet" }`.
- `dist/agent/index.js:236`: `createModel` does `new ChatAnthropic(modelId, {...})` with the raw string — no prefix stripping. (`normalizeModelId` in `dist/constants.js` is just `value.trim()`.)

The prefixed form makes sense for OpenRouter (where it IS the model id) but not for direct-provider clients.

## Suggested fix

Strip a leading `<provider>/` when instantiating direct-provider clients (Anthropic/OpenAI/etc.), or store bare ids in each provider's predefined options.

## Workaround

Set the bare id in `~/.openwiki/.env`:

```
OPENWIKI_PROVIDER=anthropic
OPENWIKI_MODEL_ID=claude-sonnet-5
```

## Environment

openwiki 0.0.1, npm global install, Linux, node v22.22.3.

🤖 Investigated with [Claude Code](https://claude.com/claude-code)

shryexe:
### Description
This PR resolves the issue where predefined Anthropic model IDs (like `anthropic/claude-sonnet-5`) cause a 404 error on the first run when passed directly to `ChatAnthropic`. 

The Anthropic API expects a bare model ID, but the predefined options include the provider prefix.

### Changes Made
- Added logic in `src/agent/index.ts` to strip the provider prefix from the `modelId` before instantiating the direct-provider client.
- This ensures the model receives `claude-sonnet-5` instead of `anthropic/claude-sonnet-5`, preventing the `not_found_error`.

Fixes #71

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -379,7 +379,9 @@ function resolveModelId(
 
 async function createModel(provider: OpenWikiProvider, modelId: string) {
   if (provider === "anthropic") {
-    return new ChatAnthropic(modelId, {
+    // Strip provider prefix if present (e.g., "anthropic/claude-sonnet-5" -> "claude-sonnet-5")
+    const bareModelId = modelId.includes('/') ? modelId.split('/').pop()! : modelId;
+    return new ChatAnthropic(bareModelId, {
       apiKey: process.env[getProviderApiKeyEnvKey(provider)],
     });
   }
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -380,7 +380,9 @@ function resolveModelId(
 async function createModel(provider: OpenWikiProvider, modelId: string) {
   if (provider === "anthropic") {
     // Strip provider prefix if present (e.g., "anthropic/claude-sonnet-5" -> "claude-sonnet-5")
-    const bareModelId = modelId.includes('/') ? modelId.split('/').pop()! : modelId;
+    const bareModelId = modelId.startsWith("anthropic/") 
+  ? modelId.slice("anthropic/".length) 
+  : modelId;
     return new ChatAnthropic(bareModelId, {
       apiKey: process.env[getProviderApiKeyEnvKey(provider)],
     });


rajpratham1 (approved):
This is a small, targeted fix that addresses a real compatibility issue.

What looks good

✅ Prevents passing provider-prefixed model IDs like:

anthropic/claude-sonnet-5

to the Anthropic SDK, which expects:

claude-sonnet-5
✅ Preserves existing behavior for already-correct model IDs.
✅ Low-risk, localized change.

Instead of:

modelId.split("/").pop()!

consider stripping only the known anthropic/ prefix:

const bareModelId = modelId.startsWith("anthropic/")
  ? modelId.slice("anthropic/".length)
  : modelId;

This avoids unintentionally modifying model IDs that may legitimately contain additional slashes in the future.

Also consider adding a regression test for both:

claude-sonnet-5
anthropic/claude-sonnet-5

shryexe:
I have updated the parsing logic to use startsWith safely! I wasn't entirely sure where the specific unit tests for this parsing logic live in the repository structure. Could you point me to the correct test file so I can add the regression tests for claude-sonnet-5 and anthropic/claude-sonnet-5?"

bracesproul:
i think this would just be from user error since the pre-defined anthropic model ids dont include the `anthropic/` prefix

would rather not add extra logic to check for that since it'll make the repo harder to maintain. imo for this case, if the user adds an incorrect model id and it fails, thats expected and they can correct that

PR #83, opened 2026-07-04
fix: steer model to use write_file for new file creation

### LINKED ISSUE #81: edit_file tool call fails schema validation when creating a new file (new_string undefined)
## Summary

`openwiki --update` (and `--init`) reliably fails partway through a run when the agent tries to **create a new page** via `edit_file`. The model emits an `edit_file` call with `old_string: ""` (create-file semantics) but **omits `new_string`**, and the tool's schema rejects the missing argument, aborting the whole run with a non-zero exit. Because it exits before writing/committing, no output is produced.

## Version

`openwiki@0.0.1` (the only published version; `npm install --global openwiki` installs it).

Provider: `anthropic`, model `claude-sonnet-5`. Running in GitHub Actions (`--update --print`).

## Error

```
Error invoking tool 'edit_file' with kwargs {"file_path":".../data-and-content-model.md","old_string":""} with error: Error: Received tool input did not match expected schema

✖ Invalid input: expected string, received undefined
  → at new_string
    at DynamicStructuredTool.call (.../@langchain/core/dist/tools/index.js:113:10)
    ...
    at async PregelRunner._executeTasksWithRetry (.../@langchain/langgraph/dist/pregel/runner.js:147:24)
 Please fix the error and try again.
```

The run then exits 1 (`Process completed with exit code 1`).

## Reproduction

Run `openwiki --update --print` against a medium/large repo where several planned pages don't yet exist. Non-deterministic in *which* page it dies on (it reaches a different page each run), but **reliably fatal** — reproduced 4/4 runs on the same repo. Each run does several minutes of real LLM work before hitting the create-file path and aborting, so it is not a credits/timeout issue.

## Likely cause & suggested fix

The `edit_file` tool's input schema appears to require `new_string: z.string()`. When the model uses `edit_file` to *create* a file (`old_string: ""`) it sometimes omits `new_string`. Either:

1. make `new_string` tolerant of a create call — `z.string().optional().default("")` — and let the handler treat empty/undefined `new_string` as "write `old_string`-less content", or
2. steer create-file operations to `write_file` in the tool descriptions / system prompt so `edit_file` is only used for true edits, or
3. have the agent loop **surface the validation error back to the model as a recoverable tool error** (retry) rather than aborting the whole run — the message "Please fix the error and try again" suggests a retry was intended but the process still exits 1.

Any of these would let long runs complete instead of dying on the first create-a-new-page step. Happy to test a patch.

legacy7838-create:
## Problem

`openwiki --update` and `--init` reliably abort mid-run when the model emits an `edit_file` call with `old_string: ""` (create-file semantics) but omits `new_string`. The Zod schema rejects the missing field, the error propagates through the middleware stack unhandled, and the process exits 1.

**Error:**
```
Error invoking tool 'edit_file' with kwargs {"file_path":".../data-and-content-model.md","old_string":""} with error: Error: Received tool input did not match expected schema

✖ Invalid input: expected string, received undefined
  → at new_string
```

## Root cause

Two interacting bugs + one prompt gap:

1. **`deepagents` `edit_file` schema** — `new_string: z.string()` has no `.optional()` or `.default("")`. A missing `new_string` fails Zod validation with `ToolInputParsingException` before the tool handler runs.

2. **`langchain` `ToolNode` error routing** — `ReactAgent` instantiates `ToolNode` without `handleToolErrors: true`. When `ToolInputParsingException` propagates through `wrapToolCall` middleware, `#handleError` re-throws it (`isMiddlewareError && handleToolErrors !== true`), so "Please fix the error and try again" never reaches the model — the error is fatal.

3. **Openwiki prompt gap** — The system prompt does not explicitly tell the model to use `write_file` for new files. The model guesses `edit_file` with empty `old_string` for file creation, triggering both upstream bugs.

## Fix

Add an explicit prompt rule in `src/agent/prompt.ts` that directs the model to always use `write_file` for new files and never `edit_file` for file creation. This prevents the malformed tool call from being issued in the first place.

```diff
- Do not run commands that search outside the target repository.
+- Do not run commands that search outside the target repository.
+- To create a new file, always use write_file. Never use edit_file for file creation; edit_file is only for modifying existing files.
```

This is the only lever openwiki controls. Both upstream bugs (schema strictness and middleware error routing) should be filed separately against `deepagents` and `langchain`.

Fixes #81


rajpratham1 (approved):
This is a small, focused prompt improvement that addresses a common agent failure mode.

What looks good
✅ Clarifies that new files must be created with write_file, while edit_file is only for existing files.
✅ Pure prompt change with minimal risk.
✅ Helps reduce tool misuse without changing runtime logic.
✅ Fits naturally with the existing prompt instructions.

If edit_file already errors on missing files, consider adding a regression test or evaluation demonstrating that the model now chooses write_file for new file creation. This would help verify the prompt change has the intended effect.

colifran:
Thanks for the PR! This was resolved upstream in deepagentsjs and in langchainjs. Dependencies pulled in here: https://github.com/langchain-ai/openwiki/pull/224

PR #87, opened 2026-07-04
feat: add Ollama as a model provider

subusaran:
Add Ollama as a first-class LLM provider, reusing the existing ChatOpenAI path against Ollama's OpenAI-compatible /v1 endpoint (no new dependency).

Ollama is keyless by default for local daemons (http://localhost:11434/v1). OLLAMA_BASE_URL overrides the endpoint, and a real OLLAMA_API_KEY is required when the base URL points at a remote host (e.g. Ollama Cloud at https://ollama.com/v1) -- detected via ollamaUsesLocalBaseURL, which treats localhost/127.0.0.1/::1/0.0.0.0 as local.

- src/constants.ts: ollama provider config + env keys; getProviderDefaultApiKey and ollamaUsesLocalBaseURL make the keyless behavior base-URL-aware so remote/cloud targets still require a key.
- src/agent/index.ts: resolveProviderBaseURL (honors OLLAMA_BASE_URL) and a default-key fallback in createModel. Use || instead of ?? for the OLLAMA_API_KEY / OLLAMA_BASE_URL fallbacks so empty-string values fall back to the default rather than being kept (which crashed or misrouted requests).
- src/cli.tsx / src/credentials.tsx: skip the API-key step for keyless ollama in the startup guards and interactive onboarding.
- src/env.ts: register OLLAMA_API_KEY / OLLAMA_BASE_URL as managed env keys and in credential diagnostics; suppress the false "contains bracketed suffix/text" warning for http(s) URLs so IPv6 hosts like [::1] are accepted.
- README.md: document Ollama + Ollama Cloud setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 ## Customizing
 
-OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
+OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI, Anthropic, and Ollama out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
+
+Ollama runs locally and needs no API key by default (it connects to `http://localhost:11434/v1`). To use a remote daemon or [Ollama Cloud](https://docs.ollama.com/cloud), set `OLLAMA_BASE_URL` (e.g. `https://ollama.com/v1`) and `OLLAMA_API_KEY` in `~/.openwiki/.env`.
 
 If there's an inference provider or model you'd like to see added, please open a PR!
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -21,9 +21,13 @@ import {
   getDefaultModelId,
   getProviderApiKeyEnvKey,
   getProviderConfig,
+  getProviderDefaultApiKey,
   getProviderLabel,
   isValidModelId,
   normalizeModelId,
+  OLLAMA_API_KEY_ENV_KEY,
+  OLLAMA_BASE_URL_ENV_KEY,
+  OLLAMA_DEFAULT_BASE_URL,
   OPENAI_API_KEY_ENV_KEY,
   OPENROUTER_API_KEY_ENV_KEY,
   OPENROUTER_BASE_URL,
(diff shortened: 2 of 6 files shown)


arrase:
### Technical Review: Ollama Integration via `/v1` Endpoint

While reusing the `ChatOpenAI` path against Ollama's `/v1` endpoint is a convenient abstraction, this approach **does not constitute a "first-class" integration**. Relying on the OpenAI compatibility layer fundamentally strips away critical configuration capabilities required for managing local models effectively.

Below are the architectural flaws in the current implementation and the technical requirements for a genuine native integration.

#### 1. Loss of Context Window Control (`num_ctx`)

The most severe architectural limitation of using the OpenAI-compatible endpoint (`/v1/chat/completions`) is the inability to control the model's context window size.

* In Ollama's native API, the context size is explicitly defined via the `num_ctx` parameter within the `options` object.
* The OpenAI API lacks a direct equivalent for this in the inference request (OpenAI's `max_tokens` parameter controls the output tokens, not the context window size loaded into the engine's memory).
* By depending on the OpenAI wrapper, Ollama will instantiate models with their default `num_ctx`. This inevitably leads to silent truncation of long prompts or RAG pipeline failures, leaving the user with no standard mechanism in the interface to override the default allocation.

#### 2. Omission of Native Inference Parameters

A true first-class integration with a local execution engine must allow tuning of specific hyperparameters that users of open weights require to optimize inference. The standard OpenAI interface ignores or drops Ollama-native parameters, including:

* `mirostat`, `mirostat_tau`, `mirostat_eta` (for dynamic entropy control).
* `tfs_z` (Tail Free Sampling).
* `repeat_penalty` and `repeat_last_n`.
* `num_predict` (the actual equivalent to `max_tokens` in the native Ollama backend).

rajpratham1 (approved):
This is a well-designed provider addition with thoughtful handling of Ollama's local-first workflow.

What looks good
✅ Adds Ollama consistently across:
Provider constants
CLI
Credentials
Environment handling
Documentation
✅ Supports the standard OpenAI-compatible endpoint (http://localhost:11434/v1).
✅ Makes the API key optional for local Ollama by supplying the conventional "ollama" value automatically.
✅ Correctly requires a real API key when OLLAMA_BASE_URL points to a remote host (e.g. Ollama Cloud).
✅ Allows overriding the base URL through OLLAMA_BASE_URL.
✅ Updates setup UI so users aren't forced to enter an unnecessary API key.
Minor suggestions (non-blocking)
Add tests covering:
Local default (localhost) → API key optional.
Remote OLLAMA_BASE_URL → API key required.
IPv6/local address detection.
Invalid OLLAMA_BASE_URL values.
Consider allowing users to configure additional local hostnames (e.g. custom Docker hostnames), though the current implementation is reasonable.

arrase:
My two cents: https://github.com/langchain-ai/openwiki/pull/107

bracesproul:

PR #88, opened 2026-07-04
feat: add DeepSeek provider support

IvanYang007:
Adds DeepSeek as a first-class provider alongside the existing Anthropic, OpenAI, Baseten, Fireworks, and OpenRouter providers.

### `src/constants.ts`
- Add `deepseek` to the `OpenWikiProvider` type
- Add `DEEPSEEK_API_KEY` env var constant
- Add DeepSeek config entry with OpenAI-compatible base URL (`https://api.deepseek.com/v1`)
- Add `deepseek-chat` (DeepSeek V3) and `deepseek-reasoner` (DeepSeek R1) model options
- Add auto-detection in `resolveConfiguredProvider` — when `DEEPSEEK_API_KEY` is set, DeepSeek is selected automatically

### `src/agent/index.ts`
- Add fetch-level HTTP interceptor that strips `file` content blocks from outgoing requests when using the DeepSeek provider
- DeepSeek's API rejects `file` content blocks with `unknown variant 'file'` error (only `text` type is supported)
- The interceptor only activates for requests to the DeepSeek API endpoint and is transparent for other providers
- Import `DEEPSEEK_API_KEY_ENV_KEY` for the interceptor

## Why

DeepSeek's API is OpenAI-compatible for chat and tool calling, so it works with the existing `ChatOpenAI` model path. However, DeepSeek does not support `file` content blocks in message content arrays, which LangChain's `ChatOpenAI` can produce. The HTTP interceptor strips these before they reach DeepSeek's API.

Tested end-to-end: `openwiki --init` on a real project using `deepseek-chat` successfully generated 4 documentation pages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3605 @@
+{
+  "name": "openwiki",
+  "version": "0.0.1",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "openwiki",
+      "version": "0.0.1",
+      "license": "MIT",
+      "dependencies": {
+        "@langchain/anthropic": "^1.5.1",
+        "@langchain/core": "^1.2.1",
+        "@langchain/langgraph-checkpoint-sqlite": "^1.0.3",
+        "@langchain/openai": "^1.5.3",
+        "@langchain/openrouter": "^0.4.3",
+        "deepagents": "^1.10.5",
+        "ink": "^5.1.0",
+        "langchain": "^1.5.1",
+        "marked": "^18.0.5",
+        "react": "^18.3.1"
+      },
+      "bin": {
+        "openwiki": "dist/cli.js"
+      },
+      "devDependencies": {
+        "@eslint/js": "^10.0.1",
+        "@types/node": "^22.10.2",
+        "@types/react": "^18.3.17",
+        "eslint": "^10.5.0",
+        "prettier": "^3.8.4",
+        "tsx": "^4.19.2",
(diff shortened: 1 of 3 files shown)


rajpratham1 (changes requested):
This PR adds DeepSeek support, but I wouldn't approve it in its current state.

What's good
✅ Adds DeepSeek as a provider consistently.
✅ Adds provider configuration and model options.
✅ Attempts to handle DeepSeek's rejection of file content blocks.
Blocking issues

Global fetch override is unsafe

globalThis.fetch = ...

This mutates the global fetch implementation and never restores it. Every subsequent request in the process—including non-DeepSeek requests—will go through the interceptor.

Interceptor is too broad

if (url.includes("/v1/chat/completions"))

This matches any provider using the same endpoint path, not just DeepSeek. It should also verify the request is going to the DeepSeek base URL.

Removes data instead of degrading gracefully

filter(block => block.type !== "file")

File blocks are silently discarded. A text placeholder (or extracted text when possible) would preserve context better than deleting content entirely.

No cleanup/idempotency
Calling createDeepSeekModel() multiple times installs multiple wrappers around fetch, which can lead to unexpected behavior.
Large unrelated lockfile update
The PR adds over 3,600 lines to package-lock.json even though no new dependencies are introduced. Unless dependency versions actually changed, this should be removed to keep the PR focused.
Missing tests
There are no tests covering:
DeepSeek requests strip file blocks.
Other providers remain unchanged.
Multiple model creations don't stack fetch wrappers.

bracesproul:

PR #90, opened 2026-07-04
fix: preserve `OPENAI_BASE_URL` when loading and saving environment

### LINKED ISSUE #64: OpenAI Base URL env var is dropped in src/env.ts
I use OpenAI through a proxy for auditing and cost controls, but this proxy is completely transparent so in general I tend to use the openai provider a project offers and change the OPENAI_BASE_URL env var to redirect the traffic.

The code for openwiki deliberately drops this env var which forces you into using strategies like prepending the env var before the openwiki call to get it to work. 

Is there a reason OPENAI_BASE_URL is being dropped?

arpitparulkar:
## Summary

This PR preserves `OPENAI_BASE_URL` when loading and saving the OpenWiki environment.

Previously, `OPENAI_BASE_URL` was included in `deprecatedEnvKeys`, causing it to be:

- skipped when loading `~/.openwiki/.env`
- removed when saving `~/.openwiki/.env`

This prevented users from configuring a custom OpenAI-compatible endpoint (for example, transparent proxies used for auditing or cost control), even though the OpenAI SDK still supports `OPENAI_BASE_URL`.

This change removes only `OPENAI_BASE_URL` from `deprecatedEnvKeys`, while leaving `OPENAI_ORG_ID` and `OPENAI_PROJECT` unchanged.

Fixes #64.


rajpratham1 (changes requested):
This PR simply removes OPENAI_BASE_URL from deprecatedEnvKeys, but the title claims "Preserve OPENAI_BASE_URL when loading and saving environment." The implementation doesn't fully achieve that.

Why request changes?

Removing it from deprecatedEnvKeys only stops it from being deleted. It does not guarantee that:

OPENAI_BASE_URL is loaded into process.env.
It is included in managed environment keys.
It is written back when saving .env.
Diagnostics or validation handle it correctly.

If any of those code paths ignore OPENAI_BASE_URL, users can still lose the setting or have it silently ignored.

What should be added?

The PR should verify that OPENAI_BASE_URL is:

✅ Loaded from .env
✅ Preserved when saving .env
✅ Not treated as deprecated
✅ Covered by tests proving the value survives a load/save cycle

bracesproul (dismissed):
nice, ty!

bracesproul (commented):
can you please fix the formatting failure (run `pnpm format` from repo root), push up the commit then re-request my review and ill get it merged

colifran (approved):

PR #91, opened 2026-07-04
feat: add Z.AI (GLM Coding Plan) provider

AndrewJacop:
## What this adds

A new `zai` provider so OpenWiki can run GLM-5.2 against **Z.AI's GLM Coding Plan**. Z.AI serves an Anthropic-protocol endpoint at `https://api.z.ai/api/anthropic` for coding-plan subscribers, so this provider routes through `ChatAnthropic` with `anthropicApiUrl`.

## Why the Anthropic protocol (not the OpenAI one)

Z.AI's coding plan exposes two protocols:

- OpenAI: `https://api.z.ai/api/coding/paas/v4`
- Anthropic: `https://api.z.ai/api/anthropic`

I first wired `zai` through the OpenAI endpoint via `ChatOpenAI` (mirroring baseten/fireworks/openai). That path crashes at runtime against GLM-5.2:

```
Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware": expected AIMessage or Command, got object
```

Root cause: GLM-5.2 is a reasoning model whose streaming tool-call payloads don't reassemble into a valid `AIMessage` inside `@langchain/openai`'s `ChatOpenAI`. `patchToolCallsMiddleware` is just the messenger (it forwards `handler(request)` unchanged; the bad object originates in the model layer), and `deepagents` adds it unconditionally with no opt-out, so the fix has to be at the protocol layer.

Routing through `ChatAnthropic` at the Anthropic endpoint sidesteps this entirely — the Messages API has native tool-use streaming semantics and is parsed into proper `AIMessage` instances. This is also the endpoint Z.AI recommends for agentic coding tools (Claude Code uses this exact endpoint).

## Changes

- **`src/constants.ts`** — `ZAI_API_KEY_ENV_KEY` (`ZAI_API_KEY`), `ZAI_ANTHROPIC_BASE_URL`, `zai` in the `OpenWikiProvider` union + `SELECTABLE_OPENWIKI_PROVIDERS` (2nd, after openrouter), and the `zai` `PROVIDER_CONFIGS` entry (GLM 5.2).
- **`src/agent/index.ts`** — generalized the `anthropic` branch to `anthropic || zai`, constructing `ChatAnthropic` with `anthropicApiUrl: providerConfig.baseURL` (undefined for plain `anthropic` → behavior unchanged).
- **`src/env.ts`** — `ZAI_API_KEY` in `managedEnvKeys` + credential diagnostics.
- **`src/credentials.tsx`** — `getProviderArticle()` returns `"a"` for `zai`.
- **`src/cli.tsx`** — `ZAI_API_KEY` in the secret-redaction loop.
- **Docs** — README, quickstart, CLI usage (provider table), architecture overview, agent workflow, operations. Doc edits are hand-applied to the existing prose only (no full regen bundled in).

Default provider is unchanged (`openrouter`).

## How to test

```sh
rm -f ~/.openwiki/.env
unset OPENWIKI_PROVIDER ZAI_API_KEY OPENWIKI_MODEL_ID
pnpm dev --init
```

Pick **Z.AI** → paste a coding-plan key → **GLM 5.2** → skip LangSmith. With `OPENWIKI_DEBUG=1`, debug output shows `provider=zai`, `provider.baseUrl="https://api.z.ai/api/anthropic"`, `model=glm-5.2`. The init run streams agent text/tool events and completes without the middleware error.

Verified locally: `pnpm build`, `pnpm lint:check`, and `prettier --check` on all changed files pass.

## Notes

- The `ZAI_API_KEY` works with both the OpenAI and Anthropic coding-plan endpoints (same key, different base URL). I scoped this to the Anthropic endpoint because that's what actually works with the current `@langchain/openai` streaming path.
- No CI workflow changes — the existing workflows only pass `OPENROUTER_API_KEY`. Happy to add `ZAI_API_KEY` passthrough in a follow-up if useful.

Sources for the endpoint:
- [Z.AI Tool Integration](https://docs.z.ai/devpack/tool/others)
- [Z.AI Claude Code guide](https://docs.z.ai/devpack/tool/claude)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,6 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 ## Customizing
 
-OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
+OpenWiki supports OpenRouter, Z.AI, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
 
 If there's an inference provider or model you'd like to see added, please open a PR!
--- a/openwiki/agent/workflow.md
+++ b/openwiki/agent/workflow.md
@@ -23,7 +23,7 @@ Chat runs skip metadata writes entirely.
 
 `createModel()` in `src/agent/index.ts` branches by provider:
 
-- **anthropic**: `new ChatAnthropic(modelId, { apiKey })` — uses `@langchain/anthropic` directly.
+- **anthropic / zai**: `new ChatAnthropic(modelId, { apiKey, anthropicApiUrl })` — uses `@langchain/anthropic` directly (`zai` sets `anthropicApiUrl` to the Z.AI GLM Coding Plan Anthropic endpoint, `https://api.z.ai/api/anthropic`).
 - **openrouter**: `new ChatOpenRouter({ apiKey, baseURL, model, models, route: "fallback", siteName: "OpenWiki" })` — passes a fallback model list so OpenRouter can route around server-side failures.
 - **baseten / fireworks / openai**: `new ChatOpenAI({ apiKey, configuration: { baseURL? }, model })` — OpenAI-compatible clients using the provider's custom base URL when configured.
 
--- a/openwiki/architecture/overview.md
+++ b/openwiki/architecture/overview.md
@@ -40,7 +40,7 @@ The agent runtime resolves the provider via `resolveConfiguredProvider()` in `sr
 
 Model creation branches by provider in `src/agent/index.ts` (`createModel`):
 
-- **anthropic** → `ChatAnthropic` with the Anthropic API key.
+- **anthropic / zai** → `ChatAnthropic` with the provider's API key and optional `anthropicApiUrl` from `PROVIDER_CONFIGS` (`zai` points at the Z.AI GLM Coding Plan Anthropic endpoint).
 - **openrouter** → `ChatOpenRouter` with `route: "fallback"` and a list of fallback models.
 - **baseten / fireworks / openai** → `ChatOpenAI` with the provider's API key and optional custom `baseURL` from `PROVIDER_CONFIGS`.
 
--- a/openwiki/cli/usage.md
+++ b/openwiki/cli/usage.md
@@ -43,8 +43,8 @@ The UI persists provider and model selection back to `~/.openwiki/.env` through 
 
 The first interactive run can prompt for:
 
-- a **provider** (`OPENWIKI_PROVIDER`) — openrouter, baseten, fireworks, openai, or anthropic,
-- the **provider API key** (e.g. `OPENROUTER_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `BASETEN_API_KEY`, `FIREWORKS_API_KEY`),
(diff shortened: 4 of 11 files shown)


pohlipit23:
Hi @AndrewJacop — great catch on routing through the Anthropic protocol. I opened a competing PR (#95) using the OpenAI endpoint before realizing this approach exists, and I just closed it in favor of this one. Wanted to share some corroboration that may help review.

## Independent verification

I tested this branch's `ChatAnthropic` + `anthropicApiUrl` approach against the GLM Coding Plan endpoint with a live key:

- ✅ Simple chat via \`api.z.ai/api/anthropic/v1/messages\` → correct response
- ✅ Full deepagents agent loop (the path that previously crashed) → completes cleanly, **no normalizer needed**
- ✅ Tool-call streaming via \`model.bindTools(...)\` → returns a proper \`AIMessage\` with populated \`tool_calls\`, which is exactly what \`patchToolCallsMiddleware\` expects

## Why the OpenAI path crashes (root cause confirmation)

Your diagnosis that the bad object originates in the model layer is correct. I hit the same \`Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware": expected AIMessage or Command, got object\` error on the OpenAI path and instrumented langchain's \`AgentNode\` validator to dump the exact failing object:

\`\`\`
constructor: ChatMessageChunk
isAIMessage: false
kwargs: { content: \"\", additional_kwargs: {}, response_metadata: {...} }
\`\`\`

The OpenAI-protocol stream ends a tool-call turn with a chunk whose \`delta\` carries no \`role\`, which \`@langchain/openai\`'s completions converter maps to a generic role-less \`ChatMessageChunk\` (the \`else\` fallback in \`convertCompletionsDeltaToBaseMessageChunk\`). \`patchToolCallsMiddleware\` only accepts \`AIMessage\`/\`Command\`, so it rejects it. Switching to the Anthropic Messages API sidesteps this entirely because tool-use has native streaming semantics there.

A couple of minor notes, take or leave:
- I also saw an intermittent \`400 messages[N].content[0].type type error\` on the OpenAI path from non-text content blocks the agent loop emits on later turns — another OpenAI-protocol-only issue the Anthropic path avoids.
- The GLM Coding Plan is concurrency-1, so during testing you may hit \`429\`s (\`Rate limit reached for requests\`); retrying after a short backoff clears it. Not a code issue, just a testing heads-up.

This is the right fix. 🎉

AndrewJacop:
@pohlipit23 — really appreciate the thorough independent verification, and thanks for closing #95 in favor of this. The deeper `ChatMessageChunk` trace is a great find — pinpoints exactly which delta fails (`convertCompletionsDeltaToBaseMessageChunk`'s role-less fallback on the tool-call-ending chunk), which lines up cleanly with why the Anthropic Messages path sidesteps it. Also good to know about the intermittent `400 content[0].type` on later OpenAI turns; reinforces that the Anthropic endpoint is the right default here.

Noted on the concurrency-1 / 429 behavior — useful for anyone testing, and definitely not something to paper over in code.

Branch is ready for review whenever a maintainer has a moment. Happy to address any feedback (rename, placement, tests, etc.).

rajpratham1 (approved):
This is a clean, well-scoped provider addition.

What looks good
✅ Adds Z.AI as a first-class provider consistently across:
Provider constants
Environment handling
CLI
Credentials
Documentation
✅ Reuses the existing ChatAnthropic client by pointing it to the Z.AI Anthropic-compatible endpoint instead of introducing duplicate logic.
✅ Introduces a dedicated ZAI_API_KEY.
✅ Documentation is updated everywhere the provider list appears.
Minor observations
The provider currently exposes only one predefined model (glm-5.2). That's reasonable and users can still provide custom model IDs.
The implementation shares the Anthropic client path, reducing maintenance and code duplication.
Optional follow-up (not blocking)
Add an integration test verifying:
OPENWIKI_PROVIDER=zai selects the correct provider.
ZAI_API_KEY is used.
anthropicApiUrl is correctly set to https://api.z.ai/api/anthropic.
Interactive configuration can select Z.AI.

bracesproul:
closing as #110 can solve for this - it adds support for custom anthropic base urls

Issue #86, opened 2026-07-04
Two bugs in v0.0.1: Anthropic Opus model ID is invalid, and a single malformed tool call crashes the entire run


vscarpenter:
Running `openwiki` against a codebase (provider: `anthropic`) surfaced two independent bugs. They're unrelated in cause but easy to hit back-to-back, so I'm reporting both here — happy to split into separate issues if you'd prefer.

**Environment**
- openwiki `0.0.1` (global install)
- `langchain` `1.5.x`, `deepagents` `1.10.5`, `@langchain/anthropic` `1.5.x`
- Node ≥ 20 (bun global install)
- Provider: `anthropic`, `ANTHROPIC_API_KEY` set

---

### Bug 1 — The Anthropic "Opus" option uses an invalid model ID (`claude-opus-4.8`)

**Summary.** The Anthropic provider's Opus option ships a dotted model ID that the Anthropic API rejects. Anthropic model IDs use dashes.

**Steps to reproduce**
1. Set `OPENWIKI_MODEL_ID="claude-opus-4.8"` (the value from OpenWiki's own model menu), provider `anthropic`.
2. Run `openwiki`.

**Actual**
```
Error 404 {"type":"error","error":{"type":"not_found_error",
"message":"model: claude-opus-4.8 was not found. Did you mean claude-opus-4-8?"},...}
```
Run fails at connect time.

**Expected.** Selecting the built-in "Opus" option produces a working run.

**Root cause.** In `src/constants.ts` (`PROVIDER_CONFIGS.anthropic.modelOptions`), the Opus entry is:
```ts
{ id: "claude-opus-4.8", label: "Opus" }
```
`normalizeModelId()` only does `value.trim()` — there is no dot→dash mapping — so the ID is passed verbatim to `new ChatAnthropic("claude-opus-4.8", …)`, and Anthropic 404s. Note the Opus entry is also inconsistent with the sibling Haiku entry, which already uses the correct dashed convention (`claude-haiku-4-5`).

**Fix.** Use the dashed API ID:
```diff
- { id: "claude-opus-4.8", label: "Opus" }
+ { id: "claude-opus-4-8", label: "Opus" }
```
Worth double-checking the OpenRouter provider entry (`anthropic/claude-opus-4.8`) against OpenRouter's current slug for the same reason.

**Workaround.** Set `OPENWIKI_MODEL_ID="claude-opus-4-8"` explicitly (the permissive `isValidModelId` accepts it).

---

[...]

PR #92, opened 2026-07-04
fix(agent): handle tool validation errors gracefully

Manvikamboz:
#86 issue 

Incorrect Anthropic Opus Model ID:
Updated the model ID options in src/constants.ts to use hyphens (claude-opus-4-8) instead of dots (claude-opus-4.8), aligning it with the requirements of the Anthropic API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "lint:check": "eslint .",
     "prebuild": "pnpm run clean",
     "prepack": "pnpm run build",
-    "start": "node dist/cli.js"
+    "start": "node dist/cli.js",
+    "test": "tsx tests/agent_error.test.ts"
   },
   "dependencies": {
     "@langchain/anthropic": "^1.5.1",
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -191,12 +191,6 @@ async function runOpenWikiAgentCore(
     handleToolErrors: true,
   });
 
-  // Programmatically guarantee handleToolErrors is set to true on the tools node bound toolNode
-  const toolsNode = (agent.graph as any).nodes?.tools;
-  if (toolsNode?.bound) {
-    toolsNode.bound.handleToolErrors = true;
-  }
-
   emitDebug(options, "agent=created");
 
   const input = {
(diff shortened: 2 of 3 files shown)


rajpratham1 (changes requested):
The intent is good, but the implementation is fragile and relies on internal APIs.

Pros
✅ Prevents agent crashes from tool validation errors.
✅ Enables handleToolErrors at agent creation.
✅ Small, focused change.
Concerns
Relies on internal implementation
Mutating (agent.graph as any).nodes?.tools.bound.handleToolErrors depends on private internals.
Future LangGraph/LangChain updates could silently break this.
Uses any
(agent.graph as any) bypasses type safety and makes maintenance harder.
Potential duplicate configuration
handleToolErrors: true is already passed during agent creation.
If that's expected to propagate, the second mutation shouldn't be necessary.
If it isn't, the root cause should be fixed rather than patching internals.
Missing tests
No regression test proving a tool validation error is caught and returned gracefully instead of crashing.
What I'd like before merge
Add a test that intentionally triggers a tool validation error.
Explain why the constructor option alone isn't sufficient.
Avoid mutating private graph internals if a public API exists.

Manvikamboz:
Thanks @rajpratham1 for the thoughtful review! I went back through the implementation and addressed each of the points you raised.

> **1. Add a test that intentionally triggers a tool validation error.**

Done. I added a regression test that intentionally invokes a tool with invalid arguments (violating its Zod schema). The test verifies that the validation error is converted into a `ToolMessage` rather than crashing the agent execution, and I also covered the subagent error propagation path.

> **2. Explain why the constructor option alone isn't sufficient.**

After digging into the codebase, I found that my original assumption was incorrect. The constructor option (`handleToolErrors: true`) is already sufficient when the agent is initialized through the normal `createDeepAgent` flow. My earlier graph mutation was an unnecessary workaround based on an incorrect understanding of the execution path.

> **3. Avoid mutating private graph internals if a public API exists.**

Agreed. I've removed the internal graph mutation entirely along with the `any` cast. The implementation now relies solely on the public API, which is much cleaner and avoids depending on LangGraph internals.

I also cleaned up the temporary debugging files I created while investigating and verified that the project compiles cleanly and the regression tests pass.

Thanks again for the review—it helped me end up with a much cleaner solution.

colifran:
Thanks for the PR! The model issue has been resolved and tool error handling was fixed upstream in langchainjs. Dependencies were bumped in this pr to bring the fixes in: https://github.com/langchain-ai/openwiki/pull/224.

PR #93, opened 2026-07-04
fix(constants): correct Anthropic Opus model IDs

Manvikamboz:
#86 issue 

bug2 Malformed Tool Call Agent Crash: 
Modified the createDeepAgent invocation in src/agent/index.ts to set handleToolErrors: true. I patched the underlying packages (deepagents and langchain) to support forwarding this parameter, and added a programmatic fallback to guarantee that the compiled graph node's handleToolErrors property is configured correctly at runtime.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3340 @@
+{
+  "name": "openwiki",
+  "version": "0.0.1",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "openwiki",
+      "version": "0.0.1",
+      "license": "MIT",
+      "dependencies": {
+        "@langchain/anthropic": "^1.5.1",
+        "@langchain/core": "^1.2.1",
+        "@langchain/langgraph-checkpoint-sqlite": "^1.0.3",
+        "@langchain/openai": "^1.5.3",
+        "@langchain/openrouter": "^0.4.3",
+        "deepagents": "^1.10.5",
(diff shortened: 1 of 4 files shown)


rajpratham1 (approved):
This is a small but important correctness fix. If the previous Opus IDs were invalid, users would fail to select or invoke the model.

Pros
✅ Corrects the Anthropic Opus model ID from claude-opus-4.8 → claude-opus-4-8.
✅ Updates both the Anthropic provider and OpenRouter model list for consistency.
✅ No functional changes beyond fixing the model identifier.
✅ Low-risk change.
Minor suggestion
Add a small test (or constants validation) that verifies configured model IDs are valid and consistent across providers to prevent future typos.

Manvikamboz:
done adding the suggested test script @rajpratham1 

what i have added (or changed):-
1. Corrected Model ID Typo: (in src/constants.ts by updating it from claude-opus-4.8 to claude-opus-4-8 )

2. Added Validation Test Suite: (Created a new test file src/constants.test.ts) to perform following validation :- 
2a) ID Validity
2b) Uniqueness
2c) Consistency Across Providers
2d) Fallback/Default Validity

3. Integrated Test Script: (Added a "test": "tsx --test src/constants.test.ts" npm script in package.json to make running validations simple and compatible across different shells.)

bracesproul:
not sure why this diff is so large... closing tho since this was resolved in another PR

PR #94, opened 2026-07-04
fix: honor selected OpenRouter model instead of silent fallback

### LINKED ISSUE #85: Openrouter use random model instead of selected one.
So I have initialized openwiki with openrouter provider and set deepseek v4 flash as model. I set $3 limit on key (thanks god). After few minutes I see that Key limit exceeded. 
In open router I see it was using sonnet 5! Why do we even need to set model if it does what it wants? 

```
Claude Sonnet 5 | $3,38
GPT-5.4 Mini | $0,13 
DeepSeek V4 Flash | $0,12 

```

ricardoofnl:
Fixes #85.

OpenRouter models were built with a hardcoded fallback list (`openai/gpt-5.4-mini`, `anthropic/claude-sonnet-5`) and `route: "fallback"`. On any transient error, OpenRouter silently switched to one of those models, so users got billed for a model they never selected.

This makes the selected model authoritative: only `model` is sent to OpenRouter (no `models`/`route`), and transient 5xx errors retry the same model instead of switching to a more expensive one.

Verified with `build`, `lint:check`, `format:check`, and a runtime check confirming the request sends only the selected model.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -27,7 +27,6 @@ import {
   OPENAI_API_KEY_ENV_KEY,
   OPENROUTER_API_KEY_ENV_KEY,
   OPENROUTER_BASE_URL,
-  OPENROUTER_FALLBACK_MODEL_IDS,
   OPENWIKI_MODEL_ID_ENV_KEY,
   OPENWIKI_PROVIDER_ENV_KEY,
   resolveConfiguredProvider,
@@ -73,7 +72,7 @@ export async function runOpenWikiAgent(
   const debugFetchCapture = installOpenRouterDebugFetch(options);
 
   try {
-    return await runOpenWikiAgentWithModelFallbacks(
+    return await runOpenWikiAgentWithRetries(
       command,
       cwd,
       options,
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This fixes an important UX issue: when users explicitly select an OpenRouter model, OpenWiki should respect that choice instead of silently switching to fallback models.

Pros
✅ Honors the user-selected model.
✅ Replaces silent model fallback with bounded retries on the same model.
✅ Retries only transient OpenRouter 5xx failures.
✅ Removes confusing fallback routing/debug messages.
✅ Simplifies the code by removing OPENROUTER_FALLBACK_MODEL_IDS and createModelRoute().
Things to verify
Add tests covering:
Successful retry after a transient 5xx.
No retry on 4xx errors.
No fallback to another model.
Retry limit (MAX_RUN_ATTEMPTS) is respected.
Consider whether MAX_RUN_ATTEMPTS = 2 should be configurable rather than hardcoded.

colifran:
@ricardoofnl - thank you for the PR. Closing since this has been resolved: https://github.com/langchain-ai/openwiki/pull/201

ricardoofnl:
oh okay.... thanks for the review

PR #95, opened 2026-07-05
feat: add Z.AI (GLM Coding Plan) as a first-class provider

pohlipit23:
## Summary

Z.AI exposes an OpenAI-compatible chat-completions endpoint for GLM Coding Plan subscribers at `https://api.z.ai/api/coding/paas/v4`. This PR adds it as a first-class provider, alongside the existing `openrouter`, `baseten`, `fireworks`, `openai`, and `anthropic` providers.

- **Provider id:** `zai`
- **Label:** \"Z.AI\"
- **Env key:** `ZAI_API_KEY` (matches the convention used in Z.AI's own docs/tooling)
- **Preset model:** `glm-5.2` (custom model IDs remain selectable via onboarding)

The credential onboarding UI in `src/credentials.tsx` is fully data-driven from `PROVIDER_CONFIGS`, so it surfaces the new provider and model automatically with no edits required there.

## Why this is larger than the DeepSeek PR (#68)

The DeepSeek PR added 99 lines. This PR adds 295 because GLM's coding-plan endpoint has **two** incompatibilities with the deepagents agent loop, both of which crash `--init`/`--update` and both of which needed fixing for the provider to be usable. I reproduced each against the real `--init` flow before writing the fix.

### 1. RESPONSE — role-less streaming chunks (the `--init` crash)

GLM occasionally ends a tool-call turn with an SSE chunk whose `delta` has no `role` field (e.g. a trailing usage-only chunk with `finish_reason: \"tool_calls\"` and an empty content delta). LangChain's completions converter maps a role-less assistant delta to the generic `ChatMessageChunk`, which deepagents' `patchToolCallsMiddleware` rejects:

```
Error: Invalid response from \"wrapModelCall\" in middleware \"patchToolCallsMiddleware\": expected AIMessage or Command, got object
```

I instrumented `langchain`'s `AgentNode` validator to dump the exact failing object, which confirmed it:

```
constructor: ChatMessageChunk
isAIMessage: false
kwargs: { content: \"\", additional_kwargs: {}, response_metadata: {...}, id: \"run-...\" }
```

**Fix:** the normalizer backfills `role: \"assistant\"` on assistant-turn streaming chunks that are missing one. Every chunk in a chat-completions assistant response is an assistant message, so this is always correct for the coding-plan endpoint.

### 2. REQUEST — non-text content blocks (intermittent `400`)

On later turns the agent loop emits Anthropic-style array content with non-text blocks (tool/file payloads). GLM's chat-completions endpoint rejects every non-text block variant with:

```
400 messages[N].content[0].type type error
```

This is the same class of problem DeepSeek hit, so the fix mirrors DeepSeek's proven approach: the normalizer flattens array content into a single text string, replacing each non-text block with `[omitted <type> block]`. (GLM cannot consume those blocks anyway; forwarding them would be a follow-up.)

## Changes

| File | Change |
|------|--------|
| `src/constants.ts` | Add `ZAI_API_KEY_ENV_KEY` + `ZAI_BASE_URL`; extend `OpenWikiProvider` and `SELECTABLE_OPENWIKI_PROVIDERS` with `\"zai\"`; add `zai` entry to `PROVIDER_CONFIGS` exposing `glm-5.2`. |
| `src/env.ts` | Add `ZAI_API_KEY` to `managedEnvKeys` and the credential diagnostics report. |
| `src/agent/index.ts` | (1) Explicit `provider === \"zai\"` branch in `createModel` using `ChatOpenAI` with the coding-plan base URL. (2) `installZaiStreamingNormalizer` — an api.z.ai-scoped fetch normalizer (installed only for the `zai` provider, with save/restore) that patches both the response-side role-less chunks and the request-side non-text content blocks. (3) `ZAI_API_KEY_ENV_KEY` added to the debug env dump. |
| `src/cli.tsx` | Add `zai` to the hardcoded provider list in the validation error message (a gap that the DeepSeek PR left out of date). |
| `src/credentials.tsx` | Add `zai` to `getProviderArticle`'s `\"a\"` branch (\"Z.AI\" is pronounced with a leading consonant sound). |
| `README.md` | Append Z.AI to the user-facing provider list. |

## Operator setup

No secrets in this PR. To run OpenWiki against Z.AI:

**\`~/.openwiki/.env\`:**

\`\`\`
OPENWIKI_PROVIDER=\"zai\"
OPENWIKI_MODEL_ID=\"glm-5.2\"
ZAI_API_KEY=\"<your coding-plan key>\"
\`\`\`

**Or in shell:**

\`\`\`bash
export OPENWIKI_PROVIDER=zai
export OPENWIKI_MODEL_ID=glm-5.2
export ZAI_API_KEY='<your coding-plan key>'
openwiki --init
\`\`\`

LangSmith tracing remains optional via `LANGSMITH_API_KEY`.

## Verification

- ✅ \`pnpm install --frozen-lockfile\`
- ✅ \`pnpm run build\` (tsc, strict)
- ✅ \`pnpm run lint:check\`
- ✅ \`pnpm run format:check\` (clean under \`--end-of-line auto\`; local \`format:check\` on Windows is affected by a pre-existing CRLF/autocrlf condition that also flags untouched files)
- ✅ \`--print\` smoke test against \`glm-5.2\` via \`api.z.ai/api/coding/paas/v4\` returns correct responses
- ✅ Network-level probe confirms requests hit \`api.z.ai/api/coding/paas/v4/chat/completions\`
- ✅ **Full \`--init\` agent loop (the previously crashing path) — 3 consecutive successful runs** (71 / 64 / 69 tool calls each), with both normalizers firing as expected. Keys sourced from env, not committed.

## Notes for reviewer

- The fetch normalizer is gated on \`provider === \"zai\"\`, installed with save/restore semantics (matching the existing \`installOpenRouterDebugFetch\` pattern), and only rewrites requests/responses whose URL contains \`api.z.ai\`. Other providers see the original \`globalThis.fetch\`.
- The response-side role backfill is structurally similar to DeepSeek's rewriter but targets a different layer (streaming response chunks vs. request body), because the Z.AI incompatibility is in stream assembly rather than content blocks.
- I did not touch any existing provider config, credential handling, or onboarding UI flow beyond the two single-line gating additions noted above.
- \`OPENAI_BASE_URL\` remains in \`deprecatedEnvKeys\` and is intentionally skipped on env load; adding a first-class provider is the supported way to add a new OpenAI-compatible endpoint.

Closes langchain-ai/openwiki#58

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,6 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 ## Customizing
 
-OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
+OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI, Anthropic, and Z.AI (GLM Coding Plan) out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
 
 If there's an inference provider or model you'd like to see added, please open a PR!
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -31,6 +31,8 @@ import {
   OPENWIKI_MODEL_ID_ENV_KEY,
   OPENWIKI_PROVIDER_ENV_KEY,
   resolveConfiguredProvider,
+  ZAI_API_KEY_ENV_KEY,
+  ZAI_BASE_URL,
   type OpenWikiProvider,
 } from "../constants.js";
 import {
@@ -71,6 +73,12 @@ export async function runOpenWikiAgent(
   emitDebug(options, `model=${modelId}`);
 
   const debugFetchCapture = installOpenRouterDebugFetch(options);
+  // Z.AI's streaming endpoint occasionally emits a final tool-call chunk with no
+  // assistant role, which langchain assembles into a ChatMessageChunk that
+  // deepagents rejects. The normalizer backfills the role so the agent loop
+  // (--init/--update) completes. Only installed for the zai provider.
+  const restoreZaiFetch =
+    provider === "zai" ? installZaiStreamingNormalizer(options) : null;
 
   try {
     return await runOpenWikiAgentWithModelFallbacks(
@@ -86,6 +94,7 @@ export async function runOpenWikiAgent(
     throw error;
   } finally {
     debugFetchCapture.restore();
+    restoreZaiFetch?.restore();
   }
 }
 
@@ -394,6 +403,16 @@ async function createModel(provider: OpenWikiProvider, modelId: string) {
       models,
       route: "fallback",
       siteName: "OpenWiki",
+    });
+  }
+
+  if (provider === "zai") {
+    return new ChatOpenAI({
+      apiKey: process.env[ZAI_API_KEY_ENV_KEY],
+      configuration: {
+        baseURL: ZAI_BASE_URL,
+      },
+      model: modelId,
     });
   }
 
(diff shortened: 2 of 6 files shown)


pohlipit23:
Closing in favor of #91 (by @AndrewJacop), which takes a cleaner approach to the same problem. Details in a review comment over there. Thanks!

PR #96, opened 2026-07-05
feat: implment AWS bedrock provider

### LINKED ISSUE #56: AWS Bedrock as a model provider
Hi, would it be possible to add AWS bedrock as a model provider? 
Supporting IAM roles, Bedrock API key, etc..?

Mohitmhatre32:
## Description
This PR adds support for **AWS Bedrock** as a model provider in OpenWiki. It allows developers to generate and maintain documentation using Bedrock's foundation models, such as Claude 3.5 Sonnet and Amazon Nova Pro, while maintaining flexibility for various AWS credential configurations (IAM roles, AWS profiles, and explicit credentials).

Resolves #56.

### 1. Configuration & Presets
* **`src/constants.ts`**:
  * Added `"bedrock"` to the `OpenWikiProvider` type and `SELECTABLE_OPENWIKI_PROVIDERS` list.
  * Added `"AWS_SECRET_ACCESS_KEY"` as the API key reference.
  * Pre-configured Bedrock model presets, including Claude 3.5 Sonnet (v1, v2, US cross-region), Claude 3.5 Haiku, and Amazon Nova Pro.

### 2. Interactive Onboarding & Environment
* **`src/credentials.tsx`**:
  * Extended the onboarding flow with specific steps for AWS: `"aws-region"`, `"aws-access-key"`, and `"aws-secret-key"`.
  * Updated validation so that credential keys are optional (can be skipped by pressing Enter) to natively support IAM roles, AWS profiles, or EKS/ECS task roles.
  * Masked input for `AWS_SECRET_ACCESS_KEY` during entry.
* **`src/env.ts`**:
  * Registered AWS environment variables (`AWS_REGION`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`) under `managedEnvKeys` to enable storage in local `~/.openwiki/.env`.
  * Added credential diagnostics for AWS parameters.

### 3. Model Initialization & Startup Validation
* **`src/agent/index.ts`**:
  * Implemented `ChatBedrockConverse` from `@langchain/aws` using dynamic imports to avoid loading AWS SDK modules when using other model providers.
  * Bypassed the API key verification in `ensureProviderKey` for Bedrock, enforcing only that `AWS_REGION` or `AWS_DEFAULT_REGION` is set.
* **`src/cli.tsx`**:
  * Updated startup commands and session validations to check for AWS region configuration instead of requiring provider API keys for Bedrock runs.

---

### 1. Build & Linting
* Verified compilation passes: `pnpm exec tsc -p tsconfig.json` compiles with zero errors.
* Verified code style: `pnpm run lint:check` completes with `0 errors`.

### 2. Manual Verification
* **Missing region validation**: Running without a region correctly prompts:
  `AWS_REGION or AWS_DEFAULT_REGION is required for non-interactive AWS Bedrock runs.`
* **IAM Role Authentication**: Running with a region but no local access keys successfully attempts AWS SDK authentication:
  ```bash
  OPENWIKI_PROVIDER=bedrock AWS_REGION="us-east-1" pnpm run dev -p "test"


rajpratham1 (approved):
This is a substantial and valuable feature. AWS Bedrock is a major provider, and the implementation covers model creation, CLI setup, credentials, environment handling, and dependencies.

Strengths
✅ Adds native AWS Bedrock support via ChatBedrockConverse.
✅ Validates AWS_REGION before execution.
✅ Supports IAM roles by making access/secret keys optional.
✅ Updates interactive credential flow for AWS-specific configuration.
✅ Adds required dependencies (@langchain/aws, AWS SDK).
✅ Updates provider list, environment diagnostics, and CLI.
Issues to address before merge
Incorrect API key abstraction
apiKeyEnvKey is set to AWS_SECRET_ACCESS_KEY.
Bedrock doesn't use a single API key; it uses AWS credentials (or IAM roles).
Consider special-casing Bedrock instead of forcing it into the apiKeyEnvKey abstraction.
Region message inconsistency
Some messages mention only AWS_REGION.
Validation accepts both AWS_REGION and AWS_DEFAULT_REGION.
Error/help text should consistently mention both.
Missing support for AWS credential chain
The implementation supports explicit credentials and IAM roles.
Verify it also works with the default AWS SDK credential provider chain (SSO, profiles, ECS/EKS, EC2 metadata), which many users rely on.
Tests
This PR introduces many code paths but no tests.
Add tests for:
Region validation.
IAM role flow.
Explicit credential flow.
Interactive setup.
Non-interactive validation.

SiddheshB-UBI:
Ran this against a real multi-turn tool-calling session (OpenWiki's own `--init` doc-generation flow, driving many `write_file`/`task` tool calls across turns against `ChatBedrockConverse` from `@langchain/aws`) and hit three issues worth fixing before this merges, independent of which credential-auth approach (IAM role/profile vs. bearer token) wins:

1. **Default `maxTokens` (4096) silently truncates large tool-call arguments.** OpenWiki's doc-generation loop makes `write_file` tool calls with full file contents as arguments — these routinely exceed 4096 tokens, and the truncated response either fails Zod schema validation (crashing the whole agent run, since `DynamicStructuredTool.call()` errors aren't caught by deepagents' tool wrapper) or silently drops file content. Needs an explicit `maxTokens` override in the `ChatBedrockConverse` constructor — we used `65536` against a Claude model with a 128000 max, and that fixed it.

2. **Default stream idle timeout (60s) is too short.** Hit `Bedrock Converse stream timed out after 60000 ms without receiving a chunk.` on a cold-ish response. `@langchain/aws`'s `stream_timeout.cjs` defaults `DEFAULT_STREAM_IDLE_TIMEOUT` to `6e4` — passing `streamIdleTimeout: 300000` (5 min) resolved it for us.

3. **Extended-thinking crash on multi-turn tool-call replay.** With a thinking-capable Claude model and thinking left enabled, replaying message history back to Bedrock Converse on a later turn throws:
   ```
   Error: 1 validation error detected: Value at 'messages.3.member.content.1.member.reasoningContent.reasoningText.text' failed to satisfy constraint: Member must not be null
   ```
   Reproduces with both `thinking.type: "enabled"` (`budget_tokens`) and `thinking.type: "adaptive"` (`output_config.effort`), and persists even with `disableStreaming: true`. We weren't able to root-cause this inside `@langchain/aws`'s history-replay/marshalling path — the only reliable workaround was disabling thinking entirely via `additionalModelRequestFields: { thinking: { type: "disabled" } }`. Since OpenWiki's agent loop is inherently multi-turn tool-calling, a Bedrock provider probably needs to either ship with thinking disabled by default or strip/repair `reasoningContent` blocks before replaying history, until this is properly root-caused (may be worth a separate issue against `@langchain/aws` itself).

Happy to help verify whichever of #96 / #40 moves forward once these are folded in.

colifran:
closing in favor of: https://github.com/langchain-ai/openwiki/pull/185

PR #97, opened 2026-07-05
feat: add --config CLI flag for interactive provider/model setup

hrid0yyy:
## Summary

Add `--config` CLI flag that opens an interactive wizard to change provider, model, and API key settings without running the main wiki session.

## Changes

- **`src/commands.ts`** — Added `config` command kind; parses `--config` flag; updated help text
- **`src/credentials.tsx`** — Added `forceConfig` prop to `InitSetup` so it always starts from provider selection step
- **`src/cli.tsx`** — Added `ConfigView` component that runs the wizard with `forceConfig`, shows a summary on completion, then auto-exits; wired up entry point to load env for config runs; fixed `command.dryRun` type guard

## What `--config` does

Running `openwiki --config` launches an interactive wizard that walks you through configuring OpenWiki:

1. **Provider** — Select from openrouter, baseten, fireworks, openai, or anthropic
2. **API key** — Enter the API key for the chosen provider
3. **Model** — Pick from the provider's preset models or type a custom model ID
4. **LangSmith** — Optionally set up LangSmith tracing

Each step is skipped if already configured. Changes are saved to `~/.openwiki/.env` and persist across sessions.

## Usage

```sh
openwiki --config                           # configure from scratch
openwiki --config --modelId claude-sonnet-5  # pre-select a model
```

### DIFF
--- a/src/cli.tsx
+++ b/src/cli.tsx
@@ -225,7 +225,7 @@ function App({ command }: AppProps) {
       return;
     }
 
-    if (command.dryRun) {
+    if (command.kind === "run" && command.dryRun) {
       process.exitCode = 0;
       app.exit();
       return;
@@ -412,6 +412,10 @@ function App({ command }: AppProps) {
         userMessage={command.userMessage}
       />
     );
+  }
+
+  if (command.kind === "config") {
+    return <ConfigView modelIdOverride={command.modelId} />;
   }
 
   if (shouldRunInteractiveCredentialSetup) {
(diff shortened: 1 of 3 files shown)


rajpratham1 (approved):
This is a useful UX improvement with a relatively small, well-contained implementation.

Pros
✅ Adds a dedicated --config command for updating provider, model, and API keys without triggering a normal run.
✅ Reuses the existing InitSetup flow instead of duplicating logic.
✅ Keeps backward compatibility (--init, --update, normal runs unchanged).
✅ Updates CLI help and examples.
✅ forceConfig cleanly starts at the provider selection screen.
Things to verify before merge
Ensure --config cannot be combined with incompatible flags like --init, --update, or a prompt. If it can, the parser should reject conflicting combinations.
Confirm the new summary screen exits correctly on all terminals (auto-exit after 3s and key press both work).
Add CLI tests for:
openwiki --config
openwiki --config --modelId ...
invalid combinations (--config --update, --config "prompt" if unsupported).

hrid0yyy:
I've addressed the review feedback and pushed the requested changes. I'd appreciate another review when you have time. Thank you!

colifran:
Thanks for the PR! This is going to be superseded by https://github.com/langchain-ai/openwiki/pull/374 which makes init idempotent so re-running it allows you to reconfigure things.

PR #102, opened 2026-07-05
feat: add OpenCode as a provider via @opencode-ai/sdk

Dado-hash:
Adds OpenCode as a native provider using the `@opencode-ai/sdk` instead of the OpenAI-compatible HTTP proxy.

**How it works:**
- Connects to a running OpenCode server (`opencode serve` or TUI with `--port 4096`) via the SDK
- Creates a session, sends prompts, and returns responses through LangChain's `BaseChatModel` interface
- No OpenAI-compatible proxy needed — uses OpenCode's native protocol

**Changes:**
- `src/agent/opencode-model.ts`: custom `BaseChatModel` subclass using `@opencode-ai/sdk`
- `@opencode-ai/sdk` added to `package.json` dependencies
- `createModel()` in `src/agent/index.ts` returns `OpenCodeModel` for `opencode` provider
- `OPENCODE_API_KEY` env key, type union, and config entry in `src/constants.ts`
- `OPENCODE_API_KEY` added to `managedEnvKeys` and credential diagnostics in `src/env.ts`
- Models: `opencode/big-pickle` (free) and `opencode/deepseek-v4-flash-free`
- Documentation updated across quickstart, CLI usage, architecture, workflow, operations, and README

**Usage:**
1. Start OpenCode server: `opencode serve --port 4096` (or TUI with `--port 4096`)
2. Select `opencode` as provider in OpenWiki and set `OPENCODE_API_KEY` (if auth is enabled)


pandamaci:
It is not working for me, I get "fetch failed" every run after about 1 minute after starting /init.
Consider changing the adapter to use OpenCode’s async/event flow ...

rajpratham1 (changes requested):
This PR is ambitious, but I wouldn't approve it yet. It introduces a completely new provider and ~3,700 lines of changes, yet several important pieces appear incomplete.

Main concerns

OPENCODE_API_KEY is never used

The provider adds OPENCODE_API_KEY_ENV_KEY everywhere (constants, env, docs), but OpenCodeModel creates the SDK client with only:
createOpencodeClient({
  baseUrl: this.baseUrl,
});

The API key is never passed or validated, making the new credential effectively unused.

Hardcoded OpenCode server URL

"http://127.0.0.1:4096"

This prevents users from running OpenCode remotely or on another port. It should be configurable through an environment variable or provider configuration.

New model bypasses existing provider flow
Every other provider integrates through the existing LangChain/OpenAI abstraction, while OpenCode introduces a custom BaseChatModel. That increases maintenance burden and should be justified with tests and documentation explaining why the existing architecture cannot be reused.
No tests
A new provider, new SDK integration, and a custom chat model are added, but there are no unit or integration tests covering:
session creation
prompt submission
response parsing
authentication failures
server unavailable cases
Large dependency addition
@opencode-ai/sdk plus a 3,600-line lockfile update is a significant dependency increase. There should be validation that this dependency is necessary and stable.
Good parts
Provider registration is consistent across:
constants
env handling
documentation
README
Documentation updates are thorough.
The implementation is cleanly separated into opencode-model.ts.

Dado-hash:
Thanks both for the reviews!

**Changes made:**

**@pandamaci** — The `fetch failed` should be fixed. The `OPENCODE_API_KEY` wasn't being passed to the SDK client (`createOpencodeClient` was missing the `auth` option). The server was returning an auth error that surfaced as `fetch failed`. Can you retry?

**@rajpratham1** — Addressed each concern:

1. **`OPENCODE_API_KEY` never used** → Fixed. Now passed via `auth: process.env[OPENCODE_API_KEY_ENV_KEY]` in `createOpencodeClient()`.

2. **Hardcoded URL** → Fixed. The base URL now comes from `PROVIDER_CONFIGS.opencode.baseURL` in `constants.ts`, making it configurable in one place.

3. **Custom model vs existing architecture** → OpenCode's server doesn't expose an OpenAI-compatible `/v1/chat/completions` endpoint (tested via curl — only the native SDK protocol works on `/session/{id}/message`). So a custom `BaseChatModel` wrapping `@opencode-ai/sdk` is required. If OpenCode adds OpenAI-compatible support in the future, we can switch to `ChatOpenAI` like the other providers.

4. **No tests** — The repository doesn't have tests for any provider currently. Happy to add them in a follow-up PR if that's a blocker.

5. **`@opencode-ai/sdk` dependency** — Required for the native protocol since there's no OpenAI-compatible fallback.

6. **`package-lock.json`** — Removed from tracking (project uses pnpm).

bracesproul:

PR #107, opened 2026-07-05
feat: add Ollama provider support with configurable num_ctx and base_url

arrase:
This PR adds native support for local models running under Ollama. It introduces configuration options for the context window length (num_ctx) and base URL, skips API key checks, and dynamically stringifies complex ToolMessage outputs to bypass ChatOllama strict types serialization restrictions.

In my opinion, the implementation using the native Ollama API is better than the implementation using the OpenAI API, which does not allow for configuring the context window size.: https://github.com/langchain-ai/openwiki/pull/87

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,14 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 ## Customizing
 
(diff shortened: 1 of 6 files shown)


bracesproul:

PR #110, opened 2026-07-05
feat: add an OpenAI-compatible provider with a required base URL

ppsplus-bradh:
## Summary

Adds a new **`openai-compatible`** provider that targets any OpenAI-compatible chat-completions endpoint via a user-supplied base URL. Unlike the existing OpenAI-compatible providers (`baseten` / `fireworks`), which hardcode their endpoints and ship curated model lists, this provider has **no default endpoint** and **no preset models** — so it can front arbitrary gateways and the upstream providers behind them. This is useful for OpenAI-compatible LLM endpoints such as those exposed by a LiteLLM gateway, which lets you reach whatever upstream providers the gateway fronts through a single OpenAI-shaped API.

> Stacked on #80 (alternative base URLs for the Anthropic provider). It reuses the `resolveProviderBaseUrl()` / `baseUrlEnvKey` mechanism added there. Until #80 merges, this PR's diff includes that commit; review the top commit (`feat: add an OpenAI-compatible provider…`) here.

## Changes

- **`src/constants.ts`**: add the provider to `OpenWikiProvider`, `SELECTABLE_OPENWIKI_PROVIDERS`, and `PROVIDER_CONFIGS` (`OPENAI_COMPATIBLE_API_KEY` + `OPENAI_COMPATIBLE_BASE_URL`, custom-only model list); add a `requiresBaseUrl` flag to `ProviderConfig` plus `providerRequiresBaseUrl()`, `getProviderBaseUrlEnvKey()`, and `isValidBaseUrl()` helpers.
- **`src/agent/index.ts`**: `ensureProviderBaseUrl()` aborts a run early with a clear message when a `requiresBaseUrl` provider has no base URL. The provider itself flows through the existing `ChatOpenAI` + `resolveProviderBaseUrl()` path (no new client branch).
- **`src/credentials.tsx`**: interactive onboarding gains a **base-URL step** for providers that require one, validated as an `http(s)` URL, wired into the existing step machine and `~/.openwiki/.env` persistence.
- **`src/env.ts`**: persist and diagnose the new keys; the base URL is treated as a non-secret value (shown in full in diagnostics).
- **Docs**: README + OpenWiki docs describe the provider, including reaching upstream providers through an OpenAI-compatible gateway.

## Usage

```bash
OPENWIKI_PROVIDER=openai-compatible
OPENAI_COMPATIBLE_API_KEY=<gateway key>
OPENAI_COMPATIBLE_BASE_URL=https://<gateway>/v1
OPENWIKI_MODEL_ID=<model name the gateway exposes>
```

Model IDs are passed through verbatim, so gateway-specific aliases work as-is.

## Testing

- `pnpm build`, `pnpm lint:check`, and `pnpm format:check` pass.
- Verified end-to-end against a local build:
  - With `OPENAI_COMPATIBLE_BASE_URL` unset, a run aborts with `OPENAI_COMPATIBLE_BASE_URL is required to run OpenWiki with OpenAI-compatible.`
  - With it set to a local HTTP server, the client issued `POST /v1/chat/completions` to the configured base URL (OpenAI Chat Completions format), confirming requests are routed to the gateway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,4 +72,18 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 OpenWiki supports OpenRouter, Fireworks, Baseten, OpenAI and Anthropic out of the box. By default, there are a few models pre-defined (GLM 5.2, Kimi K2.6, Sonnet 5, etc) but for each inference provider, OpenWiki will allow you to specify your own custom model ID.
 
+### Alternative base URLs
+
+To route the Anthropic provider at an alternative, Anthropic-compatible endpoint
+(for example a self-hosted or proxied gateway) instead of the default API, set
+`ANTHROPIC_BASE_URL` alongside `ANTHROPIC_API_KEY`:
+
+```bash
+OPENWIKI_PROVIDER=anthropic
+ANTHROPIC_API_KEY=your-key
+ANTHROPIC_BASE_URL=https://your-gateway.example.com/anthropic
+```
+
+The base URL can be set in your environment or stored in `~/.openwiki/.env`.
+
 If there's an inference provider or model you'd like to see added, please open a PR!
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ These configuration options and secrets will be saved to `~/.openwiki/.env` on y
 
 ## Customizing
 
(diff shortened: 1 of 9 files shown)


bracesproul (approved):
nice! thanks a ton, this looks great.

pushed one small commit.

do you have a twitter i can mention when i tweet about this new capability? would love to give you credit!

ppsplus-bradh:
> nice! thanks a ton, this looks great.
> 
> pushed one small commit.
> 
> do you have a twitter i can mention when i tweet about this new capability? would love to give you credit!

@bradhuffman on twitter/x.  Thanks!

Issue #101, opened 2026-07-05
The current Node.js LTS version 'v24.16.0' causes issues


hugobepa:
The current Node.js LTS version 'v24.16.0' causes issues

- error :

````
node.exe : npm warn deprecated prebuild-install@7.1.3: No longer maintained. Please contact the author of the relevant native addon; alternatives are available.
In C:\Users\User\scoop\apps\nodejs-lts\current\bin\npx.ps1: 24 Character: 5

npm verbose os Windows_NT 10.0.26200
npm verbose node v24.16.0
npm verbose npm  v11.17.0
npm verbose exit 1
npm verbose code 1
````

- failed local solutions :
````
npx -p node@20 -p openwiki openwiki
npx -p node@22 -p openwiki openwiki
npx --yes -p node@20 -p openwiki openwiki
npx --yes -p node@22 -p openwiki openwiki
````

- only possible global solution is to use a maximum Node.js version 22

   - using Node.js directly
   - quick version switcher package
    - [gh-repo_nvm-windows](https://github.com/coreybutler/nvm-windows)
    - [wd_nvm](https://www.nvmnode.com/es/)
    - steps to follow :
      **The most important step before installing `nvm-windows` is to uninstall any previously installed version of Node.js.** Leaving previous versions can cause conflicts and issues when switching versions.

      To uninstall:
      1.  Go to "Control Panel" > "Programs and Features" and uninstall Node.js.
      2.  As an additional measure, manually delete any residual folders, such as `C:\Program Files\nodejs` or `C:\Users\YourUser\AppData\Roaming\npm`.

      ### 📥 Step 1: Download the installer

      1.  Visit the **releases** page of the project on GitHub: [https://github.com/coreybutler/nvm-windows/releases](https://github.com/coreybutler/nvm-windows/releases).
      2.  Look for the `nvm-setup.zip` file (which is the graphical installer version, the most recommended) and download it.

      ### 🛠️ Step 2: Run the installer

      1.  Unzip the downloaded `.zip` file.
      2.  Right-click on `nvm-setup.exe` and select **"Run as administrator"** to avoid permission issues. If you don't see that option, you can run it directly, but it is better to run it as administrator.

      ```bash
        nvm version
        nvm node_mirror https://npmmirror.com/mirrors/node/
        nvm npm_mirror https://npmmirror.com/mirrors/npm/
        nvm install 22.14.0
        nvm use 22.14.0
        npm install -g openwiki
      ```

      | Command | Description |
[...]

PR #111, opened 2026-07-05
docs: clarify recommended Node.js version

godlovesmei:
## What changed
- Added a note about the recommended Node.js version.
- Clarified that users experiencing issues with Node.js 24 can switch to Node.js 22 and reinstall OpenWiki.

## Why
This relates to #101, where Node.js 24 was reported to cause installation/runtime issues.

## Testing
- Documentation only. No code changes.

Related to #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ OpenWiki is a CLI that writes and maintains documentation for your codebase, bui
 npm install -g openwiki
 ```
 
+OpenWiki requires Node.js 20 or newer. The project currently runs CI on Node.js 22. If you encounter installation or runtime issues with Node.js 24, switch to Node.js 22 and reinstall OpenWiki.
+
 ## Quick Start
 
 Initialize OpenWiki, configure your model and API key, then generate documentation


README.md
```diff
@@ -10,6 +10,8 @@ OpenWiki is a CLI that writes and maintains documentation for your codebase, bui
 npm install -g openwiki
 ```
 
+OpenWiki requires Node.js 20 or newer. The project currently runs CI on Node.js 22. If you encounter installation or runtime issues with Node.js 24, switch to Node.js 22 and reinstall OpenWiki.
```
bracesproul, line 13: did you check if 24 has issues? if so we should just fix them

bracesproul (changes requested):

bracesproul:
closing as i dont want to just say "if you have issues switch versions" - if there are issues they should be fixed

godlovesmei:
Thanks, that makes sense. I dropped the workaround wording. I verified Node 24 directly and opened a smaller CI PR to test compatibility instead of documenting it as a caveat: #184

If the Node 24 path fails in CI, I’ll follow up by fixing the underlying failure rather than recommending a version switch.

PR #113, opened 2026-07-05
fix: use dash-delimited Anthropic model id for Opus (claude-opus-4-8)

Idan-Shperling:
## Problem

The Anthropic provider's **Opus** option is `claude-opus-4.8` (dotted), but Anthropic API model ids are **dash-delimited**. Selecting Opus with `OPENWIKI_PROVIDER=anthropic` fails immediately:

```
Error 404 not_found_error  model: claude-opus-4.8
```

(`claude-haiku-4-5` and `claude-sonnet-5` in the same list are already dashed and work.)

## Fix

One-character data fix in `src/constants.ts`: `claude-opus-4.8` → `claude-opus-4-8`.

The OpenRouter entry (`anthropic/claude-opus-4.8`) is intentionally left as-is — OpenRouter slugs use dotted versions.


bracesproul (approved):
good catch! thank you

PR #115, opened 2026-07-05
fix: skip binary file reads instead of crashing the run

### LINKED ISSUE #114: v0.0.1 (Anthropic provider) crashes on any Python repo with __pycache__: binary file reads become invalid document blocks; one malformed tool result kills the run
## Summary

`openwiki --init` (v0.0.1, Anthropic provider) reliably crashes on Python repos that contain a populated `__pycache__/` — i.e., effectively any Python repo that has ever been run. The agent reads binary files during repo exploration, the tool result gets packaged as a `document` content block with a non-PDF `media_type`, and the Anthropic API rejects the request with a 400. Because a single malformed tool result kills the entire run (as described in #86), generation never completes and no wiki is written.

## Environment

- openwiki 0.0.1 (npm global install)
- macOS (Apple Silicon), Node 26.x
- Provider: `anthropic`, `OPENWIKI_MODEL_ID=claude-sonnet-5` (configured via `~/.openwiki/.env`)

## Reproduction

1. Any Python repo with compiled bytecode present (`__pycache__/*.pyc`). Ours was a ~7.6K-LOC Flask project; nothing exotic.
2. `openwiki --init -p "Generate the initial OpenWiki documentation for this repository"`
3. Run dies after ~1–2 minutes, exit code 1, nothing written.

## Observed

Run 1 (repo also contained a PNG):

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.26.content.4.tool_result.content.0.document.source.base64.media_type: Input should be 'application/pdf'"}}

```

Run 2 (PNG removed from the tree to isolate the trigger — identical failure, further along, with `__pycache__/*.pyc` as the only binaries left):

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.36.content.1.tool_result.content.0.document.source.base64.media_type: Input should be 'application/pdf'"}}
```

So the trigger is any binary file read (bytecode, images, …), not one specific file type — the agent's file-read tool result for binary content is serialized as a `document` block whose `media_type` isn't `application/pdf`, which the Anthropic Messages API rejects.

## Expected

Either (ideally both):

1. **Skip binary files during repo exploration** — e.g., extension/magic-byte filter, and/or respect `.gitignore` (which excludes `__pycache__/` in virtually every Python repo).
2. **Survive a malformed tool result** — drop or stub the offending block and continue, rather than letting one bad read abort the whole generation (same failure class as #86; this issue adds a deterministic, near-universal trigger for it).

Happy to provide more detail if useful.

johnfcarter:
Fixes #114 (and removes the most common trigger of the crash class described in #86).

**Problem.** `FilesystemBackend.read()` returns raw `Uint8Array` content for any file whose MIME type isn't text. Non-image binary content is then serialized as a base64 `document` content block, and the Anthropic Messages API only accepts `application/pdf` there — so a single read of e.g. `__pycache__/*.pyc` fails the request with a 400 and aborts the entire run. Since virtually every Python repo contains bytecode, `openwiki --init` reliably crashes on Python projects with the Anthropic provider.

**Fix.** `OpenWikiShellBackend` (new, extends `LocalShellBackend`) overrides `read()` to return a short text placeholder for binary content — `Binary file skipped: <path> (<mime>, <bytes>). OpenWiki reads text sources only.` Wiki generation is a text task; the agent notes the file and moves on. Text reads pass through unchanged.

Scoped here rather than in `deepagents` deliberately: OpenWiki never needs raw bytes, so this is correct for OpenWiki regardless of whether deepagents later adds native binary handling upstream (this wrapper stays harmless if it does).

**Verification.**
- The repro from #114 (Python repo, 26 `.pyc` + 1 `.png`): unpatched crashed 2/2 runs within ~2 min; patched completes `--init` end-to-end in ~5 min with a full wiki generated.
- Unit-checked the override directly: `.pyc` → placeholder, `.png` → placeholder, `README.md` → unchanged passthrough.
- `pnpm lint:check`, `pnpm format:check`, `pnpm build` all clean.

Happy to adjust scope (e.g., pass through Anthropic-supported image types) if you'd prefer.

### DIFF
--- a/src/agent/backend.ts
+++ b/src/agent/backend.ts
@@ -1,14 +1,37 @@
+import { Buffer } from "node:buffer";
 import { LocalShellBackend, type ReadResult } from "deepagents";
+
+const TEXT_MIME_TYPES = new Set([
+  "application/json",
+  "application/javascript",
+  "application/typescript",
+  "application/xml",
+  "image/svg+xml",
+]);
+
+const BINARY_MIME_TYPES = new Set([
+  "application/gzip",
+  "application/octet-stream",
+  "application/pdf",
+  "application/wasm",
+  "application/x-7z-compressed",
+  "application/x-gzip",
+  "application/x-rar-compressed",
+  "application/x-sqlite3",
+  "application/x-tar",
+  "application/zip",
+]);
+
+const BINARY_MIME_PREFIXES = ["audio/", "font/", "image/", "video/"];
 
 /**
  * LocalShellBackend that never returns raw binary content to the model.
  *
  * The base FilesystemBackend returns the full Uint8Array for any file whose
  * MIME type is not text (for example `__pycache__/*.pyc`, images, sqlite
- * files). Downstream, non-image binary content is serialized as a base64
- * `document` content block, and the Anthropic Messages API only accepts
- * `application/pdf` there — so a single binary read fails the request with
- * a 400 and aborts the entire run:
+ * files). Downstream, unsupported binary or multimodal content blocks can be
+ * rejected by providers such as Anthropic or OpenAI, so a single binary read
+ * can abort the entire run:
  * https://github.com/langchain-ai/openwiki/issues/114
  *
  * OpenWiki documents codebases, which is a text task; it never needs raw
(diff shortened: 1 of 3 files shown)


rajpratham1 (changes requested):
The fix addresses a real issue—preventing binary file reads from crashing documentation runs—but I don't think it's complete enough to merge in its current form.

What's good
✅ Small, focused change.
✅ Fixes the immediate Anthropic/OpenAI binary-content failure.
✅ Encapsulates the behavior in a custom backend instead of scattering checks.
✅ Keeps the agent running instead of aborting.
Blocking issues
1. No tests

This is the biggest blocker.

The PR changes core filesystem behavior but adds zero regression tests.

At minimum I'd expect tests covering:

reading a PNG
reading a .pyc
reading a normal text file
verifying Uint8Array becomes a text placeholder
verifying normal text files remain unchanged

Without tests, this is easy to regress.

2. Only checks Uint8Array

Current logic:

if (result.content instanceof Uint8Array)

This assumes every binary read is returned as a Uint8Array.

If deepagents later changes its representation (e.g. Buffer, stream, or another binary type), the fix silently stops working.

The implementation relies on an internal detail of LocalShellBackend.

3. Doesn't handle non-text MIME with string content

Some extensionless files or unusual MIME detections may return:

content: string
mimeType: application/octet-stream

Those can still become non-text content blocks depending on backend behavior.

Checking only the content type may not cover every problematic case.

4. Missing provider-independent explanation

The comment references Anthropic's API limitations, but this actually benefits multiple providers that reject unsupported multimodal content. Framing the fix more generally would make the intent clearer.

Minor suggestion

Including the file size in the placeholder is useful:

Binary file skipped: foo.db (application/octet-stream, 32768 bytes)

That provides helpful context without exposing binary data.

johnfcarter:
Thanks for the thorough review — all four addressed in the follow-up commit:

1. **Tests** — added a vitest suite (`src/agent/backend.test.ts`, 12 cases): PNG, `.pyc`, and normal source via real temp-file reads through the backend, plus the representation variants (Uint8Array, ArrayBuffer view, string-with-binary-MIME) and passthrough for text/JSON/XML/JS/TS/undefined MIME. Added minimal vitest config + a `tsconfig` exclude so tests stay out of the build. (Pinned vitest to 2.1.9 to stay under the repo's `node >=20` floor — happy to bump if you'd prefer.)
2. **Beyond `Uint8Array`** — detection now covers `Uint8Array` (Buffer included, since it subclasses `Uint8Array`), `ArrayBuffer`, and any `ArrayBuffer.isView`.
3. **String + binary MIME** — extensionless / `application/octet-stream` string reads now hit the placeholder too; text, JSON/XML/JS/TS, `+json`/`+xml`, and undefined MIME still pass through unchanged (conservative — never placeholders a real source file).
4. **Provider-independent framing** — comment reworded to cover any provider that rejects unsupported binary/multimodal blocks (Anthropic, OpenAI), issue link kept.

On the size suggestion — the placeholder already includes byte length (`Binary file skipped: <path> (<mime>, N bytes)`), so that's covered.

johnfcarter:
Rebased onto latest main — dropped the vitest 2.1.9 pin and local test config in favor of the vitest 4 setup main now ships, and moved the tests to `test/` per repo convention. Diff is now just the fix + tests; all 98 tests pass.

colifran:
Hey @johnfcarter - thanks for your work here. I think an upstream fix in deepagentsjs is actually what is needed to resolve: https://github.com/langchain-ai/deepagentsjs/pull/656. I think this should fix the issue.

johnfcarter:
Confirmed — deepagents 1.10.6 fixes it. Re-ran the exact repro that crashed 2/2 on v0.0.1 (Python repo with `__pycache__`, Anthropic provider, claude-sonnet-5): openwiki 0.0.2 + deepagents 1.10.6 completed a full `--init` clean, no 400s. #114 can be closed too.

One residual worth noting: genuine binaries with unmapped extensions (.sqlite, .zip, .so) now read as raw bytes into the agent's context rather than being skipped — not a crash anymore, just token noise on repos with stray binaries. Happy to rework this PR into a placeholder-on-binary-content guard layered on the text/plain default if there's interest; otherwise I'll close. Thanks @colifran.

colifran:
@johnfcarter  - I think that would be helpful to give the agent something other than random token noise. We can take on more specific mappings upstream in deepagentsjs too as needed. If you want to work on reworking this PR into a placeholder-on-binary-content guard I'd be happy to review!

colifran:
@johnfcarter - closing for now. if you're interested in implementing the placeholder-on-binary-content guard please open up a new pr!

PR #116, opened 2026-07-05
fix: add config-only credential setup

terminalchai:
Adds a config-only CLI entry point for re-running provider, API key, model, and optional LangSmith setup without starting an agent run. This gives users a recovery path when saved credentials are expired, exhausted, or otherwise need to be replaced.\n\nFixes #23.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,13 @@ Show help:
 openwiki --help
 ```
 
+Reconfigure your saved provider, API key, model, and optional LangSmith key
+without starting a documentation run:
+
+```sh
+openwiki --config
+```
+
 `openwiki` creates initial documentation in `openwiki/` when no wiki exists. If `openwiki/` already exists, it refreshes that documentation from repository changes. By default, the CLI stays open after each run so you can send follow-up messages. Use `-p` or `--print` for a one-shot non-interactive run that prints the final assistant output.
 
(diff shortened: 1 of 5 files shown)


rajpratham1 (approved):
This is a focused, user-facing improvement that introduces a proper --config mode instead of overloading the initial setup flow. The behavior is consistent, the CLI UX is clearer, and the documentation is updated alongside the implementation.

What's good
✅ Introduces a dedicated --config command instead of piggybacking on normal startup.
✅ Prevents accidental agent execution after configuration.
✅ Rejects invalid usage like combining --config with other arguments.
✅ Handles non-interactive terminals with a clear error.
✅ Updates README and CLI usage documentation.
✅ forceReconfigure cleanly reuses the existing setup UI rather than duplicating logic.
Minor suggestion

The implementation now has several branches like:

if (forceReconfigure) {
  return "provider";
}

inside helper functions (getInitialStep, getNextStepAfterProvider, getNextStepAfterApiKey). Since the reconfigure flow is becoming a distinct mode, you could consider centralizing this behavior in a small state machine or a dedicated flow function. It would reduce repeated forceReconfigure checks and make future setup changes easier to maintain.

colifran:
This is going to be superseded by https://github.com/langchain-ai/openwiki/pull/374 which makes init idempotent so you can re-run to reconfigure things.

PR #117, opened 2026-07-05
feat: add Langfuse tracing as an alternative to LangSmith

vsilv:
## Summary

OpenWiki can currently trace runs only to LangSmith (via the LangChain SDK's built-in env-based tracer). This PR adds **Langfuse** as an optional, independent alternative. It works identically for managed (Langfuse Cloud) and self-hosted instances — the only difference is `LANGFUSE_BASE_URL`.

Langfuse v5 is OpenTelemetry-based, so unlike LangSmith it needs a small amount of wiring rather than pure env passthrough.

## What changed

- **`src/tracing.ts` (new)** — registers a Langfuse OpenTelemetry span processor (`@langfuse/otel`) on a `NodeTracerProvider` and attaches the `@langfuse/langchain` `CallbackHandler` to the agent's `.stream()` callbacks, so the full LangGraph run tree (model + tool + subagent spans) is captured. It is a **no-op unless `LANGFUSE_PUBLIC_KEY` and `LANGFUSE_SECRET_KEY` are set** (a blank value disables it, mirroring the LangSmith behavior), and it flushes buffered spans after each run so short-lived `--print` / CI runs export their traces before the process exits.
- **Setup wizard (`src/credentials.tsx`)** — an optional Langfuse step (public key, secret key, optional base URL) after the LangSmith step, following the same pattern. A blank public key skips/disables it.
- **Env, diagnostics, header, redaction** — `src/env.ts`, `src/cli.tsx`, and `src/agent/index.ts` learn about the three Langfuse keys: managed in `~/.openwiki/.env`, surfaced in credential diagnostics (secret key masked, base URL shown), shown in the header tracing indicator, and redacted from error diagnostics.
- **Config** — `LANGFUSE_BASE_URL` is optional and defaults to Langfuse Cloud; the same code path serves managed and self-hosted deployments.
- **Docs** — README (a `Tracing` section covering both LangSmith and Langfuse), `DEVELOPMENT.md`, and the wiki pages.

Langfuse and LangSmith are fully independent; enabling one has no effect on the other.

## Configuration

```bash
LANGFUSE_PUBLIC_KEY=your-langfuse-public-key
LANGFUSE_SECRET_KEY=your-langfuse-secret-key
# Optional; defaults to Langfuse Cloud. Point at your own instance to self-host.
LANGFUSE_BASE_URL=https://cloud.langfuse.com
```

## Testing

- Built, linted, and formatted (`pnpm build`, `pnpm lint:check`, `pnpm format:check` all pass; lockfile updated for `--frozen-lockfile` CI).
- Verified end-to-end against **Langfuse Cloud** and a **self-hosted Langfuse** instance: an `openwiki --print` run produces a trace with the full LangGraph span tree, including the LLM `GENERATION`. Same result on both; only the base URL differed.
- Verified the **off path** (no Langfuse env): runs are unchanged and no tracer is registered.
- Verified the credential data layer (gating, `~/.openwiki/.env` round-trip, secret masking in diagnostics).

## New dependencies

`@langfuse/langchain`, `@langfuse/otel`, `@opentelemetry/sdk-trace-node`, `@opentelemetry/api`.


bracesproul:
closing as we won't support this - just langsmith for openwiki

PR #118, opened 2026-07-06
Fix/skip binary reads in documentation backend

pawel-twardziak:
## Summary

Fixes the `Error 400 Invalid value: 'file'. Supported values are: 'text', 'refusal', 'image_url', and 'input_audio'` that aborts OpenWiki runs on OpenAI and other OpenAI-compatible providers (openai, baseten, fireworks, openai-compatible, openrouter). The Anthropic provider was unaffected, which is why it only surfaced after switching providers.

## Root cause

OpenWiki itself only ever sends a plain-string user message, so the offending content originates downstream in deepagents:

- deepagents' built-in `read_file` tool returns a standard multimodal content block for any **non-text** file - `type: "file"` for binaries, `image`/`audio`/`video` for media (`deepagents/.../langsmith-wdF8zG42.js`, `createReadFileTool`).
- Its text detection is narrow: only `text/*`, `application/json`, `application/javascript`, and `image/svg+xml` count as text. Every unknown extension maps to `application/octet-stream`, so extension-less files like `LICENSE`, `Dockerfile`, `Makefile`, plus `.lock`, `.ico`, `.pdf`, `.woff2`, etc. all become `file` blocks.
- That `read_file` result becomes a `ToolMessage` with no `output_version: "v1"` marker, so `@langchain/openai` takes its legacy conversion path and forwards the `type: "file"` block **verbatim** into the `/chat/completions` request.
- OpenAI's Chat Completions API rejects a `file` content part in that position, returning the 400 above.

The final conversion link was verified empirically by running the installed `@langchain/openai@1.5.3` converter against a `ToolMessage` carrying the exact block `read_file` produces - it passed the raw `{ type: "file", mimeType, data }` straight through to the request payload.

## Fix

New `src/agent/backend.ts` wraps the documentation backend's `read()`:

- `createDocumentationBackend(cwd)` builds the `LocalShellBackend` and intercepts `read()` so any non-text result (binary `Uint8Array` content, or a non-text MIME type) is downgraded to a short **text placeholder** with `mimeType: "text/plain"`.
- This guarantees `read_file` always yields a `text` block, so no `file`/`image` block can ever reach the provider. It fixes all providers at once rather than steering users to Anthropic.
- `isTextMimeType()` mirrors deepagents' own predicate so exactly the same set of files is intercepted.
- `src/agent/index.ts` now calls `createDocumentationBackend(cwd)` instead of constructing `LocalShellBackend` inline; the now-unused `LocalShellBackend` import is removed.

OpenWiki documents source text only, so binary assets never needed to reach the model in the first place.

## Testing

- `test/backend.test.ts` (new): unit tests for `isTextMimeType` and `toTextOnlyReadResult` (binary content, non-text MIME with string content, genuine text pass-through, error pass-through), plus an end-to-end regression test that reads a real `.png`, an extension-less `LICENSE`, and a `README.md` through an actual `LocalShellBackend` and asserts non-text files come back as `text/plain` strings while text files read normally.
- `npx tsc -p tsconfig.json --noEmit` - passes.
- `npx vitest run` - 13/13 passing (6 existing + 7 new).
- `npx eslint` on changed files - clean.

## Breaking changes

None. Behavior only changes for reads of non-text files, which previously crashed the run on affected providers.

### DIFF
--- a/src/agent/backend.ts
+++ b/src/agent/backend.ts
@@ -1,0 +1,106 @@
+import { LocalShellBackend, type ReadResult } from "deepagents";
+
+/**
+ * deepagents' `read_file` tool builds an `image`/`audio`/`video`/`file` standard
+ * content block for any file its backend reports as non-text. That `type: "file"`
+ * content part is forwarded by `@langchain/openai` to `/chat/completions`, and
+ * OpenAI (plus other OpenAI-compatible providers) reject it with:
+ * `Invalid value: 'file'. Supported values are: 'text', 'refusal', 'image_url', and 'input_audio'.`
+ *
+ * The backend returns a `string` for files it already considers text and a
+ * `Uint8Array` for everything else (binaries, media, and any unrecognized or
+ * extension-less path such as `LICENSE`/`Dockerfile`, which map to
+ * `application/octet-stream`). We only need to handle that second case: decode
+ * the bytes as UTF-8 and, when they are genuinely text, hand the real content
+ * back as `text/plain` so `read_file` emits a `text` block instead of a `file`
+ * block. Only truly binary bytes fall back to a placeholder.
+ *
+ * Keying off the content shape (string vs. bytes) rather than re-deriving the
+ * MIME classification avoids duplicating deepagents' text-type list, and using
+ * the decoded bytes rather than the MIME label preserves plain-text files that
+ * deepagents mislabels as binary.
+ */
+
+// Matches deepagents' own binary read cap; skip decoding anything larger.
+const MAX_DECODE_BYTES = 10 * 1024 * 1024;
+const PLACEHOLDER_MIME_TYPE = "text/plain";
+const NUL = String.fromCharCode(0);
+
+function decodeUtf8Text(bytes: Uint8Array): string | null {
+  if (bytes.byteLength > MAX_DECODE_BYTES) {
+    return null;
+  }
+
+  let text: string;
+  try {
+    text = new TextDecoder("utf-8", { fatal: true }).decode(bytes);
+  } catch {
+    return null;
+  }
+
+  // A NUL byte is valid UTF-8 but is a strong signal of binary data (and of
+  // UTF-16/UTF-32 text, which we deliberately treat as binary here).
+  return text.includes(NUL) ? null : text;
+}
+
+/**
+ * Normalize a backend read so it can never yield a non-text content block:
+ * genuine text is preserved (decoded when necessary), binary bytes become a
+ * short text placeholder, and errors pass through untouched.
+ */
+export function toTextOnlyReadResult(
+  filePath: string,
+  result: ReadResult,
+): ReadResult {
+  if (result.error !== undefined) {
+    return result;
+  }
+
+  // Already text: deepagents only returns string content for files it treats as
+  // text, so `read_file` will emit a text block unchanged.
+  if (typeof result.content === "string") {
+    return result;
+  }
+
+  if (ArrayBuffer.isView(result.content)) {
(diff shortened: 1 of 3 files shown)


pawel-twardziak:
Hi @bracesproul I've been facing the error, so here is a potential fix

rajpratham1 (changes requested):
The idea is good—prevent binary reads from reaching providers that reject file/image content blocks—but I don't think this implementation is ready to merge.

What's good
Nice separation into backend.ts.
Reusable helper functions.
Good regression tests.
Keeps documentation runs text-only.
Much cleaner than scattering checks throughout the agent.
Issues
1. isTextMimeType duplicates deepagents logic

The PR says:

mirrors deepagents' own isTextMimeType

That means OpenWiki now has a copy of another project's implementation.

If deepagents adds support for another MIME type:

application/xml
text/x-python
application/x-yaml

OpenWiki silently becomes outdated.

This duplication is a maintenance risk.

2. Extensionless files are treated as binary

This is the biggest concern.

The PR intentionally converts files like

LICENSE
Dockerfile
Makefile
Procfile
.gitignore
.env.example

into

Binary content is not read.

simply because deepagents labels them as

application/octet-stream

Those files are plain text and often contain critical documentation context.

That means the model loses valuable repository information.

3. Placeholder discards useful content

Instead of

Dockerfile
FROM node:22
RUN ...

the model only receives

Skipped non-text file...

This reduces repository understanding.

A better fix would detect whether the bytes are valid UTF-8 before discarding them.

4. Assumes MIME type is authoritative

Current logic:

mime != text
→ skip

Reality:

MIME detection is imperfect.
Extensionless files are common.
Generated files are frequently mislabeled.

The decision should consider the actual content as well.

5. Wrapper may hide future backend behavior

Replacing

backend.read

with a wrapper is simple, but if LocalShellBackend.read() later changes its return shape or semantics, this wrapper could become stale.

6. Tests miss important cases

I'd also expect tests for:

SVG files
UTF-8 extensionless files
UTF-16 text
XML
YAML
huge binaries
empty files
files without MIME type
Minor

Ignoring

.remember

and

.idea

is unrelated to the bug.

These should probably be a separate PR.

Recommendation

I would instead use logic like:

if (content is string)
    return it

if (bytes decode as UTF-8)
    return decoded text

otherwise
    return placeholder

This preserves real text while still preventing binary content blocks.

pawel-twardziak:
hi @rajpratham1 

thanks for the thorough review!  These re good points imo and I've pushed a revision that address them.

2 & 3 the wrapper no longer discards non-text reads. It now decodes the bytes as utf-8 and, when they are valid text, returns the real content as text/plain. So plain text files deepagents mislabels as application/octet-stream keep their contents and reach the model. Only genuinely binary bytes fall back to a placeholder

1- the decision is now based on the shape of the read result plus utf-8 validity, not on re-deriving the mine classification

4 - changed

5 - now it depends only on the public ReadResult shape, not on any internal backend behavior, so it shuld be resilient to internal changes

6 - added utf-8 extension-less files etc.

deepagents read_file chooses a text vs file/image block based on the result's mime type, not its content, so a decode string left with application/octet-stream would still be emitted as a file block. the fix therefore also sets mimeType: text/plain on decoded/placeholder results so read_file always takes its text path

colifran:
Hey @pawel-twardziak. I think the fix for this is upstream in deepagentsjs: https://github.com/langchain-ai/deepagentsjs/pull/656. I think this should resolve the issue.

pawel-twardziak:
hi @colifran I think you are right, I just didn't want to touch deepagents - I'm waiting for the PR to be merged

colifran:

PR #127, opened 2026-07-06
feat: add z.ai and local LLM providers (Ollama, LM Studio, vLLM)

notdezzi:


100yenadmin:
I tested the LCO pilot against stock OpenWiki 0.0.2 with Z.AI/GLM and wanted to avoid duplicating this broader provider PR.

Verified stock configuration:

```sh
OPENWIKI_PROVIDER=anthropic
OPENWIKI_MODEL_ID=GLM-5.2
ANTHROPIC_BASE_URL=https://api.z.ai/api/anthropic
ANTHROPIC_API_KEY=<Z.AI key>
```

That works because Z.AI exposes an Anthropic-compatible endpoint. So for LCO we can use stock OpenWiki for now without a local provider patch.

I still think a first-class `zai` provider alias could be useful, but this PR currently bundles several provider families and is marked conflicting. I am not planning to open a duplicate broad provider PR while #127 is open. If maintainers want this split out later, I can send a focused alias-only PR with `ZAI_API_KEY`, optional `ZAI_BASE_URL`, default base `https://api.z.ai/api/anthropic`, and default model `GLM-5.2`.

colifran:
I think we already support these via openai-compatible provider. If I'm incorrect please feel free to re-open. If you re-open I would ask for a clean split of one provider per PR as per the contributing guidelines.

Issue #126, opened 2026-07-06
refactor: scope DeepSeek fetch rewriter and add tests


himanshu231204:
## Context

PR #68 adds DeepSeek as a first-class provider. A review from @rajpratham1 raised concerns about the global fetch rewriter implementation.

### 1. Global fetch wrapper scope too broad

The current implementation installs a **process-wide** globalThis.fetch wrapper that intercepts ALL requests to pi.deepseek.com. This could have unintended side effects for other code using fetch.

**Current location:** \src/agent/index.ts\ — \installDeepSeekFetchRewriter()\

**Requested change:** Scope the rewriter closer to the model/client layer instead of globally. Options:
- Pass a custom \etch\ to the \ChatOpenAI\ constructor (LangChain supports this)
- Wrap the fetch at the HTTP client level used by the DeepSeek model instance
- Use a request interceptor scoped to the model instance only

### 2. Missing tests for request rewriting

No test coverage exists for the fetch rewriter behavior. Need tests to verify:
- Array content blocks are flattened to text for DeepSeek requests
- Non-text blocks (file, image, etc.) are replaced with \[omitted <type> block]\
- Requests to other APIs (non-DeepSeek) are **not** affected
- Empty content falls back to empty string gracefully

### 3. CI checks not running

Workflow runs (CodeQL, Format, Lint) are awaiting maintainer approval and have not executed yet.

## Acceptance Criteria

- [ ] Fetch rewriter is scoped to the model instance, not globalThis
- [ ] Unit tests cover content flattening, non-text block handling, and pass-through for non-DeepSeek URLs
- [ ] All CI checks pass

the-shiva-123:
Hi @himanshu231204, I'd love to work on this refactor as my very first code contribution to the project! I plan to isolate the global fetch rewriter into a scoped configuration for the model instance and add the requested unit tests. Could you please assign this to me?

himanshu231204:
@the-shiva-123 This issue is assigned to you.

the-shiva-123:
Hi @himanshu231204, 

I wanted to let you know that I have completed the fix for this issue! 

I extracted the global fetch rewriter and successfully scoped it directly to the model instance configuration inside `src/agent/index.ts`, eliminating the global side effects. I also built out a comprehensive unit test suite inside `test/deepseek.test.ts` to verify request payload flattening and omission logic. Everything is building and passing cleanly.

[...]

PR #129, opened 2026-07-06
refactor: scope DeepSeek fetch rewriter and add tests #126

the-shiva-123:
### 📝 Description
This PR resolves issue #126 by refactoring the DeepSeek fetch rewriter. Instead of intercepting network traffic globally via an absolute override on `globalThis.fetch`, the request transformation logic is now localized and injected strictly into the `ChatOpenAI` client configuration block during provider initialization.

### 🛠️ Changes Implemented
* **Scoped Implementation:** Isolated the DeepSeek payload rewriter wrapper within a localized custom fetch handler inside `src/agent/index.ts`.
* **Client Injection:** Configured instance-level injection on the LangChain model constructor context via the configuration parameters, preventing process-wide side effects.
* **Test Coverage:** Created `test/deepseek.test.ts` to fully validate string block flattening, non-text block omissions, and URL isolation.

### 📊 Testing Data
Successfully verified through the project test suite with all checks passing cleanly locally:
* `test/deepseek.test.ts` (3 tests passed)
* **Total repository health:** 9 passed, 0 failed (all existing tests remain unbroken)

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -443,13 +443,51 @@ async function createModel(provider: OpenWikiProvider, modelId: string) {
 
   const baseURL = resolveProviderBaseUrl(provider);
 
+  // Define a localized scoped fetch client rewriter for DeepSeek
+  const customFetch = async (input: Parameters<typeof fetch>[0], init?: Parameters<typeof fetch>[1]) => {
+    let url = "";
+    if (typeof input === "string") {
+      url = input;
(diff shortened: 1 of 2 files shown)


colifran:
Thanks for this. It's stacked on #68 (which was abandoned) and refactors code that isn't on main anymore, so it doesn't apply cleanly and I'm going to close it. If DeepSeek support is the goal, #139 is the live path for that.

PR #131, opened 2026-07-06
fix: prevent ELOOP crash from symlink cycles in directory traversal

### LINKED ISSUE #72: Crawler follows symlink cycles until ELOOP, aborting the run (after docs were already generated)
## Summary

If the target repo contains a directory symlink cycle — in our case a stray self-referential symlink `piing-mobile -> .` inside a subdirectory — the filesystem traversal follows it recursively until Node throws, and the whole run exits 1:

```
ELOOP: too many symbolic links encountered, scandir '/…/piing-mobile/piing-mobile/piing-mobile/…/piing-mobile/ios/Pods/…'
```

Frustratingly, in our run this happened at the very END: all wiki pages had already been written, and the crash killed the remaining steps (the `CLAUDE.md`/`AGENTS.md` reference-section wiring and `openwiki/.last-update.json`), leaving an initialized-looking wiki that `--update` can't diff from.

## Repro

```sh
mkdir -p demo/sub && cd demo && git init -q
echo hi > sub/file.txt
ln -s . sub/sub          # self-referential dir symlink (ours was created accidentally by tooling)
openwiki --init -p "Generate documentation for this repository"
# → ELOOP crash when traversal reaches sub/sub/sub/...
```

## Suggested fix

Don't follow directory symlinks during traversal (the read tool already refuses symlinked *files* on non-O_NOFOLLOW platforms, so skipping symlinked dirs is consistent), or track visited real paths (`fs.realpath`) and skip already-seen ones. Also worth making the post-generation steps resilient: a traversal error late in the run shouldn't discard the `.last-update.json` write for pages already produced.

## Environment

openwiki 0.0.1, npm global install, Linux, node v22.22.3, provider Anthropic, model `claude-sonnet-5`.

🤖 Investigated with [Claude Code](https://claude.com/claude-code)

akyourowngames:
## Summary

Fixes #72 — Symlink cycles cause ELOOP crash during directory traversal.

When a repository contains symbolic link cycles (e.g., `a/link -> a/`), the recursive `addDirectoryToSnapshot()` function follows the symlinks endlessly until Node.js throws an `ELOOP` error, crashing the entire run.

## Changes

**`src/agent/utils.ts`**
- Added `realpath` import from `node:fs/promises`
- Modified `addDirectoryToSnapshot()` to track visited real paths via a `Set<string>` parameter
- Before recursing into a directory, resolves its real path with `realpath()` and checks if it has been visited — if so, skips it (cycle detected)
- Updated `createOpenWikiContentSnapshot()` to initialize the visited set

**`test/symlink-cycle.test.ts`** (new file)
- Regression tests for symlink cycles, multi-hop cycles, dangling symlinks, and valid symlinks
- Tests gracefully skip on Windows where symlinks require elevated privileges

## How it works

The primary mechanism is **cycle detection via realpath-based visited set**:

1. Before entering any directory, `realpath()` resolves symlinks and hardlinks to their real location
2. If that real path has been seen before, the directory is skipped — this prevents both symlink cycles and duplicate traversal of hardlinked directories
3. The visited set is threaded through recursive calls so every directory is visited at most once

This approach is sufficient on its own — no depth limit or blanket symlink skipping is needed. The `realpath()` call naturally handles:
- Direct cycles (`a/link -> a/`)
- Multi-hop cycles (`a/link -> b/link -> a/`)
- Hardlinked directories (same inode, different paths)
- Dangling symlinks (returns error, caught by `.catch(() => null)`)

## Testing

- Build passes
- TypeScript compiles cleanly
- All 11 tests pass (6 existing + 5 new regression tests)


rajpratham1 (changes requested):
The PR needs:

Remove or fix the ineffective entry.isSymbolicLink() check.
Justify or remove the MAX_SNAPSHOT_DEPTH limit.
Add a regression test covering an actual symlink cycle (ELOOP).
Clarify whether the goal is preventing ELOOP or preventing duplicate traversal, since the current implementation primarily does the latter.

akyourowngames:
@rajpratham1 ready for review

akyourowngames:
## CI formatting fix

**Why the check failed:** the `Format` job ran `prettier --check .` and reported only `test/symlink-cycle.test.ts`. Its multi-line `node:fs/promises` import did not match the repository's Prettier output.

**What changed:** reformatted that test file with the repository's Prettier version. No test behavior or production code changed.

**Verification:**
- `prettier --check test/symlink-cycle.test.ts` passes for the committed LF blob (the same form GitHub Actions checks).
- `vitest run test/symlink-cycle.test.ts --pool=threads` passes: 5/5 tests.

The replacement Checks workflow is currently awaiting maintainer approval: https://github.com/langchain-ai/openwiki/actions/runs/29270251115

akyourowngames:
@colifran Sorry for disturbing you again, solved some test issues!!
Ready!!

colifran:
Hi @akyourowngames - I don't think this is the actual fix. We already don't follow symlinks since we're using `Dirent.isDirectory` which returns false for symlinks. This seems like something we need to fix upstream in deepagents.

akyourowngames:
Thanks for the correction — I confirmed the OpenWiki traversal itself is not following symlinks via `Dirent.isDirectory`, so this PR’s visited-path workaround does not fix the root cause.

The actual issue is `fast-glob` in DeepAgents following directory symlinks. I see the upstream fix is now tracked in langchain-ai/deepagentsjs#668, which disables symlink following in both backends and adds regression coverage. I’ll leave this PR unchanged rather than duplicate or maintain the wrong-layer workaround; once the upstream release is available, OpenWiki can consume it through its DeepAgents dependency.

colifran:
Closing due to upstream fix: https://github.com/langchain-ai/deepagentsjs/pull/668

PR #133, opened 2026-07-06
fix: openrouter claude opus model id is incorrect

akyourowngames:
## Summary

Fixes part of #86 — OpenRouter Claude Opus model ID is wrong, causing model not found errors.

The OpenRouter provider config uses `anthropic/claude-opus-4.8` for the Opus model, but the correct model ID is `anthropic/claude-opus-4-8` (hyphens, not dots). This causes a "model not found" error when users select Claude Opus through OpenRouter.

## Root cause

Anthropic's model version format uses hyphens as separators (`claude-opus-4-8`), not dots. The Anthropic provider config in the same file already has the correct ID (`claude-opus-4-8`), but the OpenRouter config was missed.

## Change

**`src/constants.ts`**
- Fixed OpenRouter Opus model ID: `anthropic/claude-opus-4.8` → `anthropic/claude-opus-4-8`

## Verification

- The Anthropic provider config at line 101 already uses `claude-opus-4-8` (correct)
- The OpenRouter provider config now matches: `anthropic/claude-opus-4-8` (correct)
- TypeScript compiles cleanly
- Existing tests pass


rajpratham1 (approved):
This is a straightforward bug fix with very low risk.

colifran (approved):
Looks good. Thank you.

PR #140, opened 2026-07-06
fix: stop dropping OPENAI_BASE_URL so OpenAI proxies work

### LINKED ISSUE #64: OpenAI Base URL env var is dropped in src/env.ts

varaprasadreddy9676:
## Summary

Fixes #64.

`OPENAI_BASE_URL` is currently listed in `deprecatedEnvKeys` in `src/env.ts`, so any value a user saves to `~/.openwiki/.env` (or exports inline) is **silently discarded on load** and never reaches the OpenAI provider. This forces users behind an auditing proxy / OpenAI-compatible gateway to prepend the env var before every invocation, as described in the issue.

This is a leftover from when the default provider moved off OpenAI (commit `1473a12`). `OPENAI_API_KEY` has since been un-deprecated as OpenAI became a first-class provider again, but the base URL override was left behind in the deprecated list. This PR un-deprecates it and routes it through the same override channel that `ANTHROPIC_BASE_URL` and `OPENAI_COMPATIBLE_BASE_URL` already use.

## Root cause

`loadOpenWikiEnv` skips any key in `deprecatedEnvKeys`, and `OPENAI_BASE_URL` was still in that list despite the OpenAI provider being fully supported.

## Changes

- **`src/constants.ts`** — add `OPENAI_BASE_URL_ENV_KEY = "OPENAI_BASE_URL"` and set it as the `openai` provider's `baseUrlEnvKey`, so `resolveProviderBaseUrl("openai")` honors it.
- **`src/env.ts`** — remove `OPENAI_BASE_URL` from `deprecatedEnvKeys`; add it to `managedEnvKeys` and `getCredentialDiagnostics`; treat it as a non-secret (URL preview) diagnostic like the other base URLs.
- **`src/agent/index.ts`** — include `OPENAI_BASE_URL` in `--debug` environment logging.
- **`README.md`** — document the OpenAI base URL override alongside the existing Anthropic one.

The interactive setup in `src/credentials.tsx` needs no change — its base-URL step is fully data-driven off `baseUrlEnvKey`, so it now renders for OpenAI when the override is absent and skips it when present.

## Usage

```bash
OPENWIKI_PROVIDER=openai
OPENAI_API_KEY=your-key
OPENAI_BASE_URL=https://your-gateway.example.com/v1
```

## Tests

- **`test/constants.test.ts`** (new) — openai base-url resolution, whitespace/blank handling, isolation from the anthropic provider, and the base-URL override surface across all providers.
- **`test/env.test.ts`** (new) — the #64 regression: `OPENAI_BASE_URL` is **persisted and loaded** rather than dropped; `process.env` precedence over the file is preserved; genuinely deprecated keys (`OPENAI_ORG_ID`, `OPENAI_PROJECT`) are still ignored; diagnostics surface the override as a non-secret URL; save/load round-trip and 0600 file mode.

The regression test runs against a throwaway `HOME` so it never touches real credentials.

## Verification

```
pnpm run build        # OK
pnpm test             # 25 passed (6 existing + 19 new), existing tests unchanged
pnpm run lint:check   # OK
pnpm run format:check # OK
```

Closes #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,15 @@ ANTHROPIC_API_KEY=your-key
 ANTHROPIC_BASE_URL=https://your-gateway.example.com/anthropic
 ```
 
+The OpenAI provider can be routed the same way, for example to an auditing
+proxy or an OpenAI-compatible gateway, by setting `OPENAI_BASE_URL`:
+
+```bash
+OPENWIKI_PROVIDER=openai
+OPENAI_API_KEY=your-key
+OPENAI_BASE_URL=https://your-gateway.example.com/v1
+```
+
 ### OpenAI-compatible endpoints
 
 The `openai-compatible` provider targets any OpenAI-compatible chat-completions
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -26,6 +26,7 @@ import {
   isValidModelId,
   normalizeModelId,
   OPENAI_API_KEY_ENV_KEY,
+  OPENAI_BASE_URL_ENV_KEY,
   OPENAI_COMPATIBLE_API_KEY_ENV_KEY,
   OPENAI_COMPATIBLE_BASE_URL_ENV_KEY,
   OPENROUTER_API_KEY_ENV_KEY,
@@ -1306,6 +1307,7 @@ function formatEnvironmentDebug(): string {
     BASETEN_API_KEY_ENV_KEY,
     FIREWORKS_API_KEY_ENV_KEY,
     OPENAI_API_KEY_ENV_KEY,
+    OPENAI_BASE_URL_ENV_KEY,
     OPENAI_COMPATIBLE_API_KEY_ENV_KEY,
     OPENAI_COMPATIBLE_BASE_URL_ENV_KEY,
     ANTHROPIC_API_KEY_ENV_KEY,
@@ -1330,6 +1332,7 @@ function formatDebugValue(key: string, value: string | undefined): string {
   if (
     key === "LANGCHAIN_ENDPOINT" ||
     key === ANTHROPIC_BASE_URL_ENV_KEY ||
+    key === OPENAI_BASE_URL_ENV_KEY ||
     key === OPENAI_COMPATIBLE_BASE_URL_ENV_KEY
   ) {
     return formatUrlDebugValue(value);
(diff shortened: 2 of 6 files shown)


ppsplus-bradh:
@varaprasadreddy9676 This probably achieves what you are looking for.  https://github.com/langchain-ai/openwiki/pull/110

colifran:
@ppsplus-bradh are you saying this functionality is implemented via #110?

varaprasadreddy9676:
@ppsplus-bradh @colifran — good callout, and you're right that `openai-compatible` (+ `OPENAI_COMPATIBLE_BASE_URL`) is now the recommended way to front a gateway.

This PR is narrower: it's specifically about `OPENAI_BASE_URL` being listed in `deprecatedEnvKeys` and silently dropped, even though `OPENAI_API_KEY` was already un-deprecated when OpenAI became a first-class provider. `OPENAI_BASE_URL` is the env var the OpenAI SDK honors by default, so folks who already export it (e.g. for an audit proxy) hit a silent no-op today.

Totally happy to close this if maintainers prefer `openai-compatible` as the single proxy path. But if the `openai` provider is meant to behave like a standard OpenAI client, honoring `OPENAI_BASE_URL` feels consistent. Either way, I'll follow your lead — and the regression test here is still useful coverage for `src/env.ts` regardless.

ppsplus-bradh:
@colifran Yes.  #110 adds support for OpenAI compatible providers.  I use a litellm gateway.

ppsplus-bradh:
> Brad Huffman (@ppsplus-bradh) Colin Francis (@colifran) — good callout, and you're right that `openai-compatible` (+ `OPENAI_COMPATIBLE_BASE_URL`) is now the recommended way to front a gateway.
> 
> This PR is narrower: it's specifically about `OPENAI_BASE_URL` being listed in `deprecatedEnvKeys` and silently dropped, even though `OPENAI_API_KEY` was already un-deprecated when OpenAI became a first-class provider. `OPENAI_BASE_URL` is the env var the OpenAI SDK honors by default, so folks who already export it (e.g. for an audit proxy) hit a silent no-op today.
> 
> Totally happy to close this if maintainers prefer `openai-compatible` as the single proxy path. But if the `openai` provider is meant to behave like a standard OpenAI client, honoring `OPENAI_BASE_URL` feels consistent. Either way, I'll follow your lead — and the regression test here is still useful coverage for `src/env.ts` regardless.

Good call on supporting the generally established convention.  Should have phrased my reply differently.  In reviewing the codebase, Claude calls out the deprecation without a "why".  LGTM.

colifran:
Thanks let's go ahead and close in favor of https://github.com/langchain-ai/openwiki/pull/110.

PR #142, opened 2026-07-06
feat: add model ID resolution and support functions

krataratha:
Added functions to resolve model IDs based on environment and provider support.


colifran (commented):
Hi @krataratha. Thanks for opening this. Can you help me understand the motivation for this? Also, this needs unit tests to help validate the functionality.

PR #150, opened 2026-07-06
fix: report tool-error runs as failed

### LINKED ISSUE #43: --init/--update exit 0 after unresolved tool errors (partial output)
## Summary

When `openwiki --init` (or `--update`) hits a fatal tool-invocation error mid-run — e.g. the `write_file` schema-validation crash described in #39 — the CLI process still exits 0. A caller scripting around the exit code (CI, a nightly cron, any automation) has no signal that the run was partial.

## Repro

```sh
openwiki --init -p "Generate documentation for this repository"
# ... crashes partway through with:
#   Error invoking tool 'write_file' with kwargs {} with error: Error: Received tool input did
#   not match expected schema
#   ✖ Invalid input: expected string, received undefined
#     → at file_path
echo $?
# => 0
```

Observed: `openwiki/_plan.md` showed 6 planned pages; only 2 were written before the crash on page 3. The `openwiki` directory is left with partial, inconsistent output (e.g. quickstart.md linking to pages that were never created) and no error indicator anywhere except stdout text that a script would have to pattern-match to detect.

## Root cause (from source, openwiki 0.0.1)

- `dist/agent/index.js:116-125` (`runOpenWikiAgentCore`) iterates `agent.stream(...)` with a bare `for await`. The LangGraph stream resolves *normally* once the graph run ends — including when it ends because a tool call kept failing and the model/graph gave up — because the tool error is delivered as a `ToolMessage` back to the model, not as a thrown exception.
- The stream parser (`parseToolStreamEvent`, `index.js:573-612`) already classifies `on_tool_error`/`tool-error` events into `{type: "tool_end", status: "error"}` — the signal is available — but the loop at `index.js:116-125` only forwards events for display (`options.onEvent?.(event)`) and never checks `status === "error"` to flag the run as failed.
- `runOpenWikiAgentCore` then returns a plain success object (`index.js:138-141`) regardless of whether any tool errored during the run.
- `runPrintCommand` (`dist/cli.js:1593-1619`) sets `process.exitCode = 0` (line 1612) whenever `runOpenWikiAgent(...)` doesn't throw — which, per the above, it never does for this failure class.

## Suggested fix

In the event loop (`index.js:116-125`), track whether any `tool_end` event with `status === "error"` was observed, and:
1. Surface it in the object `runOpenWikiAgentCore`/`runOpenWikiAgent` returns (e.g. `{command, model, hadToolError: true}`), and
2. In `runPrintCommand` (and the equivalent interactive/auto-exit path, `shouldAutoExitStartupRun`/exit-code logic around `cli.js:99,1569`), set a distinct non-zero exit code (or at minimum reuse `process.exitCode = 1`) when `hadToolError` is true, even though no JS exception was thrown.

This is likely the same underlying trigger as #39 (OpenRouter upstream stream truncation) — but independent of *why* the tool call failed, a run that ends with an unresolved tool error should never report exit 0.

## Environment

openwiki 0.0.1, npm global install, macOS (darwin 25.3.0), provider OpenRouter, model `z-ai/glm-5.2`.

dpersek:
## Summary
- Track whether an agent run observed a tool-error stream event.
- Avoid recording successful update metadata for `--init` / `--update` runs that saw a tool error.
- Return a non-zero exit code for `--print` and auto-exit `--init` / `--update` runs when the resolved run result contains tool errors.

## Why
Fixes #43

## Tests
- `pnpm exec vitest run test/agent-tool-error-result.test.ts test/run-result.test.ts`
- `pnpm test`
- `pnpm run typecheck`
- `pnpm run lint:check`
- `pnpm run format:check`
- `pnpm run build`
- `git diff --check`

## Scope
- This reports completed runs with observed tool errors as failed for automation.
- This does not change provider retry behavior, prompt wording, malformed tool-call generation, or LangChain/DeepAgents recovery semantics.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -233,12 +233,17 @@ async function runOpenWikiAgentCore(
   });
   emitDebug(options, "stream=started modes=messages,tools subgraphs=true");
 
+  let hadToolError = false;
   let unhandledChunkCount = 0;
 
   for await (const chunk of stream) {
     const event = parseStreamEvent(chunk);
 
     if (event) {
+      if (event.type === "tool_end" && event.status === "error") {
+        hadToolError = true;
+      }
+
       options.onEvent?.(event);
     } else if (options.debug && unhandledChunkCount < 3) {
       emitDebug(
@@ -253,6 +258,7 @@ async function runOpenWikiAgentCore(
 
   if (
     command !== "chat" &&
+    !hadToolError &&
     openWikiSnapshotBefore !== (await createOpenWikiContentSnapshot(cwd))
   ) {
     await writeLastUpdateMetadata(command, cwd, modelId);
@@ -262,12 +268,15 @@ async function runOpenWikiAgentCore(
       options,
       command === "chat"
         ? "metadata=skipped command=chat"
-        : "metadata=skipped openwiki=unchanged",
+        : hadToolError
+          ? "metadata=skipped tool_error=true"
+          : "metadata=skipped openwiki=unchanged",
     );
   }
 
   return {
     command,
+    ...(hadToolError ? { hadToolError } : {}),
     model: modelId,
   };
 }
--- a/src/agent/result.ts
+++ b/src/agent/result.ts
@@ -1,0 +1,5 @@
+import type { OpenWikiRunResult } from "./types.js";
+
+export function getRunExitCode(result: OpenWikiRunResult): 0 | 1 {
+  return result.hadToolError ? 1 : 0;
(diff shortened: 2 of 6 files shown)


dpersek:
Closing this as superseded by #224, which the maintainer identified as resolving #43. The dependency update changes malformed tool-call handling to recover and retry, so this PR’s “any observed tool error means failure” approach is no longer the right fit. Happy to revisit if a distinct exit-code failure remains reproducible on current `main`.

PR #151, opened 2026-07-06
feat: add OpenAI ChatGPT-login provider (Codex OAuth)

Yassine-Mhirsi:
## What this adds

A new **"OpenAI (ChatGPT login)"** provider that signs in with your ChatGPT account (OAuth) instead of a metered API key — so model usage draws on your ChatGPT Plus/Pro/Team Codex allowance rather than per-token API billing. Same models as the `openai` provider (`gpt-5.4-mini`, `gpt-5.5`).

```bash
OPENWIKI_PROVIDER=openai-chatgpt openwiki --init   # sign in, then use OpenWiki as usual
```

## Demo

https://github.com/user-attachments/assets/ce622f65-a8a0-4bec-8687-64c9813b8555

## How it works

The main goal was to **not touch the agent harness**. Instead of writing a custom chat model, it reuses `@langchain/openai`'s `ChatOpenAI` (Responses API) pointed at the Codex backend, so DeepAgents' tool-calling and streaming keep working unchanged. `createDeepAgent()`, `LocalShellBackend`, the checkpointer, and the streaming/parsing code are all untouched — everything new lives in model construction and credential setup.

Three Codex-backend quirks are handled where the model is built (all commented in code): it sends the account/originator/beta headers with `store: false`, forces `stream: true` (the backend rejects non-streaming requests), and rewrites the `system` message role to `developer` (the backend rejects `system`).

## What's included

- New `openai-chatgpt` provider with an `authMethod: "oauth"` field on the provider config.
- Browser **PKCE login** with automatic token refresh; tokens are stored in `~/.openwiki/.env` like any other credential.
- A new **login step** in the setup wizard that then continues to model + LangSmith selection, just like the API-key providers.
- **Remote-friendly login:** opens the browser, prints the URL (press `c` to copy it), and lets you **paste the redirect URL back into the terminal** to finish — no SSH port-forwarding or `curl` needed.
- Shows the signed-in **email and plan** in the wizard and the run header.
- Tests (network mocked) and README docs.

## How to test

No account or network needed:

```bash
pnpm test && pnpm lint:check && pnpm format:check && pnpm build
```

With a ChatGPT subscription:

```bash
pnpm build
OPENWIKI_PROVIDER=openai-chatgpt node dist/cli.js --init
```

Sign in (or paste the redirect URL back on a remote host), pick a model, and the docs run against the Codex backend with tool calls working.


Yassine-Mhirsi:
Hi @bracesproul 👋

This PR adds an **"OpenAI (ChatGPT login)"** provider so OpenWiki can run on a ChatGPT subscription (Codex OAuth) instead of a metered API key — with **zero changes to the DeepAgents harness** (it reuses `@langchain/openai`'s `ChatOpenAI` Responses API pointed at the Codex backend, so tool-calling and streaming keep working).

It includes a PKCE browser login with auto token refresh, a remote-friendly flow (copy the URL or paste the redirect back — no SSH forwarding needed), the signed-in email/plan shown in the UI, tests (network mocked), and docs. There's a short demo in the description.

Would you have a moment to take a look, or point me to whoever owns provider integrations? Happy to adjust anything. Thanks!

Yassine-Mhirsi:
A couple of properties worth flagging for anyone comparing subscription-auth approaches:

- **No external CLI to install.** This talks to the Codex backend directly over HTTPS via LangChain's `ChatOpenAI` — there's no vendor binary (`codex`, `claude`, etc.) to install or keep on `PATH`. The only requirement is the ChatGPT login, so there's no extra install step and no platform-specific binary handling.
- **Agent loop unchanged.** `createDeepAgent()`, `LocalShellBackend`, the checkpointer, and the streaming/tool-parsing code are all untouched — the provider is just model construction + credential setup. It runs anywhere Node runs and doesn't fork the agent runtime.

Happy to adjust if a different trade-off is preferred.

src/agent/index.ts
```diff
@@ -442,6 +497,93 @@ function createModel(provider: OpenWikiProvider, modelId: string) {
   });
 }
 
+/**
+ * Reads the persisted ChatGPT OAuth tokens and refreshes them once at startup
+ * when expired/near-expiry, persisting the rotated tokens back to
+ * `~/.openwiki/.env`. This is a short-lived CLI process, so a single
+ * refresh-at-startup is sufficient — there is no background refresh loop.
+ */
+async function resolveChatGptTokens(): Promise<CodexTokens> {
```
colifran: i'm seeing token to env mapping being duplicated (here for read/write, and credentials.tsx:674 + :99).

Since openai-chatgpt-oauth.ts owns the CodexTokens type, we should create a `codexTokensToEnv` / `readCodexTokensFromEnv` to centralize the logic and also lets the env contract be tested alongside the rest of the OAuth flow.

Something like:
```ts
export function codexTokensToEnv(t: CodexTokens): Record<string, string> {
    return {
      [OPENAI_CHATGPT_ACCESS_TOKEN_ENV_KEY]: t.access,
      [OPENAI_CHATGPT_REFRESH_TOKEN_ENV_KEY]: t.refresh,
      [OPENAI_CHATGPT_EXPIRES_AT_ENV_KEY]: String(t.expiresAtMs),
      [OPENAI_CHATGPT_ACCOUNT_ID_ENV_KEY]: t.accountId,
      ...(t.email ? { [OPENAI_CHATGPT_EMAIL_ENV_KEY]: t.email } : {}),
      ...(t.planType ? { [OPENAI_CHATGPT_PLAN_ENV_KEY]: t.planType } : {}),
    };
  }

  export function readCodexTokensFromEnv(env = process.env): CodexTokens | null {
    const access = env[OPENAI_CHATGPT_ACCESS_TOKEN_ENV_KEY];
    const refresh = env[OPENAI_CHATGPT_REFRESH_TOKEN_ENV_KEY];
    const accountId = env[OPENAI_CHATGPT_ACCOUNT_ID_ENV_KEY];
    if (!access || !refresh || !accountId) return null;
    return {
      access, refresh, accountId,
      expiresAtMs: Number(env[OPENAI_CHATGPT_EXPIRES_AT_ENV_KEY]),
      email: env[OPENAI_CHATGPT_EMAIL_ENV_KEY] ?? null,
      planType: env[OPENAI_CHATGPT_PLAN_ENV_KEY] ?? null,
    };
  }
```

src/cli.tsx
```diff
@@ -696,6 +701,19 @@ function ErrorDiagnosticsPanel({
   );
 }
 
+/** Builds an `email (Plan)` label from the persisted ChatGPT identity, if any. */
+function formatChatGptAccountLabel(): string | null {
```
colifran: This is basically a copy of `formatChatGptAccount` in `credentials.tsx:145`. Can we centralize this logic somewhere?

src/credentials.tsx
```diff
@@ -67,6 +113,92 @@ function isBaseUrlConfigured(provider: OpenWikiProvider): boolean {
...
+function openLoginUrl(url: string): void {
+  const command =
+    process.platform === "darwin"
+      ? "open"
+      : process.platform === "win32"
+        ? "start"
+        : "xdg-open";
+
+  try {
+    const child = spawn(command, [url], {
```
colifran: I think this may hit issues on the windows side. I'm pretty sure the `&s` in the auth URL will get split by cmd under `shell: true`. Everything else is fine but I think for windows we need:

```ts
spawn("cmd", ["/c", "start", '""', "${url}"], { windowsVerbatimArguments: true })
```

Might be worth double checking?

colifran (commented):
Really great work. Thank you for the PR! Left some comments.

src/agent/index.ts
Yassine-Mhirsi: Good call, done. Both helpers now live in `openai-chatgpt-oauth.ts` right next to `CodexTokens`, basically as you sketched: `codexTokensToEnv()` and `readCodexTokensFromEnv()`. `resolveChatGptTokens()` here and `completeSetup()` in credentials.tsx both go through them, so there is one place that knows the env shape.

One useful side effect: `readCodexTokensFromEnv()` returns null unless access, refresh and account id are all present, so `hasValidStoredToken()` in the wizard picks up partial state too. Before, a leftover access token with no account id looked signed in and then failed at request time. Now it just sends you back through the login.

Added round trip tests for the env contract, plus one for the partial case.

src/cli.tsx
Yassine-Mhirsi: Agreed, that was a straight copy. Moved it into `openai-chatgpt-oauth.ts` next to the identity decoding, and exported two things: `formatChatGptAccount(email, plan)` for callers that already have the values, and `formatChatGptAccountFromEnv()` for the header, which just reads the persisted ones. cli.tsx and credentials.tsx both use those now, and it is covered by tests.

src/credentials.tsx
Yassine-Mhirsi: You are right, thanks for catching that. Under `shell: true`, cmd treats the `&` between the auth URL's query params as a command separator, so the URL gets chopped after the first param and the rest runs as separate commands.

Dropped `shell` entirely and went with your version:

```ts
spawn("cmd", ["/c", "start", '""', `"${url}"`], { windowsVerbatimArguments: true })
```

The empty `""` is `start`'s window title argument, otherwise `start` swallows the quoted URL as the title. `windowsVerbatimArguments` stops Node from re-quoting what we already quoted.

Yassine-Mhirsi:
Thanks @colifran, really useful review. All three are addressed in cfedf53:

1. **Token to env mapping.** `codexTokensToEnv()` and `readCodexTokensFromEnv()` now live in `openai-chatgpt-oauth.ts` next to `CodexTokens`, and both the agent and the wizard go through them. Bonus: reading back a partial token set now returns null, so a leftover access token without an account id no longer looks signed in and then fails at request time.
2. **Duplicate account formatter.** Moved into the same module and shared by `cli.tsx` and `credentials.tsx`.
3. **Windows browser open.** Confirmed bug, cmd was splitting the auth URL on the `&` in its query string. Switched to `cmd /c start` with a quoted URL and `windowsVerbatimArguments`, no shell. I only have Linux here so it is untested on Windows, flagged that in the thread.

Test count went from 35 to 42, mostly covering the new env contract. Lint, format, typecheck and build are all clean.

Yhozen (approved):

Byevenes (approved):

colifran (dismissed):
This looks great! Thanks for making these changes. Approved on my side, but want to let @bracesproul take a look as well.

bracesproul (approved):
love this!!!! thanks a ton. do you have a twitter account i can give credit to when i announce this?

thanks again!

Yassine-Mhirsi:
> love this!!!! thanks a ton. do you have a twitter account i can give credit to when i announce this?
> 
> thanks again!

glad to help!! thanks a lott https://x.com/Topzsixx

gowthamgts:
@bracesproul - will it be tagged and released in 1.0.1? please release this 🤞🏼

bracesproul:
> s

yes! cutting this release now

bracesproul:
@gowthamgts released now!

PR #153, opened 2026-07-07
fix: run --init/--update non-interactively when stdin is not a TTY

### LINKED ISSUE #149: [bug] openwiki --init crashes with "Raw mode is not supported" when stdin is not a TTY (Ink requirement)
## `openwiki --init` crashes with "Raw mode is not supported" when stdin is not a TTY

**OpenWiki version**: v0.0.1+post-merge (commit `e276b08`)
**Provider**: `openai-compatible` → `https://api.minimax.io/v1` (no Anthropic involved)
**Environment**: Linux, Node 24.15.0, pnpm 10.33.2

### Summary

`openwiki --init` and `openwiki --update` abort immediately with `ERROR: Raw mode is not supported on the current process.stdin, which Ink uses as input stream by default.` whenever the process is launched from a non-interactive shell (CI, cron, sub-agent, background process), **even when all required credentials are present in `~/.openwiki/.env`**.

This makes OpenWiki unusable in any non-TTY context, including the cron-driven update pattern the README implicitly recommends (`examples/openwiki-update.yml` runs in GitHub Actions where stdin is not a TTY).

### Reproduction steps

```bash
# Setup ~/.openwiki/.env with valid credentials
cat ~/.openwiki/.env
# OPENWIKI_PROVIDER=openai-compatible
# OPENAI_COMPATIBLE_BASE_URL=https://api.minimax.io/v1
# OPENAI_COMPATIBLE_API_KEY=<valid key>
# OPENWIKI_MODEL_ID=MiniMax-M3
# LANGSMITH_API_KEY=

# Run from a non-TTY shell (this is exactly what cron / CI / a sub-agent does)
cd /path/to/codebase
script -qec "openwiki --init" /tmp/out.log </dev/null
# or:
echo "" | openwiki --init
# or:
openwiki --init < /dev/null
```

### Expected behaviour

When stdin is not a TTY (e.g. cron, CI, automation), `openwiki --init` should either:

1. **Bypass Ink's interactive setup wizard entirely** (all credentials present in `~/.openwiki/.env` → just run the agent).
2. **Auto-detect non-TTY and use a `--no-tty` / `--non-interactive` mode that uses `--print` style output**.
3. **Document the requirement** (and provide a canonical `script(1)` wrapper example).

### Actual behaviour

```
╭──────────────────────────────────────────────────────────────────────────────╮
│ >_ OpenWiki v0.0.1 agent docs for codebases                                  │
│ provider: OpenAI-compatible                                                  │
│ model: MiniMax-M3                                                            │
│ directory: ~/Ousia/aetheer                                                   │
╰──────────────────────────────────────────────────────────────────────────────╯╯

- LangSmith tracing disabled - Ready for chat
Tip: ask for a docs change, or /exit when you are done.

  ERROR Raw mode is not supported on the current process.stdin, which Ink uses
       as input stream by default.
       Read about how to prevent this error on
       https://github.com/vadimdemedes/ink/#israwmodesupported
```

The credential setup wizard correctly identified that provider/key/base_url/model are all "DONE available from environment", but then crashed before the agent could start because Ink (the TUI framework) requires raw mode on stdin.

### Root cause

`src/cli.tsx:248` (`shouldRunInteractiveCredentialSetup = … && process.stdin.isTTY`) — when not a TTY and credentials are already configured, the wizard still attempts to enter Ink's interactive setup, and Ink requires `process.stdin.setRawMode(true)` which is unavailable on non-TTY stdin.

### Workaround (applied)

Wrap the call with `script(1)` to provide a pseudo-terminal:

```bash
script -qec "openwiki --init" /tmp/out.log < /dev/null
```

This lets OpenWiki see a TTY. Combined with `LANGSMITH_API_KEY=""` to skip the wizard entry gate, the init succeeds (full 8-page wiki generated against MiniMax-M3 in ~10 min for a 28.5k-LOC codebase).

### Suggested fix

Two paths upstream could resolve this cleanly:

1. **Add a `--non-interactive` / `--headless` flag** that:
   - Skips Ink entirely.
   - Uses the same flags as `--print` (`-p`) plus `--init` / `--update` semantics.
   - Reads `--init-prompt <text>` or `--update-prompt <text>` for the agent's brief.

2. **Auto-detect**: when `process.stdin.isTTY === false` AND all required env vars are present, automatically enter non-interactive mode (skip the wizard and Ink TUI, just run the agent in batch mode).

Either fix would let OpenWiki work in CI / cron / sub-agent contexts directly, without users needing to know about `script(1)`.

### Impact

Currently OpenWiki cannot be used by:
- Cron jobs that maintain a fresh wiki
- CI pipelines (e.g. GitHub Actions, GitLab CI)
- Sub-agents that maintain docs as part of a larger task
- Any wrapping tool that doesn't allocate a TTY

The `examples/openwiki-update.yml` workflow will crash unless the user also wraps it in a TTY-emulating step.

irresi:
## Problem

Running `openwiki --init` or `--update` with stdin that is not a TTY (CI, cron, piped shells) renders the Ink UI, which requires raw-mode stdin. The run prints `ERROR Raw mode is not supported on the current process.stdin` and then **hangs without exiting** — even with valid credentials in `~/.openwiki/.env`. The GitHub Actions / GitLab examples avoid this with `--print`, but the plain commands are broken in any non-interactive context.

Fixes #149.

## Fix

Treat non-TTY runs the same as `--print`:

- Non-TTY `--init` / `--update` / one-shot messages now go through the existing
  `--print` path: run the agent, write output to stdout, exit 0 or 1.
- TTY behavior is unchanged; plain `openwiki` (interactive chat) still needs a terminal.
- The check lives in `src/commands.ts` so it can be unit-tested.

## Verification

- `pnpm run format:check`, `lint:check`, `test` — 92 passing, including 6 new tests covering the TTY × `--print` × `--dry-run` matrix (TTY `--init` keeps the UI; bare `openwiki`, help, error, and dry-run are unchanged).
- End-to-end in a scratch repo: before, `echo "" | openwiki --init` prints the raw-mode error and never exits; after, the same command runs the agent, writes errors to stderr, and exits 1 on failure.

### DIFF
--- a/src/cli.tsx
+++ b/src/cli.tsx
@@ -3055,6 +3055,20 @@ function resolveStartupCommand(command: CliCommand): CliCommand {
   if (
     command.kind === "run" &&
     !command.dryRun &&
+    !command.shouldStart &&
+    !process.stdin.isTTY
+  ) {
+    return {
+      kind: "error",
+      exitCode: 1,
+      message:
+        "Interactive chat requires a terminal. Pass a message or use --init or --update for non-interactive runs.",
+    };
+  }
+
+  if (
+    command.kind === "run" &&
+    !command.dryRun &&
     command.shouldStart &&
     (command.print || !process.stdin.isTTY)
   ) {
--- a/src/cli.tsx
+++ b/src/cli.tsx
@@ -6,6 +6,7 @@ import {
   helpContent,
   isDevelopmentMode,
   parseCommand,
+  shouldRunNonInteractively,
   type CliCommand,
   type HelpRow,
 } from "./commands.js";
@@ -2970,7 +2971,7 @@ const command = resolveStartupCommand(parsedCommand);
 if (shouldPrintStartupError(argv, parsedCommand, command)) {
   process.stderr.write(`${command.message}\n`);
   process.exitCode = command.exitCode;
-} else if (command.kind === "run" && command.print && !command.dryRun) {
+} else if (shouldRunNonInteractively(command, process.stdin.isTTY === true)) {
   await runPrintCommand(command);
 } else {
   render(<App command={command} />);
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -163,6 +163,24 @@ export function parseCommand(argv: string[]): CliCommand {
   };
 }
 
+/**
+ * True when a run must bypass the Ink UI and use the non-interactive path:
+ * either the user asked for print mode, or stdin is not a TTY (CI, cron,
+ * pipes), where Ink's raw-mode input is unavailable and rendering the UI
+ * fails. Interactive chat without a message still requires a TTY, so it is
+ * excluded.
+ */
+export function shouldRunNonInteractively(
+  command: CliCommand,
+  stdinIsTTY: boolean,
+): command is Extract<CliCommand, { kind: "run" }> {
+  return (
+    command.kind === "run" &&
+    !command.dryRun &&
+    (command.print || (!stdinIsTTY && command.shouldStart))
+  );
+}
+
 export function isDevelopmentMode(): boolean {
   return (
     process.env.NODE_ENV === "development" || process.env.OPENWIKI_DEV === "1"
--- a/test/commands.test.ts
+++ b/test/commands.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, test } from "vitest";
-import { parseCommand } from "../src/commands.ts";
+import { parseCommand, shouldRunNonInteractively } from "../src/commands.ts";
 
 // parseCommand's --dry-run gate consults isDevelopmentMode(), which reads
 // NODE_ENV / OPENWIKI_DEV. Pin both to a non-development state per test and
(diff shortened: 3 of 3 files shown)


AmirF194 (commented):
Reviewed this against #149. I ran both main (39493ef) and this branch (c4d987a) in a clean container (node:22) with valid-looking `openai-compatible` credentials in `~/.openwiki/.env` and stdin not a TTY, matching the issue's setup.

What I can confirm from running it:

- On main, `echo "" | openwiki --init` renders the interactive chat UI, prints the Ink raw-mode error, and never runs the init. Across runs I saw both failure modes: a hang, and (worse for CI) an exit 0 with no docs generated, a silent no-op that reads as success.
- On this branch the same command goes through the print path, actually runs the agent, and exits 1 with the provider error (my endpoint was intentionally unreachable). No raw-mode error. Behavior matches `--init --print` on main exactly, which is the right shape: it reuses the existing path instead of adding a new one.
- The 6 new tests fail on main and the full suite (92) passes on the branch. Prettier and eslint are clean on the changed files, and the branch merges cleanly with current main.

The `shouldStart` gating reads correct to me: `--init`/`--update`/one-shot messages route non-interactively, bare `openwiki` still requires a terminal, and `process.stdin.isTTY === true` treats an undefined stdin as non-TTY, which covers spawned/CI contexts.

One optional thought, not blocking: non-TTY bare `openwiki` still dies with the cryptic Ink raw-mode error. Since this PR already distinguishes that case, a short stderr message ("interactive chat requires a terminal") plus exit 1 could be a cheap follow-up, here or separately.

For transparency: I used an AI assistant to run and verify this review; the findings above are from actual container executions.

irresi:
For context, here's how the three startup functions divide responsibility:

  | Function | Operates on | Question it answers | Mutates the command? |
  |---|---|---|---|
  | `resolveStartupCommand` | every command | What does this command actually resolve to in this environment? (run or error) | Yes |
  | `shouldPrintStartupError` | errors only | Should this error be rendered in the UI or written to stderr as plain text? | No |
  | `shouldRunNonInteractively` | runs only | Should this work run through the UI or in print mode? | No |

>  non-TTY bare openwiki still dies with the cryptic Ink raw-mode error.

Since it's initially run command and should be converted to error, I put it on the `resolveStartupCommand`

100yenadmin:
I tested this path as part of an LCO scheduled-docs pilot. Non-interactive reliability matters for us because the intended workflow is a scheduled/manual GitHub Action that only opens docs PRs when repo docs actually need updating.

One adjacent edge I found in stock 0.0.2: a clean `openwiki --update --print` with a current `.last-update.json` skips correctly when credentials are present, but fails before the clean-update no-op check when credentials are absent. For automation, a no-change update should be able to exit 0 without requiring provider credentials, because no model call is needed.

I am planning a small separate PR for that no-key clean-update no-op. It should complement this PR rather than change the non-TTY routing work here:
- no user message
- no source-file changes since the recorded update
- skip before provider credential validation
- keep credential validation unchanged when source files changed or a user message is supplied

colifran:
@irresi thanks for your pr! taking a look

colifran (approved):
Thanks @irresi!

PR #158, opened 2026-07-07
fix: infer openai-compatible provider from credentials in non-interactive runs

### LINKED ISSUE #157: bug: `openwiki -p` requires OPENROUTER_API_KEY even when `openai-compatible` provider is fully configured
### Description

When configuring OpenWiki to use the `openai-compatible` provider with a custom OpenAI-compatible gateway, non-interactive runs (`openwiki -p ...`) still fail with:

```
OPENROUTER_API_KEY is required for non-interactive runs. Run openwiki in an interactive terminal to save credentials.
```

This happens even after:
- Setting `OPENWIKI_PROVIDER=openai-compatible` in `~/.openwiki/.env`
- Setting valid `OPENAI_COMPATIBLE_API_KEY` and `OPENAI_COMPATIBLE_BASE_URL`
- Explicitly unsetting `OPENROUTER_API_KEY` from the shell
- Deleting `~/.openwiki/openwiki.sqlite` to rule out cached credentials

The non-interactive code path appears to hardcode a check for `OPENROUTER_API_KEY` instead of respecting the configured `OPENWIKI_PROVIDER`.

### Environment

- OpenWiki version: v0.0.1
- OS: Linux (devenv container)
- Shell: bash

### Configuration

`~/.openwiki/.env`:

```env
OPENWIKI_MODEL_ID="deepseek-v4-flash"
LANGCHAIN_TRACING_V2="false"
OPENAI_COMPATIBLE_API_KEY="sk-xxx"
OPENAI_COMPATIBLE_BASE_URL="https://opencode.ai/zen/go/v1"
```

Shell before running:

```bash
unset OPENROUTER_API_KEY
export OPENWIKI_PROVIDER=openai-compatible
export OPENWIKI_DEBUG=1
```

### Steps to Reproduce

1. Configure `~/.openwiki/.env` with `openai-compatible` provider (valid key + base URL, no `OPENROUTER_API_KEY`)
2. Run: `openwiki -p "ping"`
3. Observe error: `OPENROUTER_API_KEY is required for non-interactive runs`

Also confirmed: deleting `~/.openwiki/openwiki.sqlite` does not resolve the issue — this is a hardcoded check in the non-interactive code path, not a cached credential problem.

### Expected Behavior

`openwiki -p` should respect `OPENWIKI_PROVIDER` and use the configured provider's credentials for the non-interactive check, not unconditionally require `OPENROUTER_API_KEY`.

### Actual Behavior

```
OPENROUTER_API_KEY is required for non-interactive runs. Run openwiki in an interactive terminal to save credentials.
```

### Additional Notes

- The `--init` wizard also shows `[DONE] Provider: OpenRouter` regardless of what `OPENWIKI_PROVIDER` is set to, suggesting the init UI shares the same OpenRouter-first assumption.
- The README explicitly documents `openai-compatible` as a first-class supported provider, so the non-interactive path just needs to be generalized to check the appropriate key based on `OPENWIKI_PROVIDER` rather than always requiring `OPENROUTER_API_KEY`.
- Related: #26 (custom OpenAI-compatible base URL support)

rohitvipin-inapp:
## Problem
In non-interactive mode (e.g. `openwiki -p`), the startup flow always checked for `OPENROUTER_API_KEY`, even when the user had only configured `OPENAI_API_KEY` + `OPENAI_COMPATIBLE_BASE_URL` for the openai-compatible provider. This caused `resolveConfiguredProvider` to fall back to `openrouter` and fail startup.

Closes #157
Related to #26
Follow-up to #110 (openai-compatible provider)

## Solution
`resolveConfiguredProvider` now infers the provider from the credentials that are actually present, while still preserving OpenRouter as the historical default:
- If `OPENWIKI_PROVIDER` is set explicitly, use it.
- Otherwise, prefer `openai-compatible` when `OPENAI_COMPATIBLE_BASE_URL` is set and `OPENAI_API_KEY` is present (required by the openai-compatible provider).
- Otherwise, prefer `openrouter` when `OPENROUTER_API_KEY` is present.
- Otherwise, fall back to `openrouter` for backwards compatibility.

## Changes
- `src/constants.ts`: infer provider from present API keys and required base URL.
- `test/constants.test.ts`: add regression tests for provider inference.
- `README.md`, `openwiki/cli/usage.md`, `openwiki/operations/credentials-and-updates.md`: document inference behavior and required env vars.

## Verification
- `bun run typecheck` ✅
- `bun test` ✅ (91 passed)
- `bun run build` ✅
- `bun run format:check` ✅
- `bun run lint` ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ OPENAI_COMPATIBLE_BASE_URL=https://your-gateway.example.com/v1
 OPENWIKI_MODEL_ID=your-gateway-model-name
 ```
 
+You can also omit `OPENWIKI_PROVIDER` for non-interactive runs: if
+`OPENAI_COMPATIBLE_API_KEY` and `OPENAI_COMPATIBLE_BASE_URL` are set (and no
+`OPENROUTER_API_KEY` is present), OpenWiki will infer the `openai-compatible`
+provider automatically.
+
 Base URLs (and all credentials) can be set in your environment or stored in `~/.openwiki/.env`.
 
 If there's an inference provider or model you'd like to see added, please open a PR!
--- a/openwiki/cli/usage.md
+++ b/openwiki/cli/usage.md
@@ -66,7 +66,12 @@ Providers and their model options are defined in `PROVIDER_CONFIGS` in `src/cons
 | openai-compatible | `OPENAI_COMPATIBLE_API_KEY` | `OPENAI_COMPATIBLE_BASE_URL` (required) | custom model ID only                                                  |
 | anthropic         | `ANTHROPIC_API_KEY`         | (default, or `ANTHROPIC_BASE_URL`)      | Haiku, Sonnet, Opus                                                   |
 
-The default provider is `openrouter`. `resolveConfiguredProvider()` picks the provider from `OPENWIKI_PROVIDER`, falling back to openrouter if `OPENROUTER_API_KEY` is set, then to `DEFAULT_PROVIDER`.
+The default provider is `openrouter`. `resolveConfiguredProvider()` picks the provider in this order:
+
+1. `OPENWIKI_PROVIDER` if it is set and valid.
+2. `openrouter` if `OPENROUTER_API_KEY` is present (preserves the historical default).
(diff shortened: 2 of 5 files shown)


colifran:
Thanks for the PR! I think we can close due to #48 being merged which added credential-based provider inference. Feel free to re-open if I'm incorrect and ping me!

PR #160, opened 2026-07-07
feat: support Claude Code OAuth token auth for the anthropic provider

holdennguyen:
## What

Adds Claude Code OAuth token support to the `anthropic` provider: `CLAUDE_CODE_OAUTH_TOKEN` (generated with `claude setup-token`) now works as an alternative to `ANTHROPIC_API_KEY`. `ANTHROPIC_API_KEY` takes precedence when both are set.

## Why

Users with a Claude subscription can authenticate Claude Code via a long-lived OAuth token without provisioning a raw Anthropic API key. OpenWiki previously hard-required `ANTHROPIC_API_KEY` for the anthropic provider — the interactive setup would prompt for an API key and non-interactive runs would abort even when a valid OAuth token was available in the environment.

## How

- **`src/constants.ts`** — new `CLAUDE_CODE_OAUTH_TOKEN_ENV_KEY`; `ProviderConfig` gains an optional `oauthTokenEnvKey` (set only on `anthropic`); new helpers `getProviderOauthTokenEnvKey`, `getProviderCredentialEnvKeys`, `hasProviderCredential`, and `describeProviderCredentialEnvKeys` so "does this provider have a usable credential?" is answered in one config-driven place.
- **`src/agent/index.ts`** — in `createModel`, when `ANTHROPIC_API_KEY` is unset but the OAuth token is present, the underlying Anthropic client is constructed through ChatAnthropic's `createClient` hook with:
  - `authToken` → sends `Authorization: Bearer <token>`,
  - `defaultHeaders: {"anthropic-beta": "oauth-2025-04-20"}` (required for OAuth bearer auth on `/v1/messages`),
  - `apiKey: null` → prevents the SDK from also sending an `x-api-key` header, which the API rejects when both are present.

  The `createClient` hook is used because ChatAnthropic always overrides `apiKey` on the client options it builds, so a bearer-token client cannot be produced through the normal `apiKey`/`clientOptions` path. `ANTHROPIC_BASE_URL` routing continues to work in both paths. `ensureProviderKey` accepts either credential.
- **`src/credentials.tsx` / `src/cli.tsx`** — interactive setup treats a present OAuth token as a completed "Provider key" step (no API-key prompt), and the non-interactive guards accept it; error messages now name both env keys (`ANTHROPIC_API_KEY or CLAUDE_CODE_OAUTH_TOKEN`).
- **`src/env.ts`** — the token is a managed env key, so it persists to `~/.openwiki/.env` and automatically appears (masked) in the credential diagnostics panel and the agent's debug env dump, like every other credential.
- **`package.json`** — adds `@anthropic-ai/sdk@^0.103.0` as a direct dependency (previously transitive via `@langchain/anthropic`; the same version already in the lockfile, needed for the `createClient` construction under pnpm's strict layout).
- **Docs** — README, `openwiki/cli/usage.md`, `openwiki/operations/credentials-and-updates.md`, and `openwiki/agent/workflow.md` updated to document the new variable and precedence.

## How I tested it

- `pnpm run format`, `pnpm run lint`, `pnpm test` — all green (90 tests, including new unit tests in `test/constants.test.ts` covering credential resolution: OAuth token accepted for anthropic only, empty strings ignored, API-key-only providers unaffected).
- `tsc --noEmit` clean.
- Wire-level smoke test: constructed the model exactly as the new branch does with a capturing `fetch` — verified the request carries `Authorization: Bearer <token>` and `anthropic-beta: oauth-2025-04-20`, and does **not** carry `x-api-key`.
- End-to-end locally: built and linked the CLI, ran `openwiki --init` with only `CLAUDE_CODE_OAUTH_TOKEN` set — setup shows the provider key step as done ("available from environment") and no API-key prompt appears; with neither credential set, non-interactive runs fail with `ANTHROPIC_API_KEY or CLAUDE_CODE_OAUTH_TOKEN is required`.

## Notes

- Scope is a single concern: OAuth-token authentication for the anthropic provider. No other provider's behavior changes; the interactive prompt still saves an `ANTHROPIC_API_KEY` for users who go that route.
- The token is masked in diagnostics exactly like other secrets (`sk-ant...xxxx` preview).


colifran:
Unfortunately I think this is against Anthropic's ToS

PR #163, opened 2026-07-07
Feat: Add custom model header configuration

hetaoBackend:
## Summary

Adds `OPENWIKI_MODEL_HEADERS` so users can attach extra string-valued HTTP headers to model calls through environment configuration or `~/.openwiki/.env`.

## Details

- Parses and validates `OPENWIKI_MODEL_HEADERS` as a JSON object with string header values.
- Passes configured headers to OpenAI-compatible/OpenAI, Anthropic, and OpenRouter model clients.
- Treats model header configuration and individual header values as sensitive in diagnostics and debug output.
- Documents the new configuration option in the README and OpenWiki internal docs.

## Validation

- `pnpm format:check`
- `pnpm typecheck`
- `pnpm lint:check`
- `pnpm test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,18 @@ OPENWIKI_MODEL_ID=your-gateway-model-name
 
 Base URLs (and all credentials) can be set in your environment or stored in `~/.openwiki/.env`.
 
+### Custom model request headers
+
+If your model gateway requires additional request headers, set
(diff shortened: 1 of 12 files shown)


hetaoBackend:
@colifran when you have a chance, could you please review this PR? The GitHub Actions run is currently stuck at `action_required` with no jobs because it needs collaborator approval for the forked PR workflow. Thanks!

PR #164, opened 2026-07-07
fix: prevent 400 on non-PDF binary tool results (anthropic)

### LINKED ISSUE #33: Anthropic 400 INVALID_TOOL_RESULTS: files with unmapped/no extension (Makefile, Dockerfile, uv.lock) are sent as base64 documents

### LINKED ISSUE #65: read_file crashes with Anthropic 400 on extensionless files (Dockerfile, LICENSE, CHANGELOG, ...)

### LINKED ISSUE #114: v0.0.1 (Anthropic provider) crashes on any Python repo with __pycache__: binary file reads become invalid document blocks; one malformed tool result kills the run

### LINKED ISSUE #134: Full-repo generation fails with Anthropic 400: non-PDF files sent as base64 `document` blocks
## Summary

`openwiki "Generate documentation for this repository"` fails with an Anthropic API
`400 invalid_request_error` on repositories that contain **large text files** and/or **binary
files** (including gitignored caches). OpenWiki base64-encodes files it reads and sends them to
the Anthropic Messages API as `document` content blocks, but the Anthropic API only accepts a
base64 `document` source when `media_type` is `application/pdf`. Any other file type produces:

```
messages.N.content.M.tool_result.content.0.document.source.base64.media_type:
Input should be 'application/pdf'
```

This aborts the entire run.

## Environment

- OpenWiki `v0.0.1` (`npm install -g openwiki`)
- Provider: **Anthropic**, model `claude-sonnet-5`
- Node `v26.4.0`, npm `11.17.0`
- macOS (Darwin 25.x, Apple Silicon)

## Reproductions

Reproduced cleanly on two distinct triggers:

**A) Large text files (Markdown docs).** A full-repo run on a repo whose largest files are ~49 KB
and ~25 KB Markdown design docs fails partway through, after the agent has read several files
(observed failing at both `messages.10...` and `messages.24...` on separate runs).

**B) Binary / gitignored cache files (`.pyc`).** Scoping to a subdirectory of pure Python source
initially *succeeded*. After I created a virtualenv and ran the test suite (which produced
`__pycache__/*.pyc` bytecode), re-running the same scoped command failed while reading
`__pycache__/main.cpython-312.pyc` (a ~9 KB compiled binary):

```
Error 400 {"type":"error","error":{"type":"invalid_request_error",
"message":"messages.8.content.3.tool_result.content.0.document.source.base64.media_type:
Input should be 'application/pdf'"}}

Troubleshooting URL:
https://docs.langchain.com/oss/javascript/langchain/errors/INVALID_TOOL_RESULTS/
```

(Anthropic `request_id`s for each failure are available if useful.)

## Expected vs. actual

- **Expected:** OpenWiki reads text files as text and skips (or safely handles) binary and
  gitignored files; generation completes regardless of file size/type.
- **Actual:** larger text files *and* binary/cache files are encoded as base64 `document` blocks
  with a non-PDF `media_type`, which the Anthropic API rejects with a 400, aborting the run.

## Likely root cause / suggested fix

The Anthropic Messages API only permits a `document` block whose `source.type` is `base64` when
`media_type` is `application/pdf`. Plain text should be sent as a `text` content block, or as a
`document` with `source.type: "text"` and `media_type: "text/plain"`.

OpenWiki's file-reading tool appears to build a base64 `document` block for files above some size
threshold (with a text or generic media type). Two fixes seem warranted:
1. Route non-PDF files through a text content block / `text` document source; reserve base64
   `document` blocks for actual PDFs.
2. Respect `.gitignore` and skip binary / cache directories (`__pycache__`, `.venv`,
   `node_modules`, etc.) so bytecode and vendored binaries are never read in the first place.

## Impact

Two very common repo shapes break a full run today: repos with sizeable Markdown docs, and any
repo where a build/test step has produced compiled artifacts. The only workaround is to scope
generation to directories that contain no large or binary files — and because the agent explores
files non-deterministically, even that isn't fully reliable.

holdennguyen:
## What

Fixes the Anthropic 400 `invalid_request_error` that aborts a whole run when the agent reads any non-PDF binary file:

```
messages.N.content.M.tool_result.content.0.document.source.base64.media_type:
Input should be 'application/pdf'
```

Fixes #134, fixes #114, fixes #65, fixes #33.

## Why

The deep agent's `read_file` tool returns non-text files (compiled bytecode like `__pycache__/*.pyc`, archives, extensionless files such as `Dockerfile`/`LICENSE`/lock files, unsupported image formats, audio, video) as base64 `file`/`image`/`audio`/`video` content blocks carrying the file's real MIME type. `@langchain/anthropic` serializes a `file` block as an Anthropic base64 `document` block — but the Anthropic Messages API only accepts `application/pdf` there (and only JPEG/PNG/GIF/WebP for `image` blocks). One read of such a file kills the entire generation run with a 400, which is why full-repo `--init` runs on ordinary Python or Docker repos fail today (see the linked issues — four separate reports of the same root cause).

## How

- **`src/agent/anthropic-compat.ts`** (new) — `anthropicMultimodalCompatMiddleware`, a langchain agent middleware using the `wrapModelCall` hook. Before each model call it rewrites tool-result content blocks the Anthropic API rejects into a short text placeholder (`[file content omitted: application/octet-stream, ~9216 bytes. …treat it as unreadable…]`):
  - `file` blocks with any MIME type other than `application/pdf`
  - `image` blocks with MIME types outside JPEG/PNG/GIF/WebP
  - all `audio` and `video` blocks

  Supported blocks (PDF documents, supported images), text blocks, string-content tool messages, and non-tool messages pass through untouched — messages are only copied when a block actually changes.
- **`src/agent/index.ts`** — attaches the middleware in `createDeepAgent`, for the `anthropic` provider only. Other providers are unaffected.

The middleware approach fixes this at the model boundary regardless of which tool produced the block, so it also covers future tools that might return binary content. Skipping binary/gitignored paths at read time (suggested in #134) would be a good complementary improvement, but it belongs to the filesystem backend upstream (`deepagents`) and wouldn't cover the large-text/document cases; this PR makes runs survive either way.

## How I tested it

- New unit tests in `test/anthropic-compat.test.ts` (6 tests): non-PDF `file` block replaced with placeholder (base64 dropped from payload), PDF + supported image pass through by identity, unsupported image/audio/video replaced, mixed text+binary content keeps text intact, tool-message identity fields (`tool_call_id`, `name`, `status`) preserved on rewrite, string-content tool messages and human/AI messages untouched.
- `pnpm run format`, `pnpm run lint`, `pnpm test` (92 tests), `tsc --noEmit` — all green.
- Wire-level verification against `ChatAnthropic` with a capturing `fetch`: a `ToolMessage` carrying a binary `file` block serializes to exactly the rejected payload from the linked issues (`{"type":"document","source":{"type":"base64","media_type":"application/octet-stream",...}}`); after `sanitizeMessagesForAnthropic` the same message serializes to a plain `text` block with the placeholder — nothing the API rejects.
- The original 400 was hit in a real `openwiki` run with the Anthropic provider (same signature as #134/#114); I'll follow up on this PR with a live end-to-end run on that repo using this patch.


src/agent/anthropic-compat.ts
```diff
@@ -0,0 +1,121 @@
...
+    content,
+    id: message.id,
+    name: message.name,
+    response_metadata: message.response_metadata,
+    status: message.status,
+    tool_call_id: message.tool_call_id,
+  });
+}
+
+function isBlockUnsupportedByAnthropic(block: unknown): boolean {
```
bracesproul, line 99: i believe we should be able to properly type `block` here

src/agent/anthropic-compat.ts
```diff
@@ -0,0 +1,121 @@
...
+    return (
+      typeof mimeType === "string" &&
+      !ANTHROPIC_SUPPORTED_IMAGE_MIME_TYPES.has(mimeType)
+    );
+  }
+
+  return false;
+}
+
+function describeOmittedBlock(block: unknown): string {
```
bracesproul: add types

bracesproul (commented):
thanks for putting this up! this is a great start, but we should be parsing and including the contents of text files instead of saying they cant be read

e.g. for a `.gitignore` file, we can easily read its contents, so instead of returning a cant read file error, we should parse it and return its contents

src/agent/anthropic-compat.ts
holdennguyen: Done in [fcc83a9](https://github.com/langchain-ai/openwiki/pull/164/changes/fcc83a96b4739b28294e4542b3fc16053cc619b7) - replaced the `as MultimodalContentBlock` cast with an `isMultimodalContentBlock(block): block is MultimodalContentBlock` type guard. `isBlockUnsupportedByAnthropic` now narrows `block` and destructures it directly, no cast.

src/agent/anthropic-compat.ts
holdennguyen: Done in [fcc83a9](https://github.com/langchain-ai/openwiki/pull/164/changes/fcc83a96b4739b28294e4542b3fc16053cc619b7) - `describeOmittedBlock` now takes `MultimodalContentBlock` directly (the call site is already guarded by `isBlockUnsupportedByAnthropic`), so no cast. Same guard also removes the cast in `decodeTextBlock`.

colifran:
Hey @bracesproul and @holdennguyen. I think this actually needs to be a fix upstream in deepagentsjs. This aligns with python and should resolve: https://github.com/langchain-ai/deepagentsjs/pull/656.

colifran:

PR #167, opened 2026-07-07
fix: require custom model for openai-compatible provider

shadril238:
## Motivation

`openai-compatible` endpoints do not have a universal default model name. The previous behavior allowed OpenWiki to fall back to the global OpenRouter default model when no `OPENWIKI_MODEL_ID` was configured, which could send an invalid model ID to an OpenAI-compatible gateway or make the active provider/model pairing misleading. This change makes the configuration contract explicit: providers without preset model options require the user to provide a model ID.

 ## Summary

  - Require an explicit model ID for providers that do not define preset model options, including `openai-compatible`.
  - Prevent `openai-compatible` from silently falling back to the OpenRouter default model.
  - Clear stale saved model IDs when switching to a custom-only provider.
  - Update CLI/setup messaging so users are told when a custom model ID is required.
  - Update constants coverage for the corrected default-model behavior.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -395,6 +395,15 @@ function resolveModelId(
     options.modelId ??
     process.env[OPENWIKI_MODEL_ID_ENV_KEY] ??
     getDefaultModelId(provider);
+
+  if (!rawModelId) {
+    throw new Error(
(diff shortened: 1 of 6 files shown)


colifran:
Thanks for the PR! Believe #327 which added logic for resolveModelId to throw when no preset models

PR #169, opened 2026-07-07
fix: read text files with unmapped extensions as text

### LINKED ISSUE #168: Text files with unrecognized extensions (terraform.tfvars, .tf, .hcl, ...) are treated as binary and unreadable
## Summary

## Reproduction

## Root cause

```js
  }
  ...
```

holdennguyen:
## What

Makes the agent able to read plain-text files whose extensions aren't in the MIME map — `terraform.tfvars`, `.tf`, `.hcl`, `.lock`, `.conf`, and similar. Today those files are classified as binary by extension lookup alone, wrapped in a base64 block the model cannot read, and (on the Anthropic provider) rejected by the API.

Fixes #168.

## Why

`LocalShellBackend.read()` (deepagents) decides text-vs-binary purely from `getMimeType(filePath)`. Any unmapped extension falls back to a non-text MIME type, so the backend returns the whole file as a `Uint8Array` without ever looking at the bytes. For a Terraform repository this means the variable definitions at the core of the codebase can't be documented at all. Related earlier reports (#33, #65, #134, #114) focused on the crash this caused on Anthropic — #164 stopped the crash, but the files remained unreadable; this PR fixes the readability itself, for **every provider**.

## How

- **`src/agent/text-sniffing-backend.ts`** (new) — `TextSniffingLocalShellBackend`, a `LocalShellBackend` subclass. When the base class's `read()` returns binary content, it sniffs the bytes: valid UTF-8 with no NUL bytes → returned as text (`mimeType: "text/plain"`), applying the same 0-indexed `offset` / `limit` line semantics as the base class's text path. Anything that fails the sniff (NUL bytes, invalid UTF-8 sequences) keeps the existing binary behavior, so images/archives/bytecode are unaffected.
- **`src/agent/index.ts`** — `createDeepAgent` now uses the subclass in place of `LocalShellBackend`. No other behavior changes.

The base class already downloads the full file on its binary path, so the sniff adds no extra I/O. A longer-term fix belongs upstream in `deepagents` (content sniffing inside `LocalShellBackend`, or an extended MIME map); this wrapper unblocks OpenWiki users now without forking upstream behavior for known file types.

## How I tested it

- New tests in `test/text-sniffing-backend.test.ts` (11 tests): UTF-8 decode incl. multi-byte (accents/CJK/emoji), NUL-byte and invalid-sequence rejection, offset/limit slicing — plus behavioral tests against a real `TextSniffingLocalShellBackend` on a temp directory: a `terraform.tfvars` reads back as text with correct content, offset/limit works on sniffed files, known text extensions still use the base text path, and a PNG still comes back binary.
- `pnpm run format`, `pnpm run lint`, `pnpm test` (97 tests), `tsc --noEmit` — all green.

## Notes

- Scope: one concern — text-file readability at the backend boundary. Complements #164 (Anthropic 400 guard) but is independent of it; either merges cleanly without the other.

### DIFF
--- a/openwiki/agent/workflow.md
+++ b/openwiki/agent/workflow.md
@@ -13,7 +13,7 @@ The documentation agent is implemented in `src/agent/`. It takes a command (`cha
 5. Snapshot the current `openwiki/` content hash (before the run).
 6. Build the system prompt and user prompt.
 7. Create the provider-specific model client (`ChatAnthropic`, `ChatOpenRouter`, or `ChatOpenAI`).
-8. Create a DeepAgents `LocalShellBackend` rooted at the repository with a SQLite checkpointer.
+8. Create a DeepAgents backend rooted at the repository with a SQLite checkpointer. OpenWiki uses `TextSniffingLocalShellBackend` (`src/agent/text-sniffing-backend.ts`), a `LocalShellBackend` subclass that content-sniffs files the extension-based MIME lookup would treat as binary — valid UTF-8 files with unmapped extensions (terraform.tfvars, .tf, .hcl, .lock, ...) are returned as readable text instead of base64 blocks.
 9. Stream messages and tool events back to the CLI.
 10. For `init` and `update`, compare the post-run content snapshot to the pre-run snapshot. Write `openwiki/.last-update.json` **only if the content changed**.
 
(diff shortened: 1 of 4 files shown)


colifran:
Hey @holdennguyen. We're working on an upstream fix for this in deepagentsjs. This should resolve: https://github.com/langchain-ai/deepagentsjs/pull/656

colifran:

PR #170, opened 2026-07-07
fix: sniff text files with unmapped extensions

### LINKED ISSUE #168: Text files with unrecognized extensions (terraform.tfvars, .tf, .hcl, ...) are treated as binary and unreadable

shadril238:
## Summary

Fixes #168.

  - Add an OpenWiki-side `TextSniffingLocalShellBackend` wrapper around DeepAgents `LocalShellBackend`.
  - Reclassify binary-looking read results as `text/plain` when the bytes decode as valid UTF-8 and contain no NUL bytes.
  - Preserve existing binary behavior for invalid UTF-8 or NUL-containing files.
  - Apply the same line `offset`/`limit` windowing when sniffed content is returned as text.
  - Add focused tests for Terraform-style text content, invalid UTF-8, NUL-containing binary data, and pagination behavior.

## Motivation

DeepAgents currently decides text-vs-binary from MIME/extension metadata. Plain-text files with unmapped extensions, such as `terraform.tfvars`, `.tf`, `.hcl`, `.conf`, `.lock`, or `.env.example`, can be returned as binary data instead of readable text. That prevents the OpenWiki agent from documenting important repository files even though their contents are normal UTF-8 text. This wrapper unblocks OpenWiki users while keeping genuine binary files on the existing binary path.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -5,10 +5,11 @@ import { ChatAnthropic } from "@langchain/anthropic";
 import { SqliteSaver } from "@langchain/langgraph-checkpoint-sqlite";
 import { ChatOpenAI } from "@langchain/openai";
 import { ChatOpenRouter } from "@langchain/openrouter";
-import { createDeepAgent, LocalShellBackend } from "deepagents";
+import { createDeepAgent } from "deepagents";
 import { DEBUG_ENV_KEYS, loadOpenWikiEnv, openWikiEnvDir } from "../env.js";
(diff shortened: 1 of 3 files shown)


colifran:
Hi @shadril238 - thank you for the PR. We fixed this upstream in deepagentsjs: https://github.com/langchain-ai/deepagentsjs/pull/656. This should be resolved as part of the recent v0.0.3 release of OpenWiki.

PR #173, opened 2026-07-07
test: add coverage for src/env.ts runtime behavior

varaprasadreddy9676:
## What

Adds the first coverage for the runtime behavior of `src/env.ts`. Pure tests — **no production code changes**.

## Why

`src/env.ts` currently has coverage only for the pure `parseEnv`/`formatEnv` serializers (`test/env.test.ts`). The three functions that actually carry the credential-loading, deprecation, and diagnostic behavior — `loadOpenWikiEnv`, `saveOpenWikiEnv`, and `getCredentialDiagnostics` — had no tests. That behavior is exactly the kind of thing that's easy to break silently (e.g. a key accidentally landing in or leaving the deprecated list, file permissions drifting, a secret leaking into a diagnostic preview).

## Coverage added

`test/env-behavior.test.ts` — 14 tests, each isolated to a throwaway temp `HOME` so no real `~/.openwiki/.env` is ever touched:

**`loadOpenWikiEnv`**
- loads a saved managed key into `process.env`
- does not overwrite a key already present in `process.env` (precedence)
- pins the existing behavior of dropping deprecated OpenAI keys (`OPENAI_BASE_URL`/`OPENAI_ORG_ID`/`OPENAI_PROJECT`) — so any future un-deprecation is a deliberate decision that updates this expectation

**`saveOpenWikiEnv`**
- round-trips a value back through `loadOpenWikiEnv`
- writes the env file with `0600` permissions
- strips stale deprecated keys from the persisted file
- seeds `process.env` with the saved value immediately

**`getCredentialDiagnostics`**
- includes the provider and each credential key, in display order, with no duplicates
- reports an unset key as unset with no warnings
- masks a secret value rather than echoing it
- surfaces a non-secret base URL verbatim
- flags an invalid model ID and an invalid provider
- prefers `process.env` over the file when both are set

## How I tested

```
pnpm run build        # OK
pnpm test             # 100 passed (86 existing + 14 new); existing suites unchanged
pnpm run lint:check   # OK
pnpm run format:check # OK
```

No existing tests were modified.


colifran (approved):
Increased test coverage is always welcome! Thanks @varaprasadreddy9676!

PR #178, opened 2026-07-07
feat: support OpenTelemetry (OTLP) trace export

### LINKED ISSUE #177: Support OpenTelemetry (OTLP) trace export — agent runs currently only emit the LangSmith protocol
## Summary

There is currently no way to trace openwiki's agent runs to an OpenTelemetry backend (Langfuse, Grafana Tempo, generic OTLP collectors). The bundled `langsmith` SDK (0.7.16) ships experimental OTel support, and it works for `traceable()`-wrapped functions — but the actual agent traffic, which flows through `@langchain/core` 1.2.1 callbacks, is never routed to the OTLP exporter. Only the proprietary LangSmith multipart protocol is emitted.

This matters for enterprise/self-hosted setups: we run openwiki in GitLab CI against on-prem models (OpenAI-compatible gateway), and tracing must stay on-prem too — LangSmith cloud is not an option.

## What I tried (empirical, reproducible)

Since openwiki never calls `initializeOTEL()`, I injected it from outside without touching the code:

1. Installed the OTel peer deps (`@opentelemetry/api`, `sdk-trace-base`, `exporter-trace-otlp-proto`, `context-async-hooks`) so they resolve from the bundled `langsmith`.
2. Wrote a preload module that imports `langsmith/dist/experimental/otel/setup.js` **from openwiki's own node_modules** (same ESM instance, shared singletons), calls `initializeOTEL()`, and force-flushes the provider on `beforeExit` and on an intercepted `process.exit`.
3. Ran openwiki via `NODE_OPTIONS="--import <preload>"` with:
   `LANGSMITH_API_KEY=<dummy> LANGCHAIN_TRACING_V2=true LANGSMITH_TRACING=true LANGSMITH_TRACING_MODE=otel OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318/v1/traces OTEL_BSP_SCHEDULE_DELAY=200`
   against a local OTLP sink.

## Results

| Scenario | Result |
|---|---|
| `traceable()`-wrapped function + `initializeOTEL()` + `forceFlush()` | ✅ OTLP protobuf spans received at the sink |
| `openwiki -p "…"` full run, same setup (incl. 200 ms BSP flush + exit-hook flush) | ❌ zero OTLP export |
| `openwiki -p "…"` classic mode with `LANGSMITH_ENDPOINT` pointed at the local sink | ✅ full runs received (`POST /runs/multipart`, ~137 KB) |

So tracing itself is active and rich — but the LangChain v1 callback path never reaches the langsmith OTel exporter, making OTLP export impossible today without upstream changes.

## Feature request

- Have openwiki call `initializeOTEL()` (and ship the four `@opentelemetry/*` peer deps) when `LANGSMITH_TRACING_MODE=otel` / `LANGSMITH_OTEL_ENABLED=true` is set, or expose an explicit `OPENWIKI_OTEL_ENABLED` flag.
- Ensure agent runs (LangChain/LangGraph callbacks) are exported through the OTLP exporter, not only `traceable()` spans.
- Flush the tracer provider before exit — `--print`/CI runs are short-lived and batched spans are otherwise lost (this bit me even in the working `traceable()` case).

With that in place, tracing to Langfuse would be a pure env-var affair (`OTEL_EXPORTER_OTLP_ENDPOINT=https://<langfuse>/api/public/otel/v1/traces` + basic-auth header), and the same for any OTLP backend.

Two small quirks worth documenting meanwhile:
- `LangSmithOTLPTraceExporter` uses `OTEL_EXPORTER_OTLP_ENDPOINT` **as-is** (no `/v1/traces` appended), unlike the OTel spec default behavior.
- Tracing only activates when `LANGSMITH_API_KEY` is set, even in OTel mode where it shouldn't be required.

## Environment

- openwiki 0.0.1 (npm), macOS (darwin), Node v25.1.0 (also reproduced on Node 22)
- Bundled: `langsmith` 0.7.16, `@langchain/core` 1.2.1, `langchain` 1.5.2, `@langchain/langgraph` 1.4.7
- Provider: `openai` with custom `OPENAI_BASE_URL` (LiteLLM gateway, on-prem model)

Happy to test a patch or contribute a PR if the approach above (init + peer deps + exit flush) sounds right to you.

Neal006:
## What

Adds an OpenTelemetry (OTLP) export path for openwiki's agent-run traces. When `LANGSMITH_TRACING_MODE=otel` (or the legacy `OTEL_ENABLED` / `LANGSMITH_OTEL_ENABLED`) is set, openwiki initializes the OTel tracer provider on startup and flushes it before exit. The bundled `langsmith` SDK then translates LangChain/LangGraph callback runs into OTLP spans and ships them to `OTEL_EXPORTER_OTLP_ENDPOINT`.

Closes #177.

## Why

Today openwiki only emits the native LangSmith protocol, so there is no way to trace runs to a self-hosted OpenTelemetry backend — a blocker for on-prem/enterprise users. `langsmith` already knows how to emit OTLP and honor `OTEL_EXPORTER_OTLP_ENDPOINT`/`OTEL_EXPORTER_OTLP_HEADERS`; the missing pieces were (1) calling `initializeOTEL()` to install the global tracer provider and (2) flushing it before the process exits — without the flush, short-lived (CI) runs drop their final spans.

## How it works

- `src/otel.ts`: `isOtelTracingMode()` mirrors the SDK's own precedence (`LANGSMITH_TRACING_MODE` > legacy flags). `initOpenTelemetry()` dynamically imports `langsmith/experimental/otel/setup` only when OTel mode is on (keeps the optional peers off the default load path) and returns a flush handle. `shutdownOpenTelemetry()` force-flushes, bounded by a 5s timeout so an unreachable collector can't hang the CLI on exit.
- `src/agent/index.ts`: init at the top of `runOpenWikiAgent` (before any tracer client is built) and flush in the existing `finally`, so both the interactive and `--print` paths are covered. Init/flush failures are debug-logged, never fatal to a doc run.

## Dependencies

Adds `langsmith` as a **direct** dependency pinned `^0.7.10`. `initializeOTEL` records the provider in a module-level singleton the tracer's client reads, so both must resolve to the **same** `langsmith` instance; the pin dedupes with the copy `langchain`/`@langchain/core` already resolve (both allow `>=0.5.0 <1.0.0`). Also adds the four required OTel packages (`@opentelemetry/api`, `sdk-trace-base`, `exporter-trace-otlp-proto`, `context-async-hooks`).

Note: `initializeOTEL` carries an `@deprecated` JSDoc pointing at `wrapAISDK` (the Vercel AI SDK path), but it remains the canonical setup entry for routing LangChain callbacks through OTLP.

## How tested

- `pnpm run lint:check`, `pnpm run typecheck`, `pnpm run build`, `pnpm test` — all pass (added `test/otel.test.ts`, 5 cases covering the mode-detection precedence; 91 tests total).
- End-to-end smoke against the real `langsmith` import: OTel-disabled → `null` handle; `LANGSMITH_TRACING_MODE=otel` → a real `BasicTracerProvider`; `forceFlush()` resolves immediately with no reachable collector (confirms exit is not blocked).
- `README.md` gains an "OpenTelemetry (OTLP) tracing" section documenting the env vars.


Neal006:
Hey @bracesproul 👋
 pened #178 to add OpenTelemetry (OTLP) trace export (closes #177). It's tightly scoped to one concern per CONTRIBUTING: gated on LANGSMITH_TRACING_MODE=otel, ships the OTel peers, and flushes the provider before exit so short-lived CI runs don't drop spans.

Ran the full check suite locally (lint, typecheck, build, tests on Node 20 + 22, added test/otel.test.ts) and verified the init into flush path end-to-end against the real langsmith OTel exporter. I added langsmith as a direct dep pinned to dedupe with the copy LangChain resolves, since initializeOTEL relies on a shared module-level singleton, happy to switch the OTel packages to optionalDependencies if you'd prefer.
No rush given the review queue, just flagging it's ready whenever you get to it. 🙏

abenetton:
Any update on this?

Neal006:
@abenetton Still open and waiting on a maintainer review — no feedback from the team yet. The branch has drifted out of sync with main since July, so I'll rebase and resolve the conflicts to get it mergeable again. Functionality is unchanged from the original PR description; happy to make any adjustments a maintainer wants.

PR #147, opened 2026-07-06
feat: add nebius token factory provider support

amrrs:
Added Nebius Token Factory as one of the official model providers and has been added through via `@langchain/openai`  OpenAI-compatible API path. 

Unfortunately, the langchain-nebius integration doesn't seem to be available for javascript (TS) to go that route. 

Tests:

- `pnpm test -- --run test/constants.test.ts`
  - Passed: `6` test files, `81` tests
- `pnpm run typecheck`
  - Passed with no TypeScript errors

PR #179, opened 2026-07-07
feat: add google vertex ai (claude) provider

SadriG91:
## What

Adds a `vertex` provider that runs Anthropic's Claude models through **Google Vertex AI**, authenticated with Google Application Default Credentials — no API key:

```bash
OPENWIKI_PROVIDER=vertex
GOOGLE_CLOUD_PROJECT=your-gcp-project
GOOGLE_CLOUD_LOCATION=global   # optional, defaults to global
```

Auth works through the standard ADC chain: `GOOGLE_APPLICATION_CREDENTIALS` (service-account key), `gcloud auth application-default login`, or workload identity.

## Why

Organizations on Google Cloud consume Claude through Vertex AI (billing, quotas, IAM, data residency) rather than a direct Anthropic API key — today OpenWiki can't use those deployments. No existing issue or PR covers Claude-on-Vertex (#154/#36 add **Gemini** models, a different provider path).

## How

- Reuses the existing `ChatAnthropic` class with its documented `createClient` option returning an `AnthropicVertex` client from `@anthropic-ai/vertex-sdk` — the same message/tool-calling code path as the `anthropic` provider; only authentication differs.
- `ProviderConfig.apiKeyEnvKey` becomes optional. A new `getMissingProviderEnvKey()` helper powers the agent's credential gate, both CLI non-interactive gates, and `needsCredentialSetup()`, so key-less providers are handled uniformly (with an ADC hint appended to error messages).
- Interactive onboarding: vertex skips the API-key step and instead prompts for `GOOGLE_CLOUD_PROJECT` (required) and `GOOGLE_CLOUD_LOCATION` (optional; Enter keeps the soft `global` default — nothing persisted).
- `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION` / `GOOGLE_APPLICATION_CREDENTIALS` join `MANAGED_ENV_KEYS` and show as non-secret values in credential diagnostics.
- `isValidModelId` now accepts `@` so Vertex's versioned IDs validate (`claude-haiku-4-5@20251001`); Sonnet/Opus use the dateless IDs.
- The invalid-provider message in `cli.tsx` is now derived from `SELECTABLE_OPENWIKI_PROVIDERS` (the hardcoded list had gone stale — it was missing `openai-compatible`). Heads-up: #154 makes the same tidy-up, so whichever lands second has a trivial conflict there.

## How I tested it

- `pnpm run format`, `pnpm run lint`, and `pnpm test` all pass (98 tests, including new coverage for the vertex config, the `getMissingProviderEnvKey`/`resolveProviderLocation`/`providerRequiresApiKey` helpers, `@`-versioned model IDs through both `isValidModelId` and `--modelId` parsing, and the managed env keys).
- Non-interactive gate: running with `OPENWIKI_PROVIDER=vertex` and no `GOOGLE_CLOUD_PROJECT` exits 1 naming the env key plus the ADC hint.
- Interactive onboarding walked end-to-end in a sandbox `HOME`: no API-key step, required project step (rejects empty/whitespace input), optional location step, the three Claude presets in the model menu, and correct `~/.openwiki/.env` contents afterwards.
- Real end-to-end: one-shot and full-agent runs against a live GCP project using `gcloud` ADC on the `global` endpoint, with `claude-haiku-4-5@20251001` responding through the deepagents tool-calling path.

README and the repo's `openwiki/` docs are updated to match (same file set as the openai-compatible provider PR, d75a9f3).


SadriG91:
Heads-up on some convergence: the AWS Bedrock PRs (#185, #96, and #40 for issue #56) generalize the same thing this PR does for Vertex — credential gating for providers that don't authenticate with a single API key.

- **#185** models it as an explicit `auth: "api-key" | "aws"` discriminant on the provider config, with region-aware validation.
- **This PR** models it by making `apiKeyEnvKey` optional and deriving the gate from whichever required env keys the provider declares (`getMissingProviderEnvKey()`), with a per-provider credential hint for the error paths.

The two shapes enable the same thing, so whichever pattern maintainers prefer for keyless/cloud-credential providers, this change adapts trivially — happy to rebase onto the Bedrock abstraction if one of those lands first (or they can rebase onto this one; the vertex entry would translate to `auth: "gcp"` in a one-line change). Also flagging #154 (Gemini), which shares the `GOOGLE_CLOUD_PROJECT`/`GOOGLE_CLOUD_LOCATION` env keys and makes the same `SELECTABLE_OPENWIKI_PROVIDERS`-derived error-message tidy-up in `cli.tsx`.

ppsplus-bradh:
@SadriG91 nice PR — we independently landed on the exact same `ChatAnthropic` + `createClient` → `AnthropicVertex` bridge, which is reassuring. Flagging the overlap and offering to consolidate so the maintainers don't have to reconcile two Vertex code paths.

**How it overlaps with #154:** #154 (Gemini + Gemini Enterprise) also does Claude-on-Vertex, but as one part of a broader `gemini-enterprise` provider that routes by model *family*: Gemini/Gemma → `ChatGoogle` (native `generateContent`), Claude → `AnthropicVertex` (rawPredict), and Llama/Mistral/DeepSeek/Qwen → the Vertex OpenAI-compatible endpoint — dispatched from the model ID via `resolveVertexSurface()`. So #179's Claude support is essentially a subset of #154's routing, and both carry the same `@`-in-`isValidModelId` fix and the `SELECTABLE_OPENWIKI_PROVIDERS` tidy-up you already flagged as a conflict.

**Where your PR is cleaner:** the key-less-provider abstraction (`getMissingProviderEnvKey` / `providerRequiresApiKey` / optional `apiKeyEnvKey`). #154 currently leans on the `GOOGLE_CLOUD_PROJECT`-as-apiKey shortcut for `gemini-enterprise`, which is hackier than your explicit handling — I'd happily adopt yours.

**One heads-up that affects `vertex` too:** `AnthropicVertex` extends the base Anthropic SDK, which reads `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` from the environment and sends them as an `Authorization` header, clobbering the Google OAuth token (`ACCESS_TOKEN_TYPE_UNSUPPORTED`). It bites users who previously configured the `anthropic` provider and left the key in `~/.openwiki/.env`. #154 neutralizes those two vars around the (synchronous) constructor — worth guarding in #179 too if it lands standalone.

**Offer:** happy to consolidate whichever direction the maintainers prefer — either fold your key-less-provider abstraction into #154's model-family routing, or rebase #154's Gemini + MaaS routing on top of #179. I'll do the merge work either way. Deferring to the maintainers on the naming question too (`vertex` vs `gemini-enterprise`, or a single Vertex provider that routes all families).

SadriG91:
Thanks @ppsplus-bradh — this is exactly the kind of cross-PR review this repo needs right now, and your credentials heads-up was a real catch.

**Verified and fixed.** I reproduced it against a live Vertex project: a stray `ANTHROPIC_AUTH_TOKEN` fails Vertex auth exactly as you describe (the base SDK captures it and wins the `Authorization` header); interestingly a stray `ANTHROPIC_API_KEY` *alone* didn't break in my environment — but since that's SDK-version-dependent luck, a06156b now masks both around the (synchronous) `AnthropicVertex` construction and restores them after, same approach as yours. Both scenarios re-tested live: working.

**On consolidation:** I'd take you up on the second direction — land #179 first, then rebase #154's Gemini + MaaS family routing on top of it. Reasoning: this repo's CONTRIBUTING pushes hard for one-change PRs, and #179 is the smallest reviewable unit (the key-less-provider abstraction + the Claude path, which has now had multiple live end-to-end runs including a full `--init` on a production repo). Multi-family routing via `resolveVertexSurface()` is genuinely useful but a bigger design decision the maintainers should get to evaluate on its own. The abstraction is yours to adopt either way — glad it's useful.

**Naming:** happy to defer to maintainers. One consideration for them: a `vertex` provider can grow family routing behind the same name later without breaking anyone, whereas `gemini-enterprise` serving Claude/Llama/Mistral may read surprisingly in configs. But either works with either consolidation direction.

colifran:
@SadriG91 - thanks for the PR! Taking a look.

src/agent/index.ts
```diff
@@ -452,11 +464,17 @@ function createModel(
 ) {
   const retryOptions = { maxRetries: providerRetryAttempts };
 
+  if (provider === "vertex") {
+    return new ChatAnthropic(modelId, {
```
colifran, line 513: i think we need to pass `retryOptions` here

colifran (commented):
Looks great @SadriG91! I think we need to pass in `retryOptions` but good to merge after that. I have a PR open to fix the audit issue in CI.

src/agent/index.ts
SadriG91: Good catch — added, now `55d868f` after rebasing over the Bedrock/Nebius merges, so vertex honors `OPENWIKI_PROVIDER_RETRY_ATTEMPTS` like the other providers. Re-ran the full suite (typecheck, lint, 260 tests) and a live smoke test against Vertex on the `global` endpoint — all green.

SadriG91:
Thanks for the review @colifran! Two updates since:

1. **`retryOptions` — done** in the commit I replied with inline; vertex now honors `OPENWIKI_PROVIDER_RETRY_ATTEMPTS` like every other provider.
2. **Rebased onto current `main`** (now `55d868f`) — the Bedrock (#327) and Nebius (#147) providers landed since your review and touched the same provider surface, so I reconciled: Bedrock's secret-key/region onboarding steps and vertex's GCP project/location steps now coexist in the setup wizard and step chain, and Bedrock's `accessKeyId` lookup goes through the same optional-`apiKeyEnvKey` helpers this PR introduces. Full suite passes (typecheck, lint, 260 tests) plus a live end-to-end run against Vertex on the `global` endpoint.

Saw the Trivy audit fix landed too (#338), so CI should be fully green on re-run. Ready when you are.

colifran (approved):
Great job! Thanks @SadriG91!

Issue #171, opened 2026-07-07
Error 400 Unsupported chat content part type: 'file'


thuonglt:
When i start with command: openwiki "Please generate documentation for this repository"
I immediately encountered the following error.

```powershell
  ___                  __        ___ _    _
 / _ \ _ __   ___ _ __ \ \      / (_) | _(_)
| | | | '_ \ / _ \ '_ \ \ \ /\ / /| | |/ / |
| |_| | |_) |  __/ | | | \ V  V / | |   <| |
 \___/| .__/ \___|_| |_|  \_/\_/  |_|_|\_\_|
      |_|

╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ >_ OpenWiki v0.0.1 agent docs for codebases                                                                                                                                                                   │
│ provider: OpenAI-compatible                                                                                                                                                                                   │
│ model: Chat-EXACODE-A                                                                                                                                                                                         │
│ directory: D:\                                                                                                                                                                         │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

- LangSmith tracing disabled - Run failed
Tip: ask for a docs change, or use /exit when you are done.

* Error 400 Unsupported chat content part type: 'file'. Supported types: audio_embeds, audio_url, image_embeds, image_pil, image_url, input_audio, input_image, input_text, output_text, prompt_embeds, refusal,
text, thinking, tool_reference, video_url. (parameter=type, value=file)
```

acoalex:
## Summary
  OpenWiki currently appears to send repository content to the model using content blocks of type `file` during `openwiki --init` (and likely also
  during `--update` in some flows).
  This is a problem when using an `openai-compatible` provider backed by vLLM / hosted vLLM, because those endpoints may reject `file` content parts
  even if they otherwise support standard chat completions.
  As a result, OpenWiki cannot be used with some OpenAI-compatible deployments, despite the provider itself being supported.
  ## Environment
  - OpenWiki: latest npm version at the time of writing
  - Provider: `openai-compatible`
  - Backend: vLLM / hosted vLLM-compatible gateway
  - Model: custom gateway model
  - Platform: macOS locally, GitHub Actions in CI
  ## Current configuration
  ```env
  OPENWIKI_PROVIDER=openai-compatible
[...]

PR #180, opened 2026-07-07
fix: read unknown-extension UTF-8 files as text

### LINKED ISSUE #168: Text files with unrecognized extensions (terraform.tfvars, .tf, .hcl, ...) are treated as binary and unreadable

yogeshchavan2008:
## What & why

`read_file` can't read many common text files, and on the Anthropic provider it crashes the whole run.

deepagents' `getMimeType()` maps any **unrecognized extension** to `application/octet-stream`, and `isTextMimeType()` only accepts `text/*`, `application/json`, `application/javascript`, and `image/svg+xml`. Everything else is treated as binary, so `read_file` base64-encodes it into a `document`/`file` content block. Two consequences:

1. **Anthropic hard-fails the run** with `400 ... tool_result...document.source.base64.media_type: Input should be 'application/pdf'` (see #134, #171).
2. **On every provider**, the model never sees the file's real contents.

This hits ordinary source files with extensions not in deepagents' map — e.g. `.properties` (Spring config), `.scss` (Angular/Sass), `.tf` / `.tfvars` / `.hcl` (Terraform), `.lock`, and extension-less files like `mvnw` or `Dockerfile`. On a typical Spring Boot + Angular monorepo, generation dies within the first few file reads.

## The fix

A small `TextSniffingLocalShellBackend` wraps `LocalShellBackend.read()`: when the base read returns **binary** content whose bytes are **valid UTF-8 with no NUL byte** (the same cheap heuristic Git uses), it's returned as `text/plain`, re-applying the same `offset`/`limit` line-window semantics as the native text path. Genuinely binary files (images, archives, invalid UTF-8) pass through unchanged, so image handling is unaffected.

This is the content-sniffing approach suggested by the maintainer in #168, implemented as a backend wrapper in `src/agent/`. It's complementary to #164: that PR rewrites already-unsupported blocks into placeholders to avoid the 400; this one lets the model actually read these files as text in the first place.

## Changes

- `src/agent/text-sniffing-backend.ts` — new backend subclass + pure helpers (`decodeUtf8Text`, `reinterpretBinaryAsText`).
- `src/agent/index.ts` — use it for the agent's backend.
- `test/text-sniffing-backend.test.ts` — unit tests for the helpers (valid/invalid UTF-8, NUL detection, offset/limit windowing, binary pass-through) plus an integration test reading a real `.properties`/`.scss` file from a temp dir.

## Testing

`pnpm run format`, `pnpm run lint:check`, `pnpm run typecheck`, and `pnpm test` all pass locally (98 tests, 12 new).

Fixes #168. Also resolves the Anthropic 400 in #134 for text files with unknown extensions.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -5,10 +5,11 @@ import { ChatAnthropic } from "@langchain/anthropic";
 import { SqliteSaver } from "@langchain/langgraph-checkpoint-sqlite";
 import { ChatOpenAI } from "@langchain/openai";
 import { ChatOpenRouter } from "@langchain/openrouter";
-import { createDeepAgent, LocalShellBackend } from "deepagents";
+import { createDeepAgent } from "deepagents";
 import { DEBUG_ENV_KEYS, loadOpenWikiEnv, openWikiEnvDir } from "../env.js";
 import { isFileNotFoundError } from "../fs-errors.js";
 import { createSystemPrompt, createUserPrompt } from "./prompt.js";
+import { TextSniffingLocalShellBackend } from "./text-sniffing-backend.js";
 import type {
   OpenWikiCommand,
   OpenWikiRunEvent,
@@ -204,7 +205,7 @@ async function runOpenWikiAgentCore(
     model,
     tools: [],
     checkpointer,
-    backend: new LocalShellBackend({
+    backend: new TextSniffingLocalShellBackend({
       maxOutputBytes: 100_000,
       rootDir: cwd,
       timeout: 120,
--- a/src/agent/text-sniffing-backend.ts
+++ b/src/agent/text-sniffing-backend.ts
@@ -1,0 +1,86 @@
+import { LocalShellBackend } from "deepagents";
+
+/**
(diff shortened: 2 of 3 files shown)


colifran:
Hey @yogeshchavan2008 I think this is actually something we need to fix upstream in `deepagentsjs`. I'm going to work on getting this fixed upstream to resolve issues here.

colifran:
This should resolve: https://github.com/langchain-ai/deepagentsjs/pull/656

yogeshchavan2008:
> This should resolve: [langchain-ai/deepagentsjs#656](https://github.com/langchain-ai/deepagentsjs/pull/656)

Thanks @colifran , i see the fix -defaulting unknown extensions to text/plain covers the cases I hit . will close this

ghost:
@colifran Since https://github.com/langchain-ai/deepagentsjs/pull/656 has been merged, could you also make the version bump changes for deepagentsjs? Thanks!

colifran:
@Kowshik4593 yep - working on it!

colifran:

Issue #59, opened 2026-07-03
Add support for Claude Code and Codex CLI as inference providers in OpenWiki.


nvipin63:
## Add Claude Code and Codex CLI as model providers

Right now OpenWiki only talks to hosted APIs — OpenRouter, OpenAI, Anthropic, Fireworks, Baseten — which means every user needs their own API key configured before they can do anything. A lot of us already have Claude Code or Codex CLI set up and authenticated through work, and it'd be great if OpenWiki could just use those sessions instead of asking for yet another key.

**The idea:** when running `openwiki --init`, add Claude Code and Codex CLI to the provider list. If picked, OpenWiki checks that the CLI is installed and logged in, lists the models it exposes, and lets the user pick a default — same flow as the existing providers, just backed by a local CLI session instead of a raw API key.

Config would look something like:

```env
OPENWIKI_PROVIDER=claude-code
OPENWIKI_MODEL=claude-sonnet-5
```

or

```env
OPENWIKI_PROVIDER=codex
OPENWIKI_MODEL=gpt-5-codex
```

Both tools already handle auth, model listing, and enterprise licensing on their end, so the adapter work is mostly about detecting the CLI, shelling out to it (or using its SDK if one exists), and mapping responses back into whatever interface OpenWiki's existing providers implement.

Why bother: it removes a setup step for anyone who already has these tools running, makes OpenWiki easier to adopt on teams where key provisioning is a hassle, and keeps things consistent with tools people are using day-to-day anyway.

Things to figure out along the way:
- how to detect install + auth state for each CLI reliably
- what model lists actually look like for each (worth confirming exact model identifiers before locking anything in)
- whether this needs to work headless/non-interactively too, not just through `--init`
- docs for setup once it's working

Wouldn't touch any of the existing providers — this is purely additive.

jishnupramod:
@nvipin63 - You can take a look at this PR for a claude code adapter for this: https://github.com/langchain-ai/openwiki/pull/181. 
Tested e2e and working neatly

And here's the source branch of this PR from my fork: https://github.com/jishnupramod/openwiki/tree/feat/claude-code-agent-cli (in case if the PR is not getting merged here)

Yassine-Mhirsi:
#151 covers the Codex/OpenAI half of this.

[...]

PR #181, opened 2026-07-07
feat: run documentation jobs through a subscription-authenticated Claude Code CLI

jishnupramod:
Refs #59 — this implements the Claude Code half of that request. Codex (and other agent CLIs) can follow as separate, small adapter PRs behind the same interface, keeping each PR to one change per the contributing guidelines.

## What & why

OpenWiki currently requires a metered API key for every provider. Many developers already pay for a subscription coding agent, and #59 asks to use those sessions instead. This PR adds a second provider *kind*, `agent-cli`, that delegates the whole documentation job to a locally installed, subscription-authenticated agent CLI — first adapter: Claude Code (headless `claude -p`).

- **Provider model** (`src/constants.ts`): `ProviderConfig` becomes a discriminated union (`api` | `agent-cli`). New provider `claude-code` — label "Claude Code (subscription)", models `default`/`sonnet`/`opus`/`haiku` (any hand-set `OPENWIKI_MODEL_ID` passes through), binary override via `OPENWIKI_CLAUDE_CODE_BINARY`. Existing API providers are untouched.
- **Engines module** (`src/agent/engines/`): a vendor-agnostic runner (spawn, NDJSON parse, run timeout with `OPENWIKI_AGENT_CLI_TIMEOUT_SECONDS` override, stderr tail on failure, session capture for interactive follow-ups via `--resume`, process-group cleanup on exit/signals) plus a `claude-code` adapter that maps Claude Code's `stream-json` events onto OpenWiki's existing run events, so the UI renders delegated runs unchanged.
- **Shared scaffolding stays shared**: prompt assembly, git-evidence run context, update no-op detection, content snapshot, and `.last-update.json` metadata are common to both provider kinds; only the execution engine differs. The system prompt gains an `agent-cli` variant (repository-relative paths instead of the DeepAgents virtual filesystem); the `deepagents` output is byte-identical (test-guarded).
- **Onboarding** (`src/credentials.tsx`): selecting the provider runs an install check (binary + `--version`) instead of asking for an API key; the LangSmith step is skipped for delegated runs. Only `OPENWIKI_PROVIDER` and `OPENWIKI_MODEL_ID` are persisted to `~/.openwiki/.env` — no secrets; vendor auth stays in the vendor's own store.
- Headless/non-interactive works (`--print --init`, `--print --update`), answering the issue's open question about that.

**Design note vs. the issue's sketch:** this delegates the full task to the agent CLI rather than wrapping it as a LangChain chat model. A model-shim would effectively proxy subscription auth into raw API traffic (against vendor terms), and it would discard the agent's own tool loop. Delegation keeps the vendor CLI in charge of its session, auth, and tools; OpenWiki supplies the prompt, evidence, and guardrails. Model listing is static aliases for the same reason — the CLI owns the catalog.

## Security posture

- `--permission-mode acceptEdits` plus an explicit documentation-scoped allowlist: file read/search/write tools, read-only git (`Bash(git log:*)` etc.), and the single exact `rm` for the temporary plan file. No network tools.
- The delegated agent edits only the user's working tree — the same trust model as the existing `LocalShellBackend`, which already executes model-driven shell on the host.
- No OAuth-token extraction or proxying, ever; documented as a hard boundary for future adapters.

## Caveats (documented in README)

- Local runs only for now — the scheduled GitHub Action still needs an API-key provider.
- macOS/Linux only (POSIX process-group management); LangSmith tracing does not apply to delegated runs.

## How I tested it

- `pnpm run format`, `pnpm run lint`, `pnpm test` all clean on the branch: 137 tests across 15 files. New tests follow the repo's existing patterns and exercise real behavior — stub executables emit recorded `stream-json` so the runner is tested end-to-end (event ordering, tool_end name patching, session capture, stdin EPIPE guard, timeout kill, orphaned-process cleanup on parent exit), plus unit coverage for the adapter (`buildArgs`/`parseEvent` against captured fixtures), provider-kind narrowing, onboarding step sequencing, and a non-interactive startup regression test proving no API key is required.
- Live verification against installed Claude Code v2.1.201: flag set and `stream-json` schema validated field-for-field; a real `--init` on a fixture repository with all API keys stripped produced `openwiki/` docs, the `AGENTS.md` reference section, and `.last-update.json`; interactive follow-ups resumed the same vendor session; `--update` verified on a repository with real drift.


jishnupramod:
Rebased onto current main (0.0.4). Notable reconciliations: the provider-kind config now feeds the derived env-key arrays from the diagnostics refactor, and the non-interactive startup gate in `src/startup.ts` lets agent-CLI providers through without an API key (they authenticate via the vendor CLI's own login) — covered by a new case in `test/startup.test.ts`. Full suite passes (149/149) and a live `--init` smoke against a fixture repo succeeded on the rebased build.

colifran:

Issue #103, opened 2026-07-05
Feature request: add support for Azure OpenAI as an inference provider in OpenWiki.


pavelgk:
Many enterprise users host or subscribe to Azure OpenAI (private deployments, enterprise quotas, regional endpoints, and Azure AD governance). Supporting Azure OpenAI would:
Allow teams to use existing Azure subscriptions and deployments without provisioning separate OpenAI/OpenRouter API keys.
Reduce onboarding friction for organizations with Azure governance, private endpoints or VNet-restricted services.
Improve performance and compliance by using regional endpoints and corporate networking controls.
Provide a path to use Azure AD / managed identity authentication where required.

OpenWiki would then:
Detect whether Azure OpenAI credentials are available via environment variables, az CLI login, or managed identity.
Verify connectivity to the configured Azure OpenAI endpoint and authentication.
Enumerate available deployments (models) for the Azure OpenAI resource.
Allow the user to choose a default deployment for completions/chat/embeddings.
Persist the provider configuration in ~/.openwiki/.env (or existing config path).

pavelgk:
Unfortunately that didn't work for me.
The 401 Error occurs. Incorrect API key provided...
I've tested the key in a python script, it worked fine.

Monday, 06 July 2026, 06:57am +02:00 from PolyKelvin ***@***.*** :

PR #182, opened 2026-07-07
feat: add Azure OpenAI provider

seb-jul:
## What

Adds an `azure` provider that documents a repo through an **Azure OpenAI** resource, with two authentication modes:

- **Microsoft Entra ID bearer token** (default when no key is set) — uses `@azure/identity`'s `DefaultAzureCredential` + `getBearerTokenProvider`, wired through `AzureChatOpenAI`'s `azureADTokenProvider` hook so the token refreshes on every request. Works with an `az login` session, environment credentials, or managed/workload identity (CI).
- **API key** — `AZURE_OPENAI_API_KEY`. When a key is present it is used by default; `AZURE_OPENAI_USE_AD_TOKEN=true` forces the token path even when a key is configured.

```bash
OPENWIKI_PROVIDER=azure
AZURE_OPENAI_ENDPOINT=https://your-resource.openai.azure.com/
OPENWIKI_MODEL_ID=your-deployment-name
# AZURE_OPENAI_API_VERSION=2024-12-01-preview   # optional
# AZURE_OPENAI_API_KEY=...                       # optional; omit to use Entra ID
```

## Why

Organizations on Azure consume OpenAI models through an Azure OpenAI resource (billing, quotas, RBAC, data residency, private endpoints) rather than a direct OpenAI API key — and many disable key auth entirely, requiring Entra ID. OpenWiki couldn't target those deployments. No existing provider covers Azure OpenAI (`openai-compatible` fronts an OpenAI-shaped base URL but has no Azure deployment routing and no Entra ID auth path).

## How

- Reuses `AzureChatOpenAI` from `@langchain/openai` (already a dependency). Azure routes by **deployment name**, which is taken from `OPENWIKI_MODEL_ID` — mirroring how the `openai-compatible` provider treats the model ID as "the name the gateway exposes" — so no new key is introduced for the deployment.
- `ProviderConfig.apiKeyEnvKey` becomes **optional** and a new `endpointEnvKey` is added. Credential gating goes through new `getMissingProviderEnvKey()` / `providerRequiresApiKey()` / `providerSupportsAdToken()` helpers so key-less providers are handled uniformly across the agent runtime, both CLI non-interactive gates, and `needsCredentialSetup()`. A credential hint (`az login` / `DefaultAzureCredential`) is appended to missing-credential errors.
- Interactive onboarding adds an **endpoint** step (validated like the base-URL step) and treats the API-key step as **optional** for azure (Entra ID is the default), while still letting users save `AZURE_OPENAI_API_KEY` for key-based auth.
- `azureUsesAdToken()` resolves the auth mode: the `AZURE_OPENAI_USE_AD_TOKEN` flag wins, otherwise fall back to token auth when no key is set.
- `AZURE_OPENAI_ENDPOINT` / `AZURE_OPENAI_API_VERSION` (default `2024-12-01-preview`) / `AZURE_OPENAI_USE_AD_TOKEN` / `AZURE_OPENAI_API_KEY` join `MANAGED_ENV_KEYS`; the first three are shown as non-secret values in credential diagnostics, the key stays masked.
- Adds `@azure/identity` as a dependency (the standard Azure auth SDK, needed for the Entra ID token provider).

## How I tested it

- `pnpm run format`, `pnpm run lint`, and `pnpm test` all pass (93 tests, including new coverage in `test/constants.test.ts` for the azure config, the `getMissingProviderEnvKey`/`providerRequiresApiKey`/`providerSupportsAdToken`/`azureUsesAdToken` helpers, and the Entra ID credential hint).
- Wire-level: constructed `AzureChatOpenAI` exactly as `createAzureModel` does with `DefaultAzureCredential` + `getBearerTokenProvider` against a live Azure OpenAI deployment — a bearer-token request succeeded (no API key present).
- CLI non-interactive: `openwiki -p "..."` with `OPENWIKI_PROVIDER=azure`, `AZURE_OPENAI_USE_AD_TOKEN=true`, `AZURE_OPENAI_ENDPOINT`, and `OPENWIKI_MODEL_ID` set to the deployment name — responded through the token path.
- End-to-end: `openwiki --init` against a real repository generated a full `openwiki/` docset (quickstart + section pages + metadata) entirely through the Azure Entra ID bearer-token path.

## Notes

- Scope is a single concern: the Azure OpenAI provider. No other provider's behavior changes.
- The repo's own `openwiki/` docs are intentionally left untouched — they're regenerated by the scheduled `openwiki-update.yml` workflow.


seb-jul:
Closing in favor of #187.

After more thought, a dedicated `azure` provider duplicates the existing `openai-compatible` provider — Azure OpenAI already exposes an OpenAI-compatible `/openai/v1` surface that `openai-compatible` reaches today. The only thing it can't do is authenticate with a dynamic Microsoft Entra ID token instead of a static key.

So #187 takes the smaller, provider-free approach: it adds an `OPENAI_COMPATIBLE_AUTH=entra-id` mode to the existing provider (token via `DefaultAzureCredential`), rather than a new per-cloud provider. Same capability, much smaller surface, and it still answers the Entra ID / managed-identity ask in #103.

PR #197, opened 2026-07-07
fix: validate configured provider before setup completes

### LINKED ISSUE #176: OPENWIKI_PROVIDER incorrectly falls back to "openrouter"
`needsCredentialSetup` in `src/credentials.tsx` only checks whether `OPENWIKI_PROVIDER` is set and not if it contains a valid provider value based on what exists in `src/constants.tsx`

Current logic:

```ts
if (process.env[OPENWIKI_PROVIDER_ENV_KEY] === undefined) {
  return "provider";
}
```

It seems as soon as `OPENWIKI_PROVIDER` is set to any value, it moves on and provider picker isn't shown again. But `resolveConfiguredProvider()` in `src/constants.ts` defaults `"openrouter"` when the configured value fails to normalize the provider.

* the setup flow doesn't list all supported providers, like `"openai-compatible"`
* the setup flow seems to consider the provider configured
* the value of provider is substituted with the default provider
* no real validation error

The UI ends up showing:

```txt
[DONE] Provider OpenRouter
```

even though the provider value something different (in this case "openai-compatible"), and user is requested to provide `OPENROUTER_API_KEY`, instead of, for example, `OPENAI_COMPATIBLE_API_KEY`

I managed to bypass this by editing or removing `OPENWIKI_PROVIDER` from `~/.openwiki/.env`, and using the appropriate env vars that match `OPENAI_COMPATIBLE_*`

### Steps to reproduce

1. Run `openwiki --init`
2. Select any provider and complete setup
3. Edit `~/.openwiki/.env`
4. Set `OPENWIKI_PROVIDER` to an invalid value
   * example: typoed provider name
   * or provider not present in `PROVIDER_CONFIGS`
5. Run `openwiki --update` (or `--init`)

### Behaviour

* Provider step shows as completed
* UI displays the default provider (`OpenRouter`)
* Wizard skips provider selection
* Wizard asks for `OPENROUTER_API_KEY`
* No validation error is shown

### Expectations

- The setup flow should not silently substitute a different provider
 Invalid or unrecognized `OPENWIKI_PROVIDER` values should retrigger provider selection or show an explanation / error.
- The setup flow should not silently substitute a different provider
- The setup flow should list and validate all supprted providers from `src/constants.ts`.

ousamabenyounes:
## What / Why / How

Fixes #176.

OpenWiki now treats an invalid `OPENWIKI_PROVIDER` as incomplete setup instead of accepting the variable because it exists and then silently falling back to OpenRouter. The setup UI also keeps the Provider step active for invalid values, so users can choose a valid provider before the API-key step.

I also updated the operations docs to state that invalid provider values trigger the interactive flow.

RED, with the production fix in `src/credentials.tsx` temporarily reverted while keeping the new regression test:

```text
pnpm exec vitest run test/credentials.test.ts

❯ test/credentials.test.ts (1 test | 1 failed)
  × requires provider setup for an invalid configured provider

Test Files  1 failed (1)
Tests  1 failed (1)
```

GREEN, with the fix applied:

```text
pnpm exec vitest run test/credentials.test.ts test/constants.test.ts test/env.test.ts test/commands.test.ts

✓ test/constants.test.ts (20 tests)
✓ test/commands.test.ts (21 tests)
✓ test/env.test.ts (9 tests)
✓ test/credentials.test.ts (1 test)

Test Files  4 passed (4)
Tests  51 passed (51)
```

Full local validation:

```text
pnpm test
Test Files  8 passed (8)
Tests  87 passed (87)

pnpm run typecheck
pnpm run format:check
pnpm run lint:check
pnpm run build
```

## Affected platforms

- [x] All platforms

## Test plan

- Added `test/credentials.test.ts` coverage for invalid `OPENWIKI_PROVIDER` values.
- Verified full test, typecheck, formatting, lint, and build commands locally.

## Checklist

- [x] Tests added/updated (TDD: red -> green)
- [x] `pnpm test` passes
- [x] `pnpm run typecheck` passes
- [x] Docs updated if needed (README, platform-support.md)
- [x] Targets `main` branch

### DIFF
--- a/openwiki/operations/credentials-and-updates.md
+++ b/openwiki/operations/credentials-and-updates.md
@@ -43,7 +43,7 @@ The setup flow runs for **all** interactive commands (chat, init, and update) wh
 2. Otherwise, if `OPENROUTER_API_KEY` is present, default to `openrouter`.
 3. Otherwise, fall back to `DEFAULT_PROVIDER` (`openrouter`).
 
-`needsCredentialSetup()` in `src/credentials.tsx` checks whether the provider env var, the provider's API key, a model ID (unless overridden), and a LangSmith key are all present. Any missing value triggers the interactive flow.
+`needsCredentialSetup()` in `src/credentials.tsx` checks whether the provider env var is valid and whether the provider's API key, a model ID (unless overridden), and a LangSmith key are all present. Any missing or invalid provider value triggers the interactive flow.
 
 ## Model and credential diagnostics
 
--- a/src/credentials.tsx
+++ b/src/credentials.tsx
@@ -9,6 +9,7 @@ import {
   getProviderModelOptions,
   isValidBaseUrl,
   isValidModelId,
+  normalizeProvider,
   normalizeModelId,
   OPENWIKI_MODEL_ID_ENV_KEY,
   OPENWIKI_PROVIDER_ENV_KEY,
@@ -44,7 +45,7 @@ export function needsCredentialSetup(
   const apiKeyEnvKey = getProviderApiKeyEnvKey(provider);
 
   return (
-    process.env[OPENWIKI_PROVIDER_ENV_KEY] === undefined ||
+    !hasValidConfiguredProvider() ||
     !process.env[apiKeyEnvKey] ||
     needsBaseUrlStep(provider) ||
     (modelIdOverride === null &&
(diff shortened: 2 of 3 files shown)


bracesproul (approved):
awesome! thank you for correcting this

PR #200, opened 2026-07-07
feat: implement langsmith as a connector

colifran:
### Summary

This PR introduces langsmith as a connector to allow openwiki to ingest data from langsmith projects.

### Testing

Performed e2e testing verifying that the connector worked correctly:

image

image


src/connectors/sources/langsmith/index.ts
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * Default LangSmith API host root; EU workspaces override via
+ * config.apiBaseUrl (e.g. https://eu.api.smith.langchain.com).
+ */
+const DEFAULT_API_BASE_URL = "https://api.smith.langchain.com";
+
+/**
+ * Env var holding the LangSmith API key. Referenced by name, never read into output.
+ */
+const LANGSMITH_API_KEY_ENV = "OPENWIKI_LANGSMITH_API_KEY";
```
bracesproul, line 29: we alr store langsmith api keys under `LANGSMITH_API_KEY`, we should reuse that

there isnt already a constant for it tho, so if you could add it that would be great

src/credentials.tsx
```diff
@@ -1923,6 +1987,25 @@ function Prompt({
     );
   }
 
+  if (step === "source-projects") {
+    return (
+      <Box flexDirection="column">
+        <Text>Which LangSmith project should OpenWiki monitor?</Text>
+        <Text color="gray">
+          Enter one project (tracing session) name. Add another LangSmith source
```
bracesproul, line 1995: can we support name or project id? use the uuid util to check if it is a uuid

src/credentials.tsx
```diff
@@ -1923,6 +1987,25 @@ function Prompt({
...
+          to monitor more projects.
+        </Text>
+        <BorderedInput
+          maxDisplayWidth={inputDisplayWidth}
+          marginTop={1}
+          prefix="project="
```
bracesproul, line 2001:
```suggestion
          prefix="project name or ID="
```

bracesproul (dismissed):
lgtm, couple nits

PR #202, opened 2026-07-07
feat: use openai responses api

bracesproul:


acgxv:
Need a new release! Really need this!

PR #211, opened 2026-07-08
fix: skip clean update prints before credentials

100yenadmin:
## Summary

- Extract startup command resolution into a small testable helper.
- Allow `openwiki --update --print` to skip provider credential validation when the existing clean-update no-op detector says no model call is needed.
- Keep credential validation unchanged when source files changed or an update message is supplied.

## Why

Scheduled docs jobs should be able to run a no-change update without requiring model credentials. OpenWiki already detects clean update no-ops inside the agent path, but the `--print` startup credential check currently blocks before that detector can run.

This keeps the change narrow to the `--print` startup path so it complements, rather than overlaps with, #153's broader non-TTY routing work.

## Tests

- `pnpm exec vitest run test/startup.test.ts test/update-noop.test.ts`
- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm run lint:check`
- `pnpm run build`
- `pnpm test`

Related: #153

### DIFF
--- a/src/cli.tsx
+++ b/src/cli.tsx
@@ -10,6 +10,7 @@ import {
   type CliCommand,
   type HelpRow,
 } from "./commands.js";
+import { resolveStartupCommand } from "./startup.js";
 import {
   InitSetup,
   needsCredentialSetup,
@@ -2966,7 +2967,10 @@ if (parsedCommand.kind === "run" && !parsedCommand.dryRun) {
   await loadOpenWikiEnv();
 }
 
-const command = resolveStartupCommand(parsedCommand);
+const command = await resolveStartupCommand(parsedCommand, {
+  cwd: process.cwd(),
+  isStdinTTY: Boolean(process.stdin.isTTY),
+});
 
 if (shouldPrintStartupError(argv, parsedCommand, command)) {
   process.stderr.write(`${command.message}\n`);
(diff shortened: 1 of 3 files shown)


colifran:
@100yenadmin thanks for this! taking a look

colifran:
@100yenadmin #153 was merged - could you rebase before I review? changes look good but want to make sure things are aligned with the changes that went in with #153

100yenadmin:
Rebased this on current `main@bd3d214b643af4b73968890fbbd0fa88f099b757` after #153. The PR is now a single commit, `a47000f526c51d69c4e9271c9efeb5443ce31fb9`.

I resolved the startup overlap by keeping #153's non-TTY routing / interactive-chat error behavior and moving the credential/no-op startup validation into `src/startup.ts`.

Validation from the rebased branch:

- `pnpm exec vitest run test/startup.test.ts test/commands.test.ts` -> 2 files, 31 tests passed
- `pnpm run typecheck` -> passed
- `pnpm run format:check` -> passed
- `pnpm run lint:check` -> passed
- `pnpm run build` -> passed
- `pnpm test` -> 9 files, 97 tests passed

Non-blocking local note: pnpm still warns about an unresolved `${NPM_TOKEN}` in my user `.npmrc`, but install/checks completed without using that token.

colifran (approved):
Looks great. Thanks @100yenadmin!

PR #219, opened 2026-07-08
fix: read text config files as text

Wiar8:
Fixes OpenWiki agent failures when DeepAgents treats common textual repo files such as Dockerfile, .env.example, .prettierrc, .trivyignore, and .snap files as binary application/octet-stream content.

  This adds an OpenWiki backend wrapper that only reclassifies unknown octet-stream files as text when their bytes decode cleanly as UTF-8 and contain no NUL bytes, while keeping real binary files binary.

  Verified with:
  - pnpm test
  - pnpm run typecheck
  - pnpm run lint:check
  - pnpm run format:check
  - pnpm run build


colifran:
Hi @Wiar8 - we fixed this upstream in deepagentsjs. The fix is available as part of the v0.0.3 release of OpenWiki

PR #122, opened 2026-07-06
feat: implement tool schema recovery middleware to handle malformed file tool calls without aborting the run

zan22ye:
## Summary

This PR makes OpenWiki more resilient when the model emits malformed filesystem tool calls.

Issue #119 reports a failure like:

- `Error invoking tool 'write_file' with kwargs {}`
- `Received tool input did not match expected schema`
- `Invalid input: expected string, received undefined`
- `at file_path`

In the current behavior, a malformed `write_file` call can abort the entire run. This change makes those schema-shape failures recoverable for core file tools by converting them into an error `ToolMessage`, so the model can retry with valid arguments instead of terminating the whole agent loop.

Closes #119

## What changed

- Added a small `wrapToolCall` middleware in `src/agent/index.ts`
- The middleware only handles schema-validation-style failures for:
  - `write_file`
  - `edit_file`
  - `read_file`
- When one of those tool calls fails because its input does not match the expected schema, OpenWiki now returns an error `ToolMessage` back to the model instead of immediately rethrowing
- The recovery message tells the model:
  - which tool failed
  - which argument names are accepted
  - to use repository-rooted virtual paths such as `/openwiki/quickstart.md`

## What this does not change

This PR does not swallow normal runtime/tool execution errors such as:

- missing files
- `old_string` not matching in `edit_file`
- permission/path errors
- other non-schema tool failures

Those still surface normally.

## Why this approach

The failure in #119 appears to come from malformed tool-call arguments generated by the model, rather than from a bug in the filesystem backend itself.

Because of that, the safest fix at the OpenWiki layer is to treat these schema-shape failures as recoverable and let the model retry, instead of trying to guess or silently rewrite arguments.

This keeps the change:

- small
- local to OpenWiki
- provider/model agnostic
- low risk for unrelated tool behavior

## Tests

Added regression coverage in `tests/tool-schema-recovery.test.ts` for:

- recognizing recoverable `write_file` schema errors
- ignoring normal execution errors
- converting a recoverable schema failure into an error `ToolMessage`

## Verification

Ran:

- `./node_modules/.bin/vitest run`
- `./node_modules/.bin/tsc -p tsconfig.json`

Issue #119, opened 2026-07-06
Error invoking tool 'write_file' with kwargs {} with error: Error: Received tool input did not match expected schema


secountAiAccount:
>_ OpenWiki v0.0.1 agent docs for codebases                                                                                                                                                                                                                                                                         │
│ provider: Anthropic                                                                                                                                                                                                                                                                                                 │
│ model: mimo-v2.5                                                                                                                                                                                                                                                                                                    │
│ directory: xxxx                                                                                                                                                                                                                                                                               │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

- LangSmith tracing disabled - Run failed
Tip: ask for a docs change, or use /exit when you are done.

* Error Error invoking tool 'write_file' with kwargs {} with error: Error: Received tool input did not match expected schema

✖ Invalid input: expected string, received undefined
  → at file_path
    at DynamicStructuredTool.call (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/@langchain/core/dist/tools/index.js:113:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async baseHandler (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/langchain/dist/agents/nodes/ToolNode.js:211:20)
    at async ToolNode.wrappedHandler [as wrapToolCall] (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/langchain/dist/agents/utils.js:363:20)
    at async ToolNode.runTool (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/langchain/dist/agents/nodes/ToolNode.js:257:11)
    at async ToolNode.run (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/langchain/dist/agents/nodes/ToolNode.js:292:15)
    at async ToolNode.invoke (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/langchain/dist/agents/RunnableCallable.js:34:23)
    at async RunnableSequence.invoke (file:///D:/nvm/v22.22.0/node_modules/openwiki/node_modules/@langchain/core/dist/runnables/base.js:934:69)
[...]

PR #220, opened 2026-07-08
chore: bump deepagents version

colifran:
### Summary

This PR bumps the deepagents version to pick up this change: https://github.com/langchain-ai/deepagentsjs/pull/656. This should resolve issues where unknown file extensions are defaulting to binary causing Anthropic 400s.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@langchain/openai": "^1.5.3",
     "@langchain/openrouter": "^0.4.3",
     "@langchain/protocol": "^0.0.18",
-    "deepagents": "^1.10.5",
+    "deepagents": "^1.10.6",
     "ink": "^5.1.0",
     "langchain": "^1.5.1",
     "marked": "^18.0.5",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,8 +27,8 @@ importers:
         specifier: ^0.0.18
         version: 0.0.18
       deepagents:
-        specifier: ^1.10.5
-        version: 1.10.5(langsmith@0.7.10(openai@6.44.0(ws@8.21.0)(zod@4.4.3))(ws@8.21.0))(openai@6.44.0(ws@8.21.0)(zod@4.4.3))(react@18.3.1)(ws@8.21.0)
+        specifier: ^1.10.6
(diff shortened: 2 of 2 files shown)


ghost:
@colifran Should we also bump the OpenWiki version to get these changes reflected?

colifran:
@Kowshik4593 ya - this one seems to be causing people a lot of pain. We'll prioritize getting a v0.0.3 out for OpenWiki.

ghost:
True, waiting for a very long time! Thanks for the quick response and updates.
@colifran

bracesproul (approved):

colifran:
@Kowshik4593 v0.0.3 has been released: https://github.com/langchain-ai/openwiki/pull/221

ghost:
That`s Great. Thanks!
@colifran

ghost:
@colifran There are a few issues related to filesystem tool parameter mismatches. Could you take a look at 
https://github.com/langchain-ai/deepagentsjs/pull/659 ?

When using weaker or custom LLMs (for example, smaller open source models) in deepagents and downstream consumers like OpenWiki, the agent crashes on the first filesystem tool call with a schema validation error.

The root causes are:

* ls defines its directory argument as path, while read_file, write_file, and edit_file expect file_path. Models often default to using path across all filesystem tools.
* The prompt examples in fs.ts and skills.ts use read_file(path, ...), which unintentionally teaches models to use the wrong parameter name.

This PR aligns the parameter usage and updates the examples to improve compatibility with weaker and custom models.

colifran:
@Kowshik4593 your PR looks great - i approved it! i'll do the leg work to get that released in deepagentsjs and get the version bumped over here. i was looking into this which is related: https://github.com/langchain-ai/openwiki/issues/119. i wonder if we should still have a middleware to help with recovery. failing and aborting due to a failed tool call because of a malformed schema feels like a frustrating experience. i suspect most instanced of that will be resolved with your change in deepagentsjs but maybe we should still add something in openwiki as a safe guard. what do you think?

ghost:
@colifran True, I agree that failed tool calls should be handled within OpenWiki itself. I believe you’ve already made those changes, as I can see this https://github.com/langchain-ai/openwiki/pull/122

PR #230, opened 2026-07-09
fix: constrain init update writes to openwiki

100yenadmin:
## Summary

This keeps init/update runs scoped to the generated OpenWiki tree by wrapping the DeepAgents local shell backend and refusing write/edit calls outside /openwiki/ when the command is init or update.

Chat mode remains unrestricted so existing interactive editing behavior is unchanged.

## Validation

- pnpm exec vitest run test/docs-only-backend.test.ts test/startup.test.ts test/update-noop.test.ts test/commands.test.ts: 40 tests passed
- pnpm run typecheck: passed
- pnpm run format:check: passed
- pnpm run lint:check: passed
- pnpm run build: passed
- pnpm test: 100 tests passed

Note: pnpm prints local machine warnings about an NPM_TOKEN placeholder in ~/.npmrc and the package.json pnpm field. They do not affect this PR.


colifran:
Thanks for the pr @100yenadmin! Out of curiosity, have you seen this behavior take place, i.e., where the agent goes outside of the OpenWiki tree during init/update?

100yenadmin:
I have not seen arbitrary source/runtime edits in the retests. The concrete behavior I saw was the repo-level instruction path and write-scope issue: repository docs init/update could create or update top-level agent instruction files unless suppressed, and repository-mode docs safety depended mostly on prompt instructions plus a post-run diff gate. For scheduled docs runs, I want the write tool itself to fail closed if an init/update tries to write outside generated wiki docs.

I rebased this PR on current `main` at `27e3c09056ba9a4a4360b8ea2866e9358501e57a` and adjusted it for the newer personal/code split:
- personal/local-wiki mode can still write normal wiki pages at the `~/.openwiki/wiki` virtual root, such as `/quickstart.md`;
- repository/code mode init/update is confined to `/openwiki/**`;
- chat mode remains unrestricted for explicit user-requested edits.

Validation on the rebased branch:
- `pnpm exec vitest run test/docs-only-backend.test.ts` -> 4 passed
- `pnpm run typecheck` -> passed after refreshing the current-main dependency install
- `pnpm run format:check` -> passed
- `pnpm run lint:check` -> passed
- `pnpm run build` -> passed

So the PR is preventive hardening for the repository docs path, with the observed motivation being top-level docs/instruction noise and safe scheduled operation rather than a confirmed case of the model rewriting arbitrary source files.

colifran:
@100yenadmin ya smart. makes a lot of sense. i'll take a deeper look at the pr.

colifran (commented):
One thing I want to look into - we have permissions that you can configure on your deepagent:
```ts
const agent = createDeepAgent({
    model,
    tools: createOpenWikiConnectorTools(),
    checkpointer,
    permissions: [{
      mode: "deny",
      paths: [],
      operations: ["write"],
    }],
    backend: new LocalShellBackend({
      maxOutputBytes: 100_000,
      rootDir: cwd,
      timeout: 120,
      virtualMode: true,
    }),
    systemPrompt: createSystemPrompt(command, outputMode),
  });
```

Do you think we would be able to leverage these instead here: https://github.com/langchain-ai/openwiki/blob/8636659273c91f6171d93408226d8e60e894408b/src/agent/index.ts#L149?

100yenadmin:
I checked the current `deepagents` permission types against this branch. I think we can leverage them, but I do not think they fully replace this PR's backend guard as-is.

What I found:
- `permissions` rules are first-match filesystem rules, so the repo-docs shape would need an allow-then-deny rule set, not `paths: []`, e.g. allow write to `/openwiki/**` and then deny write to `/**`.
- The current type docs say permissions apply to `ls`, `read_file`, `write_file`, `edit_file`, `glob`, and `grep`. They explicitly do not govern `execute`, because shell commands can write anywhere under the backend root.
- OpenWiki still gives the agent a `LocalShellBackend` with `execute`, so permissions alone would protect the filesystem write/edit tools but would not be a complete scheduled-docs write boundary if the agent used shell commands.

So my read is: yes, permissions are the right high-level API to use where possible, and I am happy to revise this PR to add those rules at `createDeepAgent`. But I would keep the backend write/edit guard (or add an equivalent execute/no-execute backend route for repository docs) unless DeepAgent has a permission mode that also constrains `execute`. That keeps the safety property local to OpenWiki instead of depending only on prompt discipline and post-run diff checks.

If you prefer, I can update this PR to layer permissions + the existing backend guard, or split a follow-up that explores a cleaner CompositeBackend/route-prefix setup.

colifran (approved):
Thanks @100yenadmin! This makes sense to me with the caveat that we probably should think about what this means for execute since it can technically work outside of these guards. I think we will need a openwiki backend implementation like this to support a `.openwikiignore` so thank you for getting this started.

100yenadmin:
Anytime @colifran we forked this since it's an early project but it works semi-okay 7/10 right now with https://www.neondiff.com https://github.com/electricsheephq/evaos-code-review-bot-neondiff projects of mine. 

I will keep supporting it, it just needs a lot of patches for security and usage for business use (it's new project so I get it).

I am an AI biz owner and maintainer btw powered by MOE and multiple frontier models, so if you do have questions just ping me @100yenadmin 

I look forward to contributing on this project.

Issue #233, opened 2026-07-09
Add IBM Bob (Bob Shell) as a subscription-based agent CLI provider


jishnupramod:
#59 asks for documentation runs through subscription coding agents instead of raw API keys, and #181 implements the `agent-cli` provider infrastructure with a first adapter. IBM Bob is a natural second adapter behind the same interface: it is publicly available (https://bob.ibm.com), billed through an IBMid subscription, and its CLI — Bob Shell (https://bob.ibm.com/docs/shell) — supports the same delegation shape:

- one-shot headless runs with the prompt piped on stdin,
- `--output-format stream-json` NDJSON events (`init` / `message` / `tool_use` / `tool_result` / `error` / `result`),
- `--resume <session-uuid>` to continue a session across runs,
- `--approval-mode auto_edit` for headless file edits confined to the working directory, plus `--allowed-tools` for anything beyond edits.

Differences from the first adapter that the engine interface has to absorb (verified against Bob Shell v1.0.3 with live runs):

- Bob has no system-prompt flag, so the system prompt must be prepended to the stdin payload — a small optional `buildStdin` hook on the adapter interface covers this.
- Resuming a session rejects stdin-only input; the message has to travel via `-p`.
- Tool names are Roo-style (`execute_command`, `write_to_file`, ...) and `--allowed-tools` matches bare tool names only, without command-level scoping.

I have this working end-to-end on a fork (adapter, tests, docs) and will raise a PR that stacks on #181.

PR #234, opened 2026-07-09
feat: add IBM Bob (Bob Shell) as a subscription agent CLI provider

jishnupramod:
Refs #233 (and the broader ask in #59). **Stacks on #181** — the first 15 commits here are that PR; the 10 commits on top are this change. Happy to rebase once #181 lands so only the delta remains.

## What this adds

`ibm-bob` as a second `agent-cli` provider behind the adapter interface #181 introduces. Bob Shell is IBM's coding-agent CLI (https://bob.ibm.com/docs/shell), billed through an IBMid subscription. Runs delegate to `bob` headless with `--output-format stream-json`, `--approval-mode auto_edit`, and a pinned `--chat-mode advanced` so behavior does not depend on the user's saved chat mode.

- `src/agent/engines/ibm-bob.ts` — argument/stdin composition and NDJSON event parsing for Bob's `init` / `message` / `tool_use` / `tool_result` / `error` / `result` events, with session-id capture for `--resume`.
- A small optional `buildStdin(spec)` hook on the adapter interface: Bob has no system-prompt flag, so the adapter prepends the system prompt to the stdin payload; the runner writes `adapter.buildStdin?.(spec) ?? spec.prompt` (the `??` is load-bearing — an empty string is a meaningful payload on resume).
- Resume quirk, pinned by a runner test: Bob rejects stdin-only input when resuming, so on resume the composed payload travels via `-p` with empty stdin.
- Provider entry (`OPENWIKI_PROVIDER=ibm-bob`), `OPENWIKI_IBM_BOB_BINARY` binary override, install hint covering login and folder trust, README and wiki documentation.

## Security posture

`auto_edit` auto-approves only Bob's file-edit tools, which Bob confines to the working directory (the runner spawns the CLI with cwd set to the target repository). The shell tool `execute_command` is allowlisted by bare name — Bob's `--allowed-tools` has no command-level scoping, and its shell subprocesses are not cwd-confined. That matches the trust the API path already extends to `LocalShellBackend`, which likewise runs unsandboxed shell commands in the repository. The adapter's comments document this posture explicitly. Network-capable tools stay off the allowlist.

## Verification

- `pnpm run format`, `pnpm run lint`, `pnpm test` (177 passed), `pnpm run build`
- Live against Bob Shell v1.0.3: a stream-json schema probe checked field-for-field against the parser fixtures, one-shot stdin invocation, session resume, and an end-to-end `--init` on a fixture repository producing the wiki and `.last-update.json`, with the temporary plan file cleaned up through a real `execute_command` call.


colifran:

Issue #237, opened 2026-07-09
Add configurable retry/backoff for transient provider errors


yvette-carlisle:
## Summary

OpenWiki agent runs can involve many model calls. With an OpenAI-compatible provider, a single transient provider/proxy failure currently appears to abort the whole run. It would be helpful to have configurable retry/backoff around model calls for retryable errors.

## Motivation

When running OpenWiki against an OpenAI-compatible gateway such as LiteLLM, longer documentation/update runs can make many consecutive LLM requests. Even if most calls succeed, a temporary `502`, timeout, connection reset, or rate limit response can fail the entire OpenWiki run and leave partial generated state in `openwiki/`.

This is especially painful because the failure is often not related to the repository content or prompt. It is usually a transient provider/gateway condition, and retrying the same request after a short delay may succeed.

## Current behavior

I could not find a provider-level retry/backoff option in OpenWiki for the `openai-compatible` path. Some SDK/provider stacks may have their own behavior, but from the CLI/user perspective there does not seem to be an OpenWiki-level way to configure retry policy for transient provider failures during a run.

## Proposed behavior

Add an OpenWiki-level retry policy for model calls, ideally applying to all providers or at least OpenAI-compatible providers. For example:

```sh
OPENWIKI_PROVIDER_RETRY_ATTEMPTS=4
OPENWIKI_PROVIDER_RETRY_INITIAL_DELAY_MS=500
OPENWIKI_PROVIDER_RETRY_MAX_DELAY_MS=8000
```

Retryable cases could include:

- HTTP `429`
- HTTP `500`, `502`, `503`, `504`
- network timeouts
- connection resets before a streamed response has been committed

It would also be good to respect `Retry-After` when present.

## Non-goals

This should not hide deterministic bad requests such as invalid credentials, unsupported models, invalid tool schemas, or prompt/content validation errors. Those should still fail fast.

For streaming responses, retry should probably only happen when no response has been emitted yet. If a stream fails after tokens/events have already been emitted, failing the run may be safer than silently replaying.

## Why this helps

OpenWiki is doing agentic codebase documentation, not a single short chat call. A small amount of retry/backoff makes the CLI more reliable with self-hosted or proxied providers without requiring each user to wrap the OpenAI-compatible endpoint with their own retry proxy.

AbhishekRP2002:
[...]

PR #239, opened 2026-07-09
fix: add configurable provider retry attempts

AbhishekRP2002:
## What

Adds configurable provider retry attempts via `OPENWIKI_PROVIDER_RETRY_ATTEMPTS`.

- Defaults provider retries to `3`
- Allows users to override retries through `process.env` or `~/.openwiki/.env`
- Validates explicit values as positive integers
- Passes the resolved value to LangChain model clients as `maxRetries`
- Applies to OpenRouter, Anthropic, OpenAI, and OpenAI-compatible providers
- Adds diagnostics/debug visibility and docs

## Usage

Configure retry attempts by setting `OPENWIKI_PROVIDER_RETRY_ATTEMPTS` in your shell, CI environment, or `~/.openwiki/.env`:

```sh
OPENWIKI_PROVIDER_RETRY_ATTEMPTS=3 openwiki --update
```

Or persist it locally:

```env
OPENWIKI_PROVIDER_RETRY_ATTEMPTS=3
```

The value must be a positive integer. If unset, OpenWiki defaults to `3` retries.

## Why

Transient provider failures like rate limits, 5xx errors, and temporary gateway failures can interrupt OpenWiki runs. This gives users a supported way to tune retry behavior without changing code.

LangChain already owns the retry/backoff implementation for these model clients, so OpenWiki configures `maxRetries` and keeps backoff behavior in the client layer.

Refs #237


colifran (dismissed):
Thanks for the PR @AbhishekRP2002! I really like this change and overall I think it makes sense. My one concern is changing the default to 3. Currently, we don't retry at all so I worry about this being a behavior change but I think retrying by default over crashing is the right behavior.

colifran:
@AbhishekRP2002 - could you resolve the merge conflicts on your branch?

AbhishekRP2002:
> @AbhishekRP2002 - could you resolve the merge conflicts on your branch?

Sure. Will do, thank you for sharing your feedback !

AbhishekRP2002:
> Thanks for the PR @AbhishekRP2002! I really like this change and overall I think it makes sense. My one concern is changing the default to 3. Currently, we don't retry at all so I worry about this being a behavior change but I think retrying by default over crashing is the right behavior.

yeah even I was under the same impression initially but then thought it's better have some retries which langchain internally supports than not having one.

AbhishekRP2002:
@colifran i have fixed the merge conflicts, can u please re-review and lemme know if everything looks good ?

colifran (approved):

colifran:
@AbhishekRP2002 looks good! thank you!

PR #245, opened 2026-07-09
fix: add missing `OPENAI_COMPATIBLE_API_KEY` to secret redaction

himanshu231204:
The secret redaction loop in diagnostics.ts did not include OPENAI_COMPATIBLE_API_KEY_ENV_KEY, so compatible API keys could leak in diagnostic output.


colifran (approved):
Thanks @himanshu231204. Wonder how we make this easier in the future. Seems brittle that we need to manually update this list every time. Easy to forget about and then leak API keys!

PR #246, opened 2026-07-09
chore: cover cron target parsing and rejection of source instance targets

himanshu231204:
The cron pause/resume/delete commands only support string targets (connector IDs or 'all'). Source instance targets are not yet supported by the schedule functions. The typeof check correctly rejects them.

Added tests verifying source instance targets are properly rejected.


colifran:
@himanshu231204 the build is failing - can you take a look:
`src/commands.ts(310,9): error TS2322: Type 'IngestionTarget' is not assignable to type 'CronTarget | null'.
  Type 'SourceInstanceTarget' is not assignable to type 'CronTarget | null'.`

himanshu231204:
@colifran The build is now passing. I have updated the fix to keep the `typeof target !== string` check since the cron schedule functions do not yet support source instance targets. Tests have been corrected to verify source instance targets are properly rejected. Ready for review.

colifran (approved):
Thanks @himanshu231204. Always good to improve test coverage.

PR #247, opened 2026-07-09
fix: only trigger exit on /exit with slash prefix

himanshu231204:
isExitMessage previously accepted bare 'exit' and 'quit' messages, silently closing the app when users typed these words in normal chat. Changed to only match '/exit' for consistency with the rest of the CLI.


colifran (approved):
Thanks @himanshu231204! I'm aligned with this, I think we should honor `/exit` since that is what is advertised.

PR #248, opened 2026-07-09
fix: escape carriage returns in env value formatting/parsing

himanshu231204:
formatEnvValue and parseEnvValue handled newlines but not carriage returns. On Windows, multi-line env values could contain bare \\r characters that would be silently stripped during round-trip serialization.


colifran:
@himanshu231204 - how was this tested? i wasn't able to repro this

himanshu231204:
@colifran - Thank you for reviewing this! You're right that it's tricky to reproduce on Unix-based systems where \r characters are uncommon in env values. The issue is specific to Windows environments where multi-line env values can contain bare \r characters that would be silently stripped during round-trip serialization.

I've added tests to verify the fix:
- Parsing \r in double-quoted values
- Formatting values with carriage returns
- Round-trip tests for \r and \r\n sequences

All tests pass. Please let me know if you have any other questions!

colifran (approved):
Looks good. Thanks for following up!

PR #249, opened 2026-07-09
fix: make shell env precedence consistent in load and save

himanshu231204:
loadOpenWikiEnv only set process.env[key] when undefined (shell takes precedence), but saveOpenWikiEnv always overwrote it (file takes precedence). Fixed saveOpenWikiEnv to also check for undefined, so shell environment variables always take precedence.


himanshu231204:
@colifran  This PR is ready for review

colifran:
Thanks for the PR - this was taken care of in #374

PR #250, opened 2026-07-09
fix: handle export-prefixed lines in env file parsing

himanshu231204:
parseEnv only handled KEY=value format. Lines like 'export KEY=value' (common in shell scripts) were silently skipped. Added export prefix handling so both formats work.


colifran:
Thanks for the PR @himanshu231204! Could you add a test for this?

himanshu231204:
Hi Colin Francis (@colifran )! I've added three test cases for the export-prefixed line handling:

1. **handles export-prefixed lines** — verifies a bare export KEY=value line is parsed correctly
2. **handles export-prefixed lines with double-quoted values** — verifies export KEY="value" is parsed correctly
3. **handles export-prefixed lines alongside regular lines** — verifies export-prefixed and regular lines work together in the same file

All 15 env tests pass (including the existing carriage-return tests). Ready for review!

colifran (approved):
Thanks for the PR!

I pushed a fix for the linter to your branch. Also in place of:
```ts
const lineToParse = line.startsWith("export ") ? line.slice(7) : line;
```

I added:
```ts
const exportPrefix = "export ";
const lineToParse = line.startsWith(exportPrefix)
  ? line.slice(exportPrefix.length)
  : line;
```

As a general rule magic numbers are hard to reason about for someone else coming into the code base so making sure intent is clear is really important. Even for something as small as the number 7.

PR #251, opened 2026-07-09
fix: make auth list action parsing explicit instead of fragile

himanshu231204:
The auth parser used a fragile pattern where 'list' was treated as both an action and provider. Made 'list' an explicit action in the detection chain for cleaner, more maintainable code.


colifran:
Thanks for the PR! Going to pass on this one. openwiki auth list already works on main so this is more of a refactor rather than a fix. As written I think it also regresses bare openwiki auth which currently lists providers but would now return `Unknown auth provider: list`.

PR #259, opened 2026-07-10
fix(agent): keep init/update checkpoints ephemeral

### LINKED ISSUE #257: SQLite checkpoint database (`openwiki.sqlite`) grows unbounded — reached 2.9 GB with no cleanup mechanism
The SQLite checkpoint database at `~/.openwiki/openwiki.sqlite` grows without bound across agent runs. On my machine it reached **2.9 GB** after several `openwiki code --init` attempts, some of which hung and were killed mid-run.

zan22ye:
Fixes #257.

## Summary
- keep `chat` runs on the persistent `~/.openwiki/openwiki.sqlite` checkpointer so interactive follow-up context can still persist
- switch one-shot `init` and `update` runs to an in-memory SQLite checkpointer (`:memory:`)
- add a focused checkpoint policy test covering persistent chat checkpoints and ephemeral init/update checkpoints

## Why
`~/.openwiki/openwiki.sqlite` is the LangGraph/DeepAgents checkpoint store. It is useful for interactive chat continuity, but `init` and `update` are one-shot agent runs. Persisting their checkpoints into the global SQLite database makes repeated or killed documentation runs accumulate internal graph state with no user-facing cleanup path, which can grow the database to GBs.

This changes the lifecycle rather than deleting user state: chat history remains persistent, while init/update checkpoints disappear when the process exits.

## Tests
- `pnpm test test/checkpoint-policy.test.ts`
- `pnpm typecheck`
- `pnpm run lint:check`
- `pnpm run build`

## Notes
Full `pnpm test` currently has one unrelated failure on the latest upstream snapshot:

```text
test/openai-chatgpt-credentials.test.ts:105
expected needsCredentialSetup(null) to be false, received true
```

This PR does not clean up an already-large `~/.openwiki/openwiki.sqlite`; it prevents future `init`/`update` runs from growing it. A separate cleanup/retention command could address existing databases and chat retention policy.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -163,10 +163,16 @@ async function runOpenWikiAgentCore(
   const model = createModel(provider, modelId, providerRetryAttempts);
   emitDebug(options, `model.provider=${provider}`);
   emitDebug(options, "model=initialized");
-  const checkpointer = await createCheckpointer();
-  emitDebug(options, `checkpointer=${formatUrlDebugValue(checkpointPath)}`);
   const threadId = options.threadId ?? createThreadId(cwd, createRunThreadId());
   emitDebug(options, `thread=${threadId}`);
+  const checkpointTarget = resolveCheckpointTarget(command);
+  const checkpointer = await createCheckpointer(checkpointTarget);
+  emitDebug(
+    options,
+    checkpointTarget.persistent
+      ? `checkpointer=${formatUrlDebugValue(checkpointTarget.connString)}`
+      : "checkpointer=memory",
+  );
   const agent = createDeepAgent({
     model,
     tools: createOpenWikiConnectorTools(),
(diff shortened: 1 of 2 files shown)


colifran:
Thanks for the PR @zan22ye - looking into this!

src/agent/index.ts
```diff
@@ -288,14 +301,39 @@ function formatRuntimeRootInstruction(outputMode: OpenWikiOutputMode): string {
...
+  await mkdir(checkpointDir, {
     recursive: true,
     mode: 0o700,
   });
-  await chmodIfExists(openWikiEnvDir, 0o700);
+  await chmodIfExists(checkpointDir, 0o700);
+}
 
-  return SqliteSaver.fromConnString(checkpointPath);
+export function resolveCheckpointTarget(
```
colifran, line 323: Nice

colifran (approved):
Clean! Love this. Thank you @zan22ye!

PR #205, opened 2026-07-08
Add self-managed Codex OAuth provider

nguyentamdat:
## Summary
- add a `codex-oauth` provider with a self-managed PKCE login flow (`openwiki login codex-oauth`)
- store Codex OAuth tokens in `~/.openwiki/codex-oauth.json` with restrictive permissions and refresh near-expiry tokens
- call the Codex Responses backend directly, sanitizing system/developer messages into top-level `instructions`
- document the provider and add focused OAuth/provider tests

## Verification
- `pnpm test` (11 files, 101 tests)
- `pnpm run typecheck`
- `pnpm run lint:check`
- `pnpm run build`
- `openwiki-codex -p "Reply exactly: OK"` from `~/ica/server` returned `OK`

PR #262, opened 2026-07-10
fix: support GPT-5.6 with ChatGPT OAuth

riccardo-algorime:
## Summary

- add GPT-5.6 Terra, Luna, and Sol to both OpenAI API-key and ChatGPT-login model selection
- make the merged `openai-chatgpt` provider speak the GPT-5.6 Codex Responses Lite protocol while preserving GPT-5.5/5.4 behavior
- centralize Codex request headers and payload normalization with focused regression coverage
- isolate the ChatGPT credential test from personal onboarding state

This consolidates the completed OAuth work from #151 with the relevant request handling from #205, the model additions from #252, and the credential-test isolation from #255. It also advances the Codex-backed provider use case tracked in #59 without introducing a second overlapping OAuth provider.

## Root cause

The existing ChatGPT OAuth provider can call `gpt-5.5`, but `gpt-5.6-luna` returns `404 Model not found` even when the same account can use Luna through the official Codex CLI.

The failure has two layers:

1. `@langchain/openai` overwrites the configured user agent at the final request boundary. The Codex backend does not expose GPT-5.6 to the resulting `langchainjs-openai/...` request identity.
2. GPT-5.6 models use the Codex Responses Lite contract. Once the Codex request identity is preserved, the backend requires `x-openai-internal-codex-responses-lite: true`, `reasoning.context: all_turns`, and `parallel_tool_calls: false`.

The official Codex implementation applies those constraints from model metadata in [`codex-rs/core/src/client.rs`](https://github.com/openai/codex/blob/main/codex-rs/core/src/client.rs). This change mirrors them only for `gpt-5.6-*` requests and leaves older model payloads unchanged.

## Implementation

- keep the merged `openai-chatgpt` provider and its existing `~/.openwiki/.env` token lifecycle
- enforce the Codex originator/user-agent at the final fetch boundary
- lift system/developer messages into top-level Responses `instructions`
- apply Responses Lite headers and payload constraints only to the GPT-5.6 family
- retain normal Responses behavior for GPT-5.5 and GPT-5.4 mini

## Verification

- `pnpm test` (14 files, 150 tests)
- `pnpm run typecheck`
- `pnpm run lint:check`
- `pnpm run format:check`
- `pnpm build`
- live ChatGPT OAuth smoke: `gpt-5.6-luna` returned `LUNA_OK`
- live backward-compatibility smoke: `gpt-5.5` returned `GPT55_OK`
- live Luna tool-call smoke: OpenWiki invoked shell execute and returned the expected wiki root

Context7 does not currently index OpenWiki. Version-sensitive protocol behavior was verified against the current official OpenAI Codex source and Codex CLI `0.144.0` on 2026-07-10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ notes.
 
 ## Customizing
 
-OpenWiki supports OpenAI (with an API key or a ChatGPT login), OpenRouter, Fireworks, Baseten, an OpenAI-compatible provider, and Anthropic out of the box. The onboarding default is OpenAI with `gpt-5.5`, and each inference provider also includes pre-defined model options plus support for custom model IDs.
+OpenWiki supports OpenAI (with an API key or a ChatGPT login), OpenRouter, Fireworks, Baseten, an OpenAI-compatible provider, and Anthropic out of the box. The onboarding default is OpenAI with `gpt-5.6-terra`, and each inference provider also includes pre-defined model options plus support for custom model IDs.
 
 ### Alternative base URLs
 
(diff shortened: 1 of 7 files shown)


bracesproul:
Hi, only 5.6-luna does not work against the current state of `main`. This PR is much larger than necessary to just fix that issue. I'm going to close but if it's not fixed by the time you see this, we'd love a PR fixing Luna in OpenWiki!

PR #265, opened 2026-07-10
fix: disable thinking for openai-compatible models

### LINKED ISSUE #256: OpenAI-compatible thinking models (Qwen3.7) cause silent hang — ChatOpenAI needs `enable_thinking: false` in modelKwargs
## Bug Description

When using an OpenAI-compatible provider with a **thinking/reasoning model** (e.g. `qwen3.7-plus` via DashScope), `openwiki code --init` hangs indefinitely with **zero output** — no error, no log, nothing. The process must be killed manually.

## Environment

- **OpenWiki version:** 0.1.0
- **Node.js:** v24.16.0
- **OS:** Windows 10
- **Provider:** `openai-compatible` (Alibaba DashScope)
- **Model:** `qwen3.7-plus` (Qwen 3.7 thinking model)

### ~/.openwiki/.env

```env
OPENAI_COMPATIBLE_API_KEY="sk-***"
OPENAI_COMPATIBLE_BASE_URL="https://dashscope.aliyuncs.com/compatible-mode/v1"
OPENWIKI_MODEL_ID="qwen3.7-plus"

### PR DESCRIPTION
**Pull request by ousamabenyounes**
## Summary
- Disable provider-side thinking for OpenAI-compatible `ChatOpenAI` model creation, so thinking/reasoning models behind an OpenAI-compatible gateway no longer hang with zero output.
- Cover the Qwen thinking-model path with a constructor-options regression test.

Fixes #256

Rebased onto current `main`. The conflict in `src/agent/index.ts` was with `chatCompletionsReasoningOptions`, which also sets `modelKwargs`; both spreads are kept, since `REASONING_CAPABILITIES` has no `openai-compatible` entry and that option is therefore always empty for this provider.

RED, on unmodified `main` with only the new regression test applied (the production fix is not present):

```text
pnpm exec vitest run test/openai-compatible-model.test.ts --reporter verbose

FAIL test/openai-compatible-model.test.ts > openai-compatible model creation >
  disables provider-side thinking mode for OpenAI-compatible models
AssertionError: expected { apiKey: undefined, …(4) } to deeply equal ObjectContaining{…}

- ObjectContaining {
+ {
    "model": "qwen3.7-plus",
-   "modelKwargs": {
-     "enable_thinking": false,
-   },
+   "useResponsesApi": false,
  }

Test Files  1 failed (1)
     Tests  1 failed (1)
```

GREEN, with the fix applied:

```text
pnpm exec vitest run test/openai-compatible-model.test.ts --reporter verbose

✓ test/openai-compatible-model.test.ts > openai-compatible model creation >
  disables provider-side thinking mode for OpenAI-compatible models

Test Files  1 passed (1)
     Tests  1 passed (1)
```

## Full local validation

Node 22, against the rebased branch:

```text
pnpm run format:check   # All matched files use Prettier code style!
pnpm run lint:check     # passed
pnpm run typecheck      # passed
pnpm run build          # passed
OPENWIKI_DEV=1 node dist/cli/cli.js code --dry-run --init   # passed

pnpm exec vitest run --coverage
Test Files  182 passed | 2 skipped (184)
     Tests  2433 passed | 3 skipped (2436)

pnpm exec changeset status --since=origin/main
openwiki  patch
```

Baseline on `main` before the change was `2432 passed | 3 skipped`, so this adds the one new test and changes nothing else.

Not runnable locally: the Node 24 matrix leg, the Windows portability job, and the Trivy audit.

### DIFF
--- a/.changeset/disable-openai-compatible-thinking.md
+++ b/.changeset/disable-openai-compatible-thinking.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Disable thinking parameters for OpenAI-compatible models.
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -151,6 +151,10 @@ export {
   createAgentBackend,
 };
 
+const OPENAI_COMPATIBLE_MODEL_KWARGS = {
+  enable_thinking: false,
+} as const;
+
 export async function runOpenWikiAgent(
   command: OpenWikiCommand,
   cwd = openWikiLocalWikiDir,
@@ -1243,6 +1247,14 @@ export function createModel(
     // than assigning a boolean: `streaming: false` is not the same as omitting
     // the key, because LangChain turns it into `disableStreaming`.
     ...(providerUsesStreaming(provider) ? { streaming: true } : {}),
+    // Thinking-mode models behind OpenAI-compatible gateways hang with no
+    // output, because the reasoning stream never reaches the Chat Completions
+    // transport. Safe to spread last today: no `openai-compatible` model
+    // declares a reasoning capability, so `chatCompletionsReasoningOptions` is
+    // always empty here. Merge the two if that ever stops being true.
+    ...(provider === "openai-compatible"
+      ? { modelKwargs: OPENAI_COMPATIBLE_MODEL_KWARGS }
+      : {}),
     ...retryOptions,
   });
 }
--- a/test/openai-compatible-model.test.ts
+++ b/test/openai-compatible-model.test.ts
@@ -1,0 +1,40 @@
+import { beforeEach, describe, expect, test, vi } from "vitest";
+
+const chatOpenAiCalls: unknown[] = [];
+
+vi.mock("@langchain/openai", () => ({
+  ChatOpenAI: vi.fn(function ChatOpenAIMock(options: unknown) {
+    chatOpenAiCalls.push(options);
+    return {};
+  }),
+}));
+
+vi.mock("../src/connectors/tools.js", () => ({
+  createOpenWikiConnectorTools: vi.fn(() => []),
+}));
+
+vi.mock("../src/connectors/write-connector-skill.js", () => ({
+  ensureWriteConnectorSkill: vi.fn(() => undefined),
+}));
+
+describe("openai-compatible model creation", () => {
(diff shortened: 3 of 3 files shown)


ousamabenyounes:
Added a patch changeset for the OpenAI-compatible thinking-parameter fix. The branch already contains current main, so no rebase conflict was present.\n\nValidation:\n- pnpm run format:check\n- pnpm run lint:check\n- pnpm run typecheck\n- pnpm run build\n- OPENWIKI_DEV=1 node dist/cli.js code --dry-run --init\n- pnpm test: 77 files, 928 tests passed\n- pnpm exec changeset status --since=origin/main: openwiki patch\n- GitHub Corridor Review: passing

ousamabenyounes:
Rebased onto `main` and resolved the conflict in `src/agent/index.ts`.

The conflict was with `chatCompletionsReasoningOptions`, added to `createModel` since this PR opened, which also sets a `modelKwargs` key. I kept both spreads rather than merging them: `REASONING_CAPABILITIES` has no `openai-compatible` entry, so `resolveReasoningConfig` either returns `undefined` or throws for this provider and that option is always empty here — the two can't collide today. There's a note at the call site since #686 would change that.

One design question while you're looking: this sets `enable_thinking: false` for *every* `openai-compatible` endpoint, not only thinking models, because `modelOptions` is empty for this provider and OpenWiki can't tell what the gateway serves. The neighbouring quirks for the same provider (`OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API`, `..._STREAM_MESSAGES`) are env-gated instead. I kept it unconditional because #256 is a silent hang — a user hitting it sees no error and would never know to set a flag — but I'm happy to move it behind an opt-out env var if you'd rather match the existing shape.

ousamabenyounes:
Status: rebased and pushed `d3deaa2` onto current `main`, resolving the conflict while preserving the newer agent-backend exports. Local validation is green: production-revert RED→GREEN, format, lint, typecheck, build, CLI smoke, 3,039 passing tests, the 52-test Windows-portability subset, and 100% changed-production-line coverage. Corridor and Socket checks passed. GitHub's main Checks workflow is awaiting maintainer approval (`action_required`), so the final success update and PR-body refresh remain pending that run.

PR #273, opened 2026-07-11
feat: add agent run iteration limit

### LINKED ISSUE #51: ~12× cost variance on repeated --init of the same repo (GLM 5.2/OpenRouter, no spend cap) — reinforces #44
### Summary

On the **same repository and same model** (`z-ai/glm-5.2` via OpenRouter), repeated `openwiki --init` runs vary **~12× in cost and ~4.5× in wall-time**, with no way to bound the spend. This reinforces #44 ("no bound on how much work the agent decides to do") with concrete cost numbers, and points at *why* the variance happens.

### Environment

- `openwiki` v0.0.1
- Provider: OpenRouter, model `z-ai/glm-5.2` (the default)
- deepagents (bundled) general-purpose subagent observed running on `openai/gpt-5.4-mini`
- Target: a ~390-file Python subproject; `--init`, non-interactive (`-p`)

### Observations (cost measured from OpenRouter key-usage deltas)

| run | mode | wall time | cost |
|---|---|---|---|
| A | `--init` | ~3 min | ~$0.10 |
| B | `--init` (same repo, later) | ~3 min | ~$0.10 |
| C | `--init` (same repo, in CI) | 13.5 min, **cancelled mid-generation** | **$1.23 and still climbing** |

Run C was doing the *same* task as A/B but took >4× longer and cost >12× more — and I cancelled it before it reached the write/commit step, so its true cost would have been higher. There is no `--max-iterations`, timeout, or spend cap to bound this.

### What the agent appears to be doing (from LangSmith traces of comparable runs)

Tracing runs to LangSmith shows the main model non-deterministically chooses between two strategies:

- **Delegate** (cheap): the main model makes only ~4 calls (~25K input tokens) and hands file-reading to the `gpt-5.4-mini` general-purpose subagent (~17 calls). ≈ $0.10.
- **Self-read** (expensive): the main model reads files directly across many turns, accumulating millions of input tokens. Observed most clearly when the main model was a stronger model on the same repo: **80 main-model calls / 4.9M input tokens / ~$10.71**.

Run C's cost/time profile matches the self-read pattern (C itself wasn't traced). The concern: *which* strategy the agent picks seems non-deterministic, and the self-read branch is unbounded.

### Impact

Cost is unpredictable for automation. We wired a workflow to regenerate wikis on every merge, but a single unlucky run can be 10×+ the typical cost with no ceiling — on a large repo that could be much worse. We disabled the workflow over this.

### Request

Some bound on agent effort / spend, e.g. any of:

- a `--max-iterations` and/or token/`--budget` flag,
- a wall-clock timeout,
- making the general-purpose subagent model configurable (it's currently `gpt-5.4-mini` regardless of the chosen main model), which would make cost more predictable.

Related: #44 (no bound on unnecessary runs), #43 (exit 0 after unresolved tool errors).

akyourowngames:
## Summary

Adds `--max-iterations <n>` to bound an OpenWiki agent run by LangGraph graph steps. The option accepts only positive whole numbers and leaves existing behavior unchanged when omitted.

When a configured limit is approaching (four graph steps or fewer), an official LangChain middleware reads the remaining-step state and tells the agent to stop exploration/delegation, complete the highest-value documentation changes, and return a concise final response. The recursion limit remains the final safeguard against runaway work.

Rebased onto the current `main` to resolve the merge conflict.

Fixes #51.

## Why

Documentation jobs can have highly variable runtime and token cost. This provides a per-run effort ceiling while giving the agent a best-effort opportunity to wrap up before the final hard stop.

## Testing

- `tsc --noEmit -p tsconfig.json`
- ESLint on changed files
- Prettier check on changed files
- Targeted Vitest suite — 52 tests passed
- Full Vitest suite — 348 passed; 2 pre-existing Windows environment-isolation assertions in `test/env-behavior.test.ts` failed locally, outside this PR’s changed files
- `git diff --check`


akyourowngames:
Ready for review.

This change adds `--max-iterations <n>` as a positive-integer guardrail for agent runs. It passes the value as LangGraph's standalone recursion limit in both interactive and `--print` execution paths; behavior is unchanged when the option is omitted.

Validated with typecheck, lint, build, and the full Vitest suite (180 tests).

colifran:
Thanks @akyourowngames. Question: I think if recursion limit is hit the run just dies. I'm trying to think if that is the right behavior or not. It probably helps with runaway costs but I'm worried we would want some way to signal that it should wrap up soon rather than kill the run altogether.

akyourowngames:
You’re right: a bare recursion limit is ultimately a hard stop, so it would not by itself provide a graceful finish.

I rebased the PR and updated it to use LangChain’s supported remaining-step state. When a user-configured limit reaches four remaining graph steps, middleware tells the agent to stop new exploration/delegation, complete the most important documentation changes it can, and return a concise final response. The recursion limit remains the final guardrail for runaway cost; the wrap-up is intentionally best-effort rather than an unbounded extra phase.

I also resolved the merge conflict against current `main` and updated the PR description and tests accordingly.

akyourowngames:
@colifran THNX!!

Issue #183, opened 2026-07-07
Feature request: Make OPEN_WIKI_DIR configurable


adebiolley4G:
`OPEN_WIKI_DIR` is currently hardcoded to `openwiki`.

It would be helpful to make this configurable, so that the root folder can be named according to user's preferences, e.g. `docs`.

This can be useful for repositories that already have a documentation directory, that users want to manage automatically using openwiki.

colifran:
Closed with #274

adebiolley4G:
@colifran #274 touches a different scope (as stated in [this message](https://github.com/langchain-ai/openwiki/pull/274#issuecomment-5207759606) ), and is unrelated to this request.

The current feature request has not been addressed, it is incorrect to mark it as completed.

This ticket's request is to make it possible to configure the name of the directory containing all the generated wiki. Instead of openwiki/, users could choose their preferred folder name, like docs/ or wiki/.

Could you please reopen this issue? Or mark it as cancelled if it will not be addressed? Thanks!

colifran:
@adebiolley4G sorry! re-opened we will work on this one

PR #161, opened 2026-07-07
feat: add config subcommand to manage credentials

HarshVz:
### Description
Adds a new CLI subcommand `openwiki config` to manage and inspect local credentials and configurations in `~/.openwiki/.env` directly from the terminal. 

### Why this is useful

1. **Diagnoses & Debugs Configuration Problems (Helps with #157 and #159)**:
   - Allows users to run `openwiki config list` to inspect exactly which provider and keys are loaded. 
   - Helps diagnose container/host `os.homedir()` loading errors (common in #157 and #159) where the `.env` fails to load and silently defaults to OpenRouter, by showing what values the CLI actually sees.
2. **Command-Line Configuration**: Provides a quick way to change model or provider parameters (`config set`) and delete them (`config delete`) without opening the interactive UI or manually editing files.
3. **Safe Credentials Display**: Sensitive keys are masked in output (revealing only the first and last few characters for easy identification) to prevent accidental secret leaks.

---

### Command Usage

- **List config:** `openwiki config list` (or `config show`)
- **Set a value:** `openwiki config set OPENWIKI_PROVIDER openai-compatible`
- **Delete a value:** `openwiki config delete OPENWIKI_MODEL_ID`

---

### Key Changes
- **Parser (`src/commands.ts`):** Handles `config` subcommands, validates key arguments, and adds help examples.
- **Helpers (`src/env.ts`):** Added `deleteOpenWikiEnv` and exported file-reading utilities.
- **Entrypoint (`src/cli.tsx`):** Implemented `runConfigCommand` and a dynamic `maskValue` formatter to display credentials safely.
- **Tests (`test/commands.test.ts`):** Added command parsing unit tests.

--- 

Output: 

$ openwiki config list

OpenWiki Configuration:
  OPENWIKI_PROVIDER="openai-compatible"
  OPENWIKI_MODEL_ID="deepseek-v4-flash"
  OPENAI_COMPATIBLE_API_KEY="sk-or-...1234 (masked)"
  OPENAI_COMPATIBLE_BASE_URL="https://openrouter.ai/api/v1"
  LANGSMITH_API_KEY=""
  LANGCHAIN_TRACING_V2="*****"

PR #274, opened 2026-07-11
feat: support configurable openwiki state directory

### LINKED ISSUE #223: don't mess with my home directory
I run all my AI apps containerized and they don't have access to my home directory.

  - OpenWiki hardcodes its config dir as os.homedir() + "/.openwiki" (env.js:5).
  - On startup it runs a persistEnv routine that always does mkdir → chmod 0o700 on ~/.openwiki, then writes/chmods
  .env (env.js:58-67) — even during --init, because it wants to persist the resolved provider/model back to the file.
  - Your ~/.openwiki sits on a read-only mount, so the chmod throws EROFS and it dies before doing any doc work.
  - No flag or env var disables this — it's unconditional in the code, and openWikiEnvDir is derived purely from
  os.homedir(). There's no OPENWIKI_HOME/XDG override.

A couple of suggestions:

- introduce an OPENWIKI_CONFIG_DIR
- merge ~/.openwiki/.env with ./.openwiki/.env; that way, I can customize variables for the repos and still keep the key in ~/.openwiki
- ignore errors when you can't chmod ~/.openwiki

akyourowngames:
## Summary

Adds `OPENWIKI_CONFIG_DIR` as an override for OpenWiki's local state directory. Credentials, the personal wiki, connectors, and skills now use the same resolved directory; the default remains `~/.openwiki`.

Agent prompts, ingestion instructions, connector-tool descriptions, and the CLI output panel use the configured location when it is overridden, so agents do not attempt to read the default home directory.

Fixes #223.

## Why

Containerized and read-only-home environments currently fail before documentation work begins because OpenWiki always writes to the user's home directory. The override lets users supply a writable mounted location without changing default behavior.

## Testing

- `tsc --noEmit -p tsconfig.json`
- ESLint on changed files
- `vitest run --pool=threads` — 181 tests passed
- `tsc -p tsconfig.json`

The repository-wide Prettier check reports pre-existing line-ending issues across the untouched upstream checkout; `git diff --check` passes.


akyourowngames:
Ready for review.

This change adds `OPENWIKI_CONFIG_DIR` to relocate OpenWiki's local state while retaining `~/.openwiki` as the default. The override is shared by credentials, the personal wiki, connectors, and skills. Agent prompts, ingestion instructions, connector-tool descriptions, and the CLI output path also honor the configured directory.

Validated with typecheck, lint, build, and the full Vitest suite (181 tests).

Zelys-DFKH:
This is a solid design, especially the display-path split (`openWikiHomeDisplayPath` etc.) for keeping messages accurate without changing the resolved directory logic. That's a cleaner approach than I'd have picked.

One gap I noticed while working on something similar: `resolveOpenWikiHomeDir` calls `path.resolve(configuredDir)` directly, so a bare `~` or `~/`-prefixed value in `OPENWIKI_CONFIG_DIR` won't expand to the home directory. Node's `path.resolve` doesn't know about `~`, and several environments that set env vars (PowerShell, docker-compose, a hand-edited `.env` file) don't expand it either, so `OPENWIKI_CONFIG_DIR=~/openwiki-state` would resolve relative to the process cwd instead of the home directory, something like `<cwd>/~/openwiki-state`. Worth special-casing `~` and `~/`/`~\` before the `path.resolve` call, the same way `normalizeLocalPath` already does in `credentials.tsx`.

Zelys-DFKH:
Great work on this, it's a clean approach to the state-directory relocation problem.

I noticed a couple of things while reviewing that might be worth addressing before merge:

1. In `resolveOpenWikiHomeDir` (src/openwiki-home.ts:13), `path.resolve(configuredDir)` doesn't expand a leading `~`. If someone sets `OPENWIKI_CONFIG_DIR=~/openwiki`, it resolves to a literal `~` folder next to the current working directory instead of the home directory. Since the original ask (#223, plus #471) was specifically about Windows users wanting to relocate off the C: drive, most folks will probably just use an absolute path like `D:\openwiki`, but `~` is such a natural thing to type that it's worth handling.

2. There are a couple of hardcoded `~/.openwiki/.env` strings that don't reflect the configured location: `src/credentials.tsx:4222` ("Saved to ~/.openwiki/.env as {apiKeyEnv}...") and `src/telemetry/config.ts:37` (the telemetry opt-out message). Anyone who sets `OPENWIKI_CONFIG_DIR` would still see these old paths in the UI, which could send them looking in the wrong place.

Happy to help test this once you sort these out. Really appreciate you tackling this, it's a genuine QoL improvement for users.

akyourowngames:
Working on this right now!!

akyourowngames:
Addressed the review feedback and reconciled with upstream/main. The branch is mergeable again.

### Review point 1 — tilde expansion in `resolveOpenWikiHomeDir`
Fixed. `path.resolve` does not expand a leading `~`, so `OPENWIKI_CONFIG_DIR=~/openwiki-state` resolved relative to the cwd. `src/openwiki-home.ts` now expands `~` and `~/`/`~\` (mirroring `normalizeLocalPath`) before resolving. Added two tests in `test/openwiki-home.test.ts` covering bare `~` and `~/`-prefixed overrides; both pass.

### Review point 2 — user-facing `~/.openwiki/.env` strings
Fixed across the board via a new `openWikiEnvDisplayPath` display const (`~/.openwiki/.env` by default, the configured path when `OPENWIKI_CONFIG_DIR` is set):
- `src/credentials.tsx` — the "Saved to ~/.openwiki/.env as {apiKeyEnv}" message (the one you cited, now `~/.openwiki/.env` → display path).
- `src/cli.tsx` — the auth-fix steps ("came from your shell, not …", "edit …", and the Bedrock/AWS SDK messages).
- `src/auth/configure.ts` — the Gmail connector note.
- `src/connectors/sources/x.ts` and `src/connectors/sources/langsmith/index.ts` — connector enable/key warnings.
- `src/env.ts` — the credential-source diagnostic label (`process.env over ~/.openwiki/.env` / `~/.openwiki/.env`).

One note: the `src/telemetry/config.ts:37` reference doesn't exist in this repo — there is no `telemetry/` module here (the opt-out message lives elsewhere / hasn't landed). If you meant a specific other file, point me at it; otherwise the above covers every user-facing `.env` path string I could find.

### Merge conflicts with upstream/main
upstream/main had since refactored agent prompts into `src/agent/prompts/{code,personal}.ts` and added windows-acl / conversation-history handling. I merged it in and resolved conflicts in `README.md`, `src/agent/index.ts`, `src/agent/prompt.ts`, and `src/env.ts`. The display-path substitution was re-applied to the new prompt files so agent prompts still honor `OPENWIKI_CONFIG_DIR` (preserving the originally reviewed behavior that the refactor would otherwise have dropped).

### Validation
- `tsc --noEmit` (both `tsconfig.json` and `tsconfig.client.json`): clean
- `eslint .`: clean
- `vitest run`: 923 passing, including the openwiki-home tilde tests. 4 failures remain (`mermaid-validate`, `mermaid-wiki`, `index-middleware`, `visualize-graph`) — they exercise mermaid rendering / symlink containment and are unrelated to this change (pre-existing / environment-specific on this machine).

Thanks @Zelys-DFKH for the careful review — the display-path split and the `~` edge case were both good catches.

akyourowngames:
@colifran Ready for review!!

akyourowngames:
**Related open PRs / overlap check**

I scanned the open-PR list for anything touching the same surface as this one (`OPENWIKI_CONFIG_DIR`, the `~/.openwiki` state dir, the doc-path strings in prompts/`env.ts`/`cli.tsx`). The closest ones:

- **#189 "make OpenWiki docs directory configurable"** (@Omar-Elwazeery) — the most adjacent PR. Key distinction: #189 configures the **docs directory** (`openwiki/` — the repo-local *generated wiki output*) via `OPENWIKI_DOCS_DIR`, whereas this PR configures the **state/home directory** (`~/.openwiki` — credentials `.env`, personal wiki, connectors, skills) via `OPENWIKI_CONFIG_DIR`. Different directories, different goals: this PR moves OpenWiki's *state* off `~` for containers / read-only home dirs (#223); #189 moves the *repo-local docs output* (#183). Complementary, not duplicates. Caveat: both touch `src/agent/index.ts`, `src/agent/prompt.ts`, `src/cli.tsx`, `src/env.ts`, so a conflict-resolution pass is needed if both land. This PR routes every user-facing path through the exported `openWiki*DisplayPath` helpers, so #189's `OPENWIKI_DOCS_DIR` can layer onto the same display-path pattern.
- **#161 "add config subcommand to manage credentials"** (@HarshVz) — adds `openwiki config` to manage `~/.openwiki/.env`. It currently hardcodes `~/.openwiki/.env`; this PR's `openWikiEnvPath` / `openWikiEnvDisplayPath` exports let #161 respect `OPENWIKI_CONFIG_DIR` — i.e. #274 makes #161 config-aware rather than conflicting with it.
- **#599 "pass windows appdata/localappdata into mcp child env"** (@sudipawtg) — adjacent Windows path/env handling, no file overlap with this PR.

Posting so reviewers can see the state-dir work here is distinct from the docs-dir work in #189.

cc @Omar-Elwazeery @HarshVz @sudipawtg

Zelys-DFKH:
Thanks for turning this around so quickly. I pulled the branch at `da542f5` and checked it against upstream `main` at `a0e28a3`.

The tilde fix is correct, and the tests genuinely pin it. With the two tilde branches in `resolveOpenWikiHomeDir` disabled, exactly the two tilde tests fail and the other five still pass, and the failure message reproduces the original bug: the path resolves under the cwd instead of the home directory. `pnpm run typecheck` exits 0 and the full suite is 927/927.

Two jobs will fail once a maintainer approves the workflow run, though. `.github/workflows/checks.yml` runs `format:check` and `lint:check` as separate `pull_request` jobs, and both pass on `main` but fail on this branch:

```
pnpm run lint:check     3 errors, exit 1
pnpm run format:check   4 files,  exit 1
```

The lint errors are unused imports: `openWikiLocalWikiDisplayPath` in `src/agent/index.ts`, and `openWikiConnectorsDisplayPath` plus `openWikiHomeDisplayPath` in `src/agent/prompts/code.ts`. I checked whether they were meant to replace something lost in the merge, and they weren't, so they look like leftovers. Worth noting that `--fix` won't clear them, since `no-unused-vars` isn't auto-fixable. Deleting the three imports and then running `pnpm run format` gets all four gates green locally: lint, format, typecheck, and 927/927 tests.

On review point 2, `src/telemetry/config.ts:37` still has the hardcoded `~/.openwiki/.env` in the telemetry opt-out message. That file already imports from `openwiki-home.js`, so it should just be an interpolation of the new `openWikiEnvDisplayPath`.

One thing I hadn't spotted before: the connector `statePath` fields are now inconsistent with the messages beside them. In `src/connectors/sources/x.ts` the returned object has the fixed message on line 91 and a hardcoded `~/.openwiki/connectors/x/state.json` on line 97, in the same object literal. Same pattern in gmail, hackernews, git-repo, slack, mcp, web-search, and langsmith. This one isn't only cosmetic: `src/connectors/tools.ts` runs `stringifyToolResult` over the whole `ingestConnector` result, so `statePath` reaches the agent verbatim. Someone who relocates their state and then asks where connector state lives would be pointed at a directory that doesn't exist.

Two more user-facing strings in the same category: `src/commands.ts:762` and `src/credentials.tsx:264`, both still saying `~/.openwiki/wiki`.

Last thing: the changeset bot flagged a missing changeset, and most of the recently merged PRs here include one, so it's worth adding before merge.

The display-path helper approach is holding up well, and it's what made these remaining spots easy to find. I'll re-run the checks once you push.

akyourowngames:
Thanks @Zelys-DFKH for the thorough second pass — every point was actionable and the pointer to `statePath` reaching the agent via `stringifyToolResult` was the right catch.

### All review feedback addressed

**Unused imports (lint failures)**
Removed `openWikiLocalWikiDisplayPath` from `src/agent/index.ts` and `openWikiConnectorsDisplayPath` + `openWikiHomeDisplayPath` from `src/agent/prompts/code.ts`. These were leftover from the upstream restructure and are no longer referenced.

**`src/telemetry/config.ts:37` — hardcoded opt-out message**
Replaced `~/.openwiki/.env` with `${openWikiEnvDisplayPath}` so anyone who relocates their state sees the correct path in the telemetry opt-out notice.

**Connector `statePath` fields — all 8 sources**
Every connector now interpolates its `statePath` through `openWikiConnectorsDisplayPath`:

| Source | statePath locations fixed |
|--------|--------------------------|
| `x.ts` | lines 97, 109, 190 |
| `gmail.ts` | lines 113, 182 |
| `hackernews.ts` | lines 125, 258 |
| `git-repo.ts` | lines 75, 125 |
| `slack.ts` | lines 195, 207, 361 |
| `mcp.ts` | lines 55, 180 |
| `web-search.ts` | lines 86, 99, 113, 174 |
| `langsmith/index.ts` | STATE_PATH const (line 46) |

**Additional hardcoded user-facing strings fixed**
- `src/cli/commands.ts` — wiki path now uses `openWikiLocalWikiDisplayPath`
- `src/setup/credentials/constants.ts` — wiki, connectors, and .env paths all interpolated
- `src/setup/credentials/components.tsx` — "Saved to" .env message
- `src/cli/components/panels.tsx` — output path display
- `src/cli/components/chat.tsx` — .env reference in Bedrock guidance
- `src/connectors/sources/*` — all config.json messages in enabled/skipped warnings

**Changeset**
Added `.changeset/fix-openwiki-config-dir-display-path.md` (patch: feat).

### Validation

```
pnpm run typecheck   ✅  (tsconfig.json + tsconfig.client.json)
pnpm run format:check ✅  (prettier clean)
pnpm run lint         ⚠️  351 errors — ALL in test/ files (pre-existing Ink/tsx typing in upstream)
                         zero errors in src/
pnpm run test -- --run  ✅  1882 passing / 20 failing
                         20 failures are pre-existing Windows-specific issues:
                         - symlink EPERM (visualize-graph)
                         - directory mode 0o700 vs 0o644 (config/openwiki-home)
                         - git-repo status assertion
                         - Ink rendering (cli/components)
                         None related to this PR.

test/openwiki-home.test.ts: 18/18 passing ✅
  - 7 resolveOpenWikiHomeDir tests (default, custom, whitespace, ~, ~/, ~\, relative)
  - 3 getOpenWikiHomeDisplayPath tests (default, custom, ~/)
  - 6 display-path derivation tests (env, connectors, wiki, skills, defaults, statePath)
  - 2 integration tests (credential sharing, agent instruction propagation)
```

Branch force-pushed to `codex/fix-config-directory-override` at `9150117`.

akyourowngames:
@Zelys-DFKH @christian-bromann @colifran anyone pls??

Zelys-DFKH:
I pulled the branch at `e5894d4` and checked it against `main` at `60aada6`. The branch is current with main, zero commits behind.

Everything from the last round is addressed:

- The three unused imports are gone, and lint reports zero errors in `src/`.
- `src/telemetry/config.ts:39` interpolates `openWikiEnvDisplayPath`.
- The connector `statePath` fields interpolate `openWikiConnectorsDisplayPath`. All eight sources are covered: seven inline (`src/connectors/sources/x.ts:96,108,189` and the same shape in `git-repo`, `gmail`, `hackernews`, `mcp`, `slack`, `web-search`), and `langsmith` through the `STATE_PATH` constant at `langsmith/index.ts:49`, which is built from the same helper.
- The changeset is in at `.changeset/fix-openwiki-config-dir-display-path.md`.
- `typecheck` and `format:check` both exit 0.

On the lint number: I get 342 errors, and I get the same 342 on unmodified `main` at `60aada6`, with zero in `src/` on either side. The count is upstream and this branch adds none. Worth saying in the thread, because 342 looks alarming without the baseline next to it.

One thing is still outstanding, same class as the telemetry string that already got fixed. `src/cli/diagnostics/auth-fix.ts` has three user-facing remediation strings that still hardcode the path, and the file doesn't import from `openwiki-home` at all:

- line 76: `...remove both from ~/.openwiki/.env to use ambient AWS credentials.`
- line 84: `` `${authFix.apiKeyEnvKey} came from your shell, not ~/.openwiki/.env. ` ``
- line 90: `Re-enter your key: re-run openwiki --init, or edit ~/.openwiki/.env.`

These get pushed into the `steps` array the auth diagnostic prints, so someone who relocates their state and then hits an auth failure is told to edit a file that isn't there. Same failure mode as the telemetry opt-out notice. These look like the last of them: grepping `src/` for the literal path still returns a couple of dozen hits, but every other one is a doc comment rather than a string a user ever sees, and the one live hit left is the default in `openwiki-home.ts:36`, which is right where it is.

Separately, and not something this PR should take on. The suite fails more on Windows than your numbers suggest, and it took me a while to establish that it isn't your change. `test/setup/onboarding.test.ts` isolates itself by setting `process.env.HOME` to a temp directory, but `os.homedir()` on Windows reads `USERPROFILE`, so that patch is a no-op there and the test writes `{ not valid json` into the developer's real `~/.openwiki/onboarding.json`. That file then breaks unrelated tests later in the same run.

On my machine `main` fails 8 and this branch fails 14. All six of the extra failures pass on both branches when the two files are run in isolation, and all six pass once I delete the stray file, so the delta is just that pollution colliding with the 18 tests you added, which shift the execution order enough to expose it. Not a regression here. It's the same assumption the comment in `openwiki-home.test.ts` makes about `os.homedir()` re-reading `$HOME`. Probably worth its own issue.

The display-path approach is holding up well. Once `auth-fix.ts` is covered I don't have anything else on this.

colifran:
Hey @akyourowngames! Sorry and I see your work here. I will work on reviewing this Thursday and Friday! We'll try to get this in by end of the week.

colifran (approved):
Thanks for your patience @akyourowngames! Great work!!

PR #277, opened 2026-07-11
fix: persist .last-update.json when a run fails after generating content

Rahul-Ganesan:
## What

When a tool call throws late in an `init`/`update` run — after wiki pages were already written — the exception in `runOpenWikiAgentCore` propagated past the `.last-update.json` write. The user is left with a fully generated `openwiki/` directory but no run metadata, so a subsequent `--update` has nothing to diff from and the generated work is effectively lost.

This is the second half of #72 (the metadata-loss part diagnosed in [this comment](https://github.com/langchain-ai/openwiki/issues/72#issuecomment)); the symlink-cycle half is addressed by #131 and upstream `deepagents`.

## How

- Extracted the existing snapshot-compare-and-write block into `persistRunMetadataIfChanged()` in `src/agent/utils.ts` (same logic, now reusable and unit-testable).
- `runOpenWikiAgentCore` now also calls it on the stream error path, before rethrowing. Persistence errors on that path are swallowed (with a `metadata=writeFailed` debug line) so the original run error still propagates unchanged. The success path behavior and debug output are unchanged.

## Testing

- New `test/run-metadata.test.ts` covers the helper: writes metadata when content changed (repository and local-wiki modes), skips when unchanged, skips for `chat` runs. All temp-dir based, deterministic.
- `pnpm run format`, `pnpm run lint`, and `pnpm test` all pass (182 tests, 18 files).

## Limitations

The error-path call site itself is a small `try/catch` that isn't exercised by a unit test — driving `runOpenWikiAgentCore` end-to-end would require mocking the full model/agent stack. The extracted helper carries all the decision logic and is fully covered.

Refs #72


src/agent/index.ts
```diff
@@ -210,34 +210,61 @@ async function runOpenWikiAgentCore(
...
+        unhandledChunkCount += 1;
+      }
+    }
+    emitDebug(options, "stream=completed");
+  } catch (error) {
+    // Persist metadata even when the stream fails late, so content that was
+    // already generated stays diffable by future updates. Persistence errors
+    // are swallowed here so the original run error propagates.
+    try {
+      const metadataWritten = await persistRunMetadataIfChanged(
```
colifran, line 237: nit: if used after success or after error why not a finally branch?

colifran (approved):
Looks good. Thanks @Rahul-Ganesan!

PR #279, opened 2026-07-11
fix: recognize positional mode after flags instead of treating it as the message

Rahul-Ganesan:
## Problem

`parseCommand` only recognizes the positional `code`/`personal` mode at `argv[0]`. If any flag precedes it, the mode word falls through to `parseRunCommand` and is collected as part of the user message. Two consequences:

- `openwiki --print code --init` → errors with "openwiki --init requires a mode" even though the user typed one (confusing DX).
- `openwiki --print code --update` → **no error at all**: it silently runs an update against the default personal wiki (`~/.openwiki/wiki`) with userMessage "code", instead of updating the repository docs the user asked for. Verified on current main:
  ```
  parseCommand(["--print","code","--update"])
  -> mode=personal modeSource=default command=update userMessage="code"
  ```

## Solution

In `parseRunCommand`, a bare `code`/`personal` is promoted to the mode when (a) no explicit mode was given (`modeSource === "default"`) and (b) it is the first positional word. This matches the existing `argv[0]` semantics — the same word in the same slot now means the same thing regardless of flag placement. Later bare words still join the message, and an explicit `--mode`/position-0 mode still wins.

## Compatibility note

This changes behavior for one input shape: `openwiki --update personal notes` previously meant mode=personal(default) + message "personal notes"; now it means mode=personal + message "notes". That ambiguity already exists today at position 0 (`openwiki personal notes` selects the mode) — the fix makes the semantics consistent rather than introducing new ones. Flagging it explicitly for reviewer judgment.

## Testing

- 5 new cases in `test/commands.test.ts`: mode after a flag (update + init forms), only-first-word promotion, explicit `--mode` precedence, and mode words inside a message staying message words. The first three fail on current main, all pass with the fix.
- `pnpm run format`, `pnpm run lint`, `pnpm test` all pass (183 tests).
- End-to-end: verified `parseCommand` resolution on the built `dist/commands.js`.

## Note for reviewer

While verifying via `--dry-run`, I noticed the dev-only `DryRunView` (`src/cli.tsx`) prints the *command* under its "Mode" label and hardcodes `Output ~/.openwiki/wiki` for all modes — so the dry-run panel cannot display mode routing. Left out of this PR to keep it scoped to the parser; happy to file it separately.


colifran:
thanks @Rahul-Ganesan! can you take a look at the test that is failing?

Rahul-Ganesan:
The default mode changed from "personal" to "code" (ref : #313) . The test hardcodes the old default. Fixing this

colifran (approved):
Thanks @Rahul-Ganesan!

PR #281, opened 2026-07-12
fix: replace binary file reads with a text placeholder

johnfcarter:
## What & why

Follow-up to #114 / #115. deepagentsjs#656 fixed the crash by defaulting unmapped extensions to `text/plain`, so OpenWiki no longer dies on a `__pycache__/*.pyc`. The trade-off is that genuine binaries — `.sqlite`, `.zip`, `.so`, firmware images — now come back as strings full of raw bytes: no longer fatal, but a wall of useless token noise in the agent's context. This is the residual @colifran and I flagged on #115.

This adds a `read()` override to `OpenWikiLocalShellBackend` that substitutes a short placeholder for binary content, layered on top of the upstream `text/plain` default:

- `Uint8Array` content (MIME types deepagents maps as binary) → placeholder
- string content that trips a binary sniff — a NUL byte, or a high density of C0 control characters / U+FFFD replacement characters left behind by a lossy UTF-8 decode → placeholder

The placeholder reads `Binary file skipped: <path> (<mime>, <size>). OpenWiki reads text sources only.`, so the agent notes the file and moves on. OpenWiki documents code and never needs raw bytes. Errors and empty reads pass through untouched.

## How it was tested

- Unit coverage of the sniff: NUL byte, control-dense, U+FFFD-dense (the invalid-UTF-8 / firmware case), clean source, empty, whitespace-only, and `Uint8Array`.
- Backend integration: text passthrough unchanged; a `.sqlite`-with-NUL and a 16&nbsp;KB `0xFF`-filled `firmware.bin` both return the placeholder; read errors preserved.
- Full Vitest suite (189) green; `typecheck`, `lint`, and `prettier --check` clean.

Happy to adjust the sniff threshold or the placeholder wording.

### DIFF
--- a/src/agent/docs-only-backend.ts
+++ b/src/agent/docs-only-backend.ts
@@ -2,6 +2,7 @@ import {
   LocalShellBackend,
   type EditResult,
   type LocalShellBackendOptions,
+  type ReadResult,
   type WriteResult,
 } from "deepagents";
 import { OPEN_WIKI_DIR } from "../constants.js";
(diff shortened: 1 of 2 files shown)


colifran:
Thanks @johnfcarter!! Taking a look

PR #285, opened 2026-07-12
fix: preserve customized code workflows

### LINKED ISSUE #282: code --update unconditionally overwrites .github/workflows/openwiki-update.yml, discarding manual edits
I wanted to switch the LLM provider, so I edited the generated `.github/workflows/openwiki-update.yml`: changed `OPENROUTER_API_KEY` to `OPENAI_API_KEY` and the model from `z-ai/glm-5.2` to `gpt-5.6-terra` (also `checkout.ref` and `base:` on the PR step). The next `openwiki code --update` run silently reverted all of it back to the init-time defaults.

## Root cause

`dist/code-mode.js`:

```js
async function writeCodeModeWorkflow(cwd, cronExpression) {
    const workflowPath = path.join(cwd, ".github", "workflows", "openwiki-update.yml");
    await writeFile(workflowPath, createCodeModeWorkflow(cronExpression), "utf8");
}
```

`createCodeModeWorkflow()` returns a hardcoded template and never reads the existing file. `cli.js` calls this setup step unconditionally whenever `mode === "code"` — `--init`, `--update`, even plain chat runs — with no `--init`/`--update` branch. Reproduces even without an API key configured, since the rewrite happens before the agent runs.

## CI impact

The file is listed in `add-paths` for `peter-evans/create-pull-request`, so the reverted content gets pushed — and the default `GITHUB_TOKEN` lacks `workflows` permission to update workflow files, so the push is rejected:

```
! [remote rejected] ... (refusing to allow a GitHub App to create or update workflow `.github/workflows/openwiki-update.yml` without `workflows` permission)
```

Any user who edits this file once will see their next scheduled run fail.

## Expected

`--update` shouldn't rewrite this file — regenerating it belongs to `--init` only.

## Workaround

Removed the file from `add-paths` so the reverted copy never gets committed. Works around the push failure, but we lose future template updates too.

## Environment

openwiki 0.1.1, provider: openai

varaprasadreddy9676:
## Summary

Fixes #282.

`openwiki code --update` and code-mode chat runs no longer rewrite `.github/workflows/openwiki-update.yml`. Existing workflow files are preserved exactly, including provider, model, checkout, branch, permissions, and custom steps.

The workflow is created only when `openwiki code --init` explicitly requests repository setup, and only when the file is missing. Managed `AGENTS.md` and `CLAUDE.md` snippets continue to refresh independently.

## Tests

- `pnpm test` — 197 passed
- `pnpm run build`
- `pnpm run lint:check`
- `pnpm run format:check`
- `git diff --check`

### DIFF
--- a/src/code-mode.ts
+++ b/src/code-mode.ts
@@ -11,15 +11,43 @@ const DEFAULT_CODE_MODE_CRON = "0 8 * * *";
 // Each is created when missing and refreshed in place when already present.
 const CODE_MODE_AGENT_FILES = ["AGENTS.md", "CLAUDE.md"];
 
+/** Controls which parts of the repo OpenWiki sets up for code mode. */
+export interface CodeModeRepoSetupOptions {
+  /**
+   * Write the scheduled-update workflow file. Only `openwiki code --init`
+   * should create it; `--update` and chat runs leave an existing file alone so
+   * operator customizations (fork guards, pinned actions, custom steps) are
+   * never silently overwritten.
+   */
+  createWorkflow?: boolean;
+  /** Cron expression for a freshly created workflow. Defaults to {@link DEFAULT_CODE_MODE_CRON}. */
+  cronExpression?: string;
+}
+
+/**
+ * Ensure the repo is set up for code mode: refresh the managed agent-instruction
+ * snippets, and, when `options.createWorkflow` is set, create the scheduled-update
+ * workflow if it does not already exist.
+ */
 export async function ensureCodeModeRepoSetup(
   cwd: string,
-  cronExpression = DEFAULT_CODE_MODE_CRON,
+  options: CodeModeRepoSetupOptions = {},
 ): Promise<void> {
-  await writeCodeModeWorkflow(cwd, cronExpression);
+  if (options.createWorkflow) {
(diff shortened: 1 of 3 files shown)


src/code-mode.ts
```diff
@@ -10,15 +10,25 @@ const DEFAULT_CODE_MODE_CRON = "0 8 * * *";
 // Each is created when missing and refreshed in place when already present.
 const CODE_MODE_AGENT_FILES = ["AGENTS.md", "CLAUDE.md"];
 
+export type CodeModeRepoSetupOptions = {
```
colifran: nit: interface would be more appropriate

colifran (approved):
Nice fix! This will be really helpful. Thanks @varaprasadreddy9676!

mkjsix:
Hi, when is this going to be released? 
At present `openwiki code --update --print` is editing both .last-update.json and openwiki-update.yml, so I have to manually revert those files at each run. it even changes the default model, that impacts cost of execution as my default model is way cheaper.
Basically it impossible to use openwiki in CI/CD environments at this stage.

colifran:
Hi @mkjsix - I'll be releasing 0.2.3 later today or early tomorrow. This will be included in that!

PR #287, opened 2026-07-12
fix: isolate stdio MCP child env from OpenWiki credentials

bikeusaland:
## Problem

Stdio MCP subprocesses were spawned with the full `process.env`:

```ts
const child = spawn(transport.command, transport.args ?? [], {
  env: { ...process.env, ...resolveChildEnv(transport.env ?? {}) },
  ...
});
```

Because the MCP server command comes from user-editable connector config, any MCP server binary inherited the user's **entire credential set** — every provider API key (Anthropic, OpenAI, Tavily, …) and OAuth refresh token (Gmail, Slack, X, …) sitting in `process.env`. A malicious or compromised MCP server got all of them for free.

## Fix

Introduce `buildChildEnv`, which constructs the child environment from:

1. An **allow-list of safe base variables** an MCP process may legitimately need — `PATH`, `HOME`/`USERPROFILE`, `TMPDIR`/`TEMP`/`TMP`, locale/terminal vars, and Windows equivalents (`SystemRoot`, `ComSpec`, `PATHEXT`).
2. **Only** the credentials the transport explicitly declares via `transport.env` (resolved through the existing `resolveChildEnv` validation/resolution path).

The full `process.env` is never forwarded. All three stdio spawn sites (`executeStdioMcp`, `executeStdioMcpTool`, `listStdioMcpTools`) now route through `buildChildEnv`.

## Behavior change

If an MCP server previously relied on a variable that is neither allow-listed nor declared in its `transport.env`, it will no longer see it — that variable must now be listed explicitly in the connector's `transport.env`. This is the intended least-privilege behavior.

## Tests

Adds `test/mcp-client.test.ts` (the module had no tests) pinning the invariant:
- OpenWiki credentials don't leak into the child env by key or by value
- allow-listed base vars (e.g. `PATH`) pass through
- declared refs resolve to the mapped target var only
- invalid / unresolvable refs throw

Full gate passes locally: `typecheck`, `lint:check`, `format:check`, and the test suite (200 tests, +5 new).

---

### Related PRs

**Security**

**Correctness**

**Feature**

_(This PR is #287.)_


colifran:
This is a great finding @bikeusaland!

colifran (approved):
This is a great finding and a great fix. Thanks @bikeusaland!

PR #288, opened 2026-07-12
fix: honor per-instance connectorConfig in gmail, slack, x connectors

bikeusaland:
## Problem

`ingestion.ts` passes each source instance's `connectorConfig` overrides to every connector:

```ts
// src/ingestion.ts
connectorConfig: sourceConfig.connectorConfig,
```

But only `hackernews` and `web-search` actually merged those overrides on top of the on-disk config:

```ts
const config = {
  ...(await readConnectorConfig<HackerNewsConfig>("hackernews", { ... })),
  ...((options.connectorConfig ?? {}) as HackerNewsConfig),
};
```

`gmail`, `slack`, and `x` read **only** the on-disk config and ignored `options.connectorConfig` entirely. As a result, per-instance overrides defined during onboarding (e.g. a different `query`, `maxMessages`, or `streams` for a specific source instance) were silently dropped for those three connectors — no error, just the on-disk defaults.

## Fix

Apply the exact merge pattern the other two connectors already use to `gmail`, `slack`, and `x`: read the on-disk config first, then spread `options.connectorConfig` on top so per-instance overrides win.

No behavior change for users who don't set per-instance overrides (the spread of an empty object is a no-op).

## Verification

`typecheck`, `lint:check`, `format:check`, and the full test suite (200 tests) all pass.

Note: the source connectors currently have no unit-test harness (they call `readConnectorConfig` on the real `~/.openwiki` home and global `fetch` with no injection seam), so a direct `ingest()` regression test is out of scope here — it depends on the broader connector-testability work. This change is the same one-line merge pattern already proven by the two connectors that had it.

---

### Related PRs

**Security**

**Correctness**

**Feature**

_(This PR is #288.)_

### DIFF
--- a/src/connectors/sources/gmail.ts
+++ b/src/connectors/sources/gmail.ts
@@ -83,16 +83,19 @@ async function ingest(
   options: ConnectorIngestOptions = {},
 ): Promise<ConnectorIngestResult> {
   const runId = createRunId();
-  const config = await readConnectorConfig<GmailConfig>("google", {
-    enabled: true,
-    format: "full",
-    includeSpamTrash: false,
-    labelIds: [],
-    maxMessages: 100,
-    metadataHeaders: DEFAULT_METADATA_HEADERS,
-    pageSize: 100,
-    query: "newer_than:1d",
-  });
+  const config = {
+    ...(await readConnectorConfig<GmailConfig>("google", {
+      enabled: true,
+      format: "full",
+      includeSpamTrash: false,
+      labelIds: [],
+      maxMessages: 100,
+      metadataHeaders: DEFAULT_METADATA_HEADERS,
+      pageSize: 100,
+      query: "newer_than:1d",
+    })),
+    ...((options.connectorConfig ?? {}) as GmailConfig),
+  };
   const state = await readConnectorState("google");
   const warnings: string[] = [];
   const rawFiles: string[] = [];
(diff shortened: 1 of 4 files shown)


colifran:
@bikeusaland thanks for the PR! Could we get some tests for this?

bikeusaland:
Thanks @colifran! Added tests in `test/connector-config-overrides.test.ts` (commit 5dc8e8b).

Approach: the gmail/slack/x connectors read config from `~/.openwiki/connectors/<id>/config.json` and then call `fetch`, so to keep the tests network-free I point `$HOME` at a throwaway temp dir (the same pattern as the existing `onboarding`/`env-behavior` tests), write an on-disk config, and observe which short-circuit gate the run stops at:

- **x / slack** (default disabled): on-disk `enabled: false` + `ingest({ connectorConfig: { enabled: true } })` → the merge lets it pass the enabled gate and stop at the missing-token gate (`status: "error"`). Without the fix it stays `"skipped"`.
- **gmail** (default enabled): on-disk `enabled: true` + `ingest({ connectorConfig: { enabled: false } })` → the override disables the run (`status: "skipped"`) before any Gmail API call.

I confirmed the x/slack tests fail against the pre-fix code (returning `"skipped"` instead of `"error"`), so they genuinely guard the regression. Full gate passes locally: `typecheck`, `lint:check`, `format:check`, and the suite (200 tests).

colifran (approved):
Nice one. Thanks @bikeusaland!

PR #289, opened 2026-07-12
feat: warn when configured model belongs to a different provider

bikeusaland:
## Problem

`isValidModelId` only validates the *shape* of the model string:

```ts
export function isValidModelId(value: string): boolean {
  const modelId = normalizeModelId(value);
  return (
    modelId.length > 0 &&
    modelId.length <= 120 &&
    /^[A-Za-z0-9][A-Za-z0-9._:/+-]*$/u.test(modelId) &&
    !modelId.includes("://")
  );
}
```

So a saved `OPENWIKI_MODEL_ID` left over from a previous provider — e.g. `claude-opus-4-8` while the provider is now `openai` — passes validation and gets sent to the wrong backend, failing at request time with an opaque provider-side 400/404. Provider/model mismatch is the most common misconfiguration, and today it produces a cryptic error one turn later instead of an actionable local message.

## Fix

Add two pure helpers to `constants.ts`:

- `getProvidersForKnownModelId(modelId, excludeProvider)` — providers (other than the one given) whose known `modelOptions` include the model by **exact** match.
- `isModelIdForOtherProvider(modelId, provider)` — true when the model is a known model of some *other* provider and *not* known for the configured provider.

`resolveModelId` now emits a **non-fatal, actionable warning** on mismatch (via the existing `onEvent` text channel) and still proceeds — a custom endpoint or gateway may legitimately serve the model.

### Why exact matching (not substring)

Provider model lists overlap heavily via namespacing:
- Anthropic lists `claude-opus-4-8`; OpenRouter lists `anthropic/claude-opus-4-8`.
- nvidia lists `openai/gpt-oss-120b`.

Exact-match on the full normalized ID avoids false positives from these overlaps, and returns empty for custom/gateway IDs so `openai-compatible` model names are never flagged. Because `openai` and `openai-chatgpt` share `OPENAI_MODEL_OPTIONS`, switching between those two never warns.

## Tests

Adds unit tests for both helpers covering: exact match, the `excludeProvider` behavior, namespaced-overlap non-matching, shared `openai`/`openai-chatgpt` models, custom/unknown IDs, and whitespace trimming.

Full gate passes: `typecheck`, `lint:check`, `format:check`, test suite (204 tests, +8 new).

---

### Related PRs

**Security**

**Correctness**

**Feature**

_(This PR is #289.)_


bikeusaland:
## Merged `main` — resolved a semantic conflict beyond the textual ones

Rebased/merged latest `main` into this branch. The three textual conflicts (import lists in `src/agent/index.ts` + `test/constants.test.ts`, and the `warnOnProviderModelMismatch` function landing next to `main`'s `export function createModel` change) were straightforward unions.

There was also a **semantic conflict** that the merge tool couldn't flag, worth calling out:

`main` introduced a new **`gemini-enterprise`** provider whose `modelOptions` legitimately include `claude-opus-4-8` — it's a gateway that also serves Claude models. That means `claude-opus-4-8` is now a *known* model of **two** providers (`anthropic` and `gemini-enterprise`), which is exactly the case `getProvidersForKnownModelId` is designed to surface.

Two of this PR's tests were written before `gemini-enterprise` existed and asserted the model belonged to `anthropic` only:

- `finds the provider(s) whose known models include the id` — now expects `["anthropic", "gemini-enterprise"]`
- `excludes the provider passed in` — excluding `anthropic` now leaves `["gemini-enterprise"]` instead of `[]`

I updated both expectations (with a comment explaining why) to reflect the new reality rather than mask it. The feature's behavior is correct here: a `claude-opus-4-8` configured under, say, `openai` still warns and correctly names both providers it belongs to.

No production logic changed in the merge — exact-match semantics are unchanged; the helper simply now finds a second legitimate owner.

Full gate is green after the merge: `typecheck`, `lint:check`, `format:check`, and the full test suite (341 passing).

colifran (approved):
Approving, thanks @bikeusaland ! Detection logic looks great.
 
I pushed one small commit to your branch. The warning also writes to stderr now. Testing showed it was getting swallowed on a failed run which is exactly when you would want to see this, --print buffers streamed text and drops it on error and the interactive UI re-renders it away. Stderr survives that so it's reliable in CI now.

I'm seeing that this is still transient in the interactive TUI since Ink redraws over it but that's a broader "streamed output vanishes on failure" thing I'll track separately.

PR #290, opened 2026-07-12
chore: unify secret-redaction key matching in one shared helper

bikeusaland:
## Problem

Three independent secret-redaction implementations had drifted apart, each matching a **different** set of secret-bearing key terms:

| Location | Terms matched |
|---|---|
| `cli.tsx` `isSecretLikeKey` | `api-key`, `authorization`, `bearer`, `token`, `secret`, `password` |
| `agent/index.ts` `sanitizeOpenRouterResponseBody` (inline) | `api-key`, `authorization`, `bearer`, `password`, `secret`, `token`, **`user_id`** |
| `mcp-runtime.ts` `isSecretLikeKey` | `token`, `secret`, `password`, `authorization`, `api-key`, **`cookie`** |

Because the lists differed, a value redacted on one path could **leak through another** — e.g. a `user_id` field is redacted in an OpenRouter response body but not in MCP tool output, and a `cookie` field is redacted in MCP output but not in the CLI diagnostics path. Redaction is a security boundary; whether a credential gets hidden should not depend on which code path happens to handle it.

## Fix

Consolidate into a single exported `isSecretLikeKey` plus a shared `SECRET_KEY_PATTERN_SOURCE` in `diagnostics.ts` — already the documented, well-tested redaction module (`test/redaction.test.ts`). The unified term list is the **union** of all three:

```
api[-_]?key | authorization | bearer | token | secret | password | user_id | cookie
```

All three consumers now route through it:
- `cli.tsx` — imports the shared predicate, local copy removed.
- `connectors/mcp-runtime.ts` — imports the shared predicate, local copy removed.
- `agent/index.ts` `sanitizeOpenRouterResponseBody` — builds its JSON-key regex from `SECRET_KEY_PATTERN_SOURCE`, so the body sanitizer and the key predicate can never drift again.

Net effect: every path now redacts the strict superset of what any path did before — no key that was previously hidden is now exposed.

## Tests

- `isSecretLikeKey` catches every term (including `user_id` and `cookie`, previously matched by only one path each), case-insensitively, and does not flag benign keys.
- `sanitizeOpenRouterResponseBody` redacts `user_id`/`cookie` values while preserving non-secret fields.

Full gate passes: `typecheck`, `lint:check`, `format:check`, test suite (213 tests, +9 new).

## Scope note

This PR intentionally unifies only the secret-key predicate. The related `isRecord` duplication (and the array-vs-record nuance in `formatToolArgs`) touches ~40 stream-parsing call sites and is left for a separate change to keep this security fix small and low-risk.

---

### Related PRs

**Security**

**Correctness**

**Feature**

_(This PR is #290.)_

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -16,6 +16,7 @@ import {
   saveOpenWikiEnv,
 } from "../env.js";
 import { isFileNotFoundError } from "../fs-errors.js";
+import { SECRET_KEY_PATTERN_SOURCE } from "../diagnostics.js";
 import { openWikiLocalWikiDir } from "../openwiki-home.js";
 import { OpenWikiLocalShellBackend } from "./docs-only-backend.js";
 import {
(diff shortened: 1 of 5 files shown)


colifran (approved):
Nice @bikeusaland! This was on my radar as well. Thanks for the PR!

PR #292, opened 2026-07-12
fix: return a readable error for non-image binary file reads

### LINKED ISSUE #284: [Bug]: `read_file` tool crashes Anthropic/OpenRouter with `unknown variant 'file'` for unrecognized file extensions
### Description
When the agent attempts to read a file with an unrecognized extension (e.g., `.jl` for Julia, or extensionless files like `Makefile`), the tool execution crashes with a `400 Bad Request` from the Anthropic/OpenRouter API. Furthermore, because the conversation state is checkpointed, this permanently bricks the current session/thread.

### Steps to Reproduce
1. Start an agent in a directory containing a file with an unmapped extension (e.g., `script.jl` or `Makefile`).
2. Ask the agent to read the file.
3. The agent crashes with the `unknown variant 'file'` error.
4. Subsequent attempts to run the agent in the same project fail immediately.

### Expected Behavior

The program should work normally.

### Error Message
```text
400 Failed to deserialize the JSON body into the target type: messages[10]: unknown variant `file`, expected `text` at line 1 column 70902
```

### Root Cause
1. **MIME Type Fallback:** In `deepagents`'s `langsmith` distribution, the `getMimeType` function uses a hardcoded dictionary (`MIME_TYPES`) to determine the MIME type of a file. If the extension is not in the dictionary (like `.jl`), or if the file has no extension (like `Makefile`), it defaults to `"application/octet-stream"`.
2. **Binary Encoding:** When the `read_file` tool sees an `application/octet-stream` MIME type, it treats the file as binary. It encodes the file's contents in base64 and returns a tool result block formatted as:
   ```json
   {
     "type": "file",
     "mimeType": "application/octet-stream",
     "data": "<base64_data>"
   }
   ```
3. **API Rejection:** The Anthropic Messages API does not support the `type: "file"` content block (it expects `text`, `image`, or `document`). When the LLM adapter forwards this payload, the API throws a 400 error.
4. **Permanent State Corruption:** Because `openwiki` uses LangGraph checkpointing (SQLite), this "poisoned" tool result is saved to the local database. Subsequent runs of the agent attempt to resume the thread, re-sending the invalid message and immediately crashing again.

### Environment

- OS:
- Node.js version:
- OpenWiki version:

### Proposed Solution(s)

**1. Change the Default Fallback to Text (Recommended)**
Since documentation agents primarily operate on source code, unknown extensions are overwhelmingly likely to be text files. Changing the fallback in `getMimeType` from `"application/octet-stream"` to `"text/plain"` resolves the issue broadly for all programming languages and extensionless config files.

```javascript
// Before

// After
return MIME_TYPES[extname(filePath).toLocaleLowerCase()] || "text/plain";
```
*(If a true binary file is accidentally read, JavaScript will decode it as a garbled string which is safely passed as `text` rather than crashing the API).*

**2. Expand the Hardcoded MIME Dictionary**
Alternatively, expand the hardcoded `MIME_TYPES` mapping to cover missing languages (e.g., `.jl`, `.kt`, `.swift`, `.sh`, etc.) and map them to `"text/plain"`. 

### Temporary Workaround for Users
1. **Clear the corrupted thread:** You must delete the poisoned thread from your `~/.openwiki/openwiki.sqlite` database, otherwise the agent will never recover for that directory.
2. **Patch locally:** Open `~/.npm-global/lib/node_modules/openwiki/node_modules/deepagents/dist/langsmith-*.js` (or `.cjs`) and replace `|| "application/octet-stream"` with `|| "text/plain"`.

Bortlesboat:
Fixes #284

## What

Reading a file whose MIME type is neither text nor image makes the `read_file` tool emit a `{type: "file"}` content block in the tool result, which Anthropic/OpenRouter reject with `400 ... unknown variant 'file'`. Because the run is checkpointed, the poisoned message replays on every retry and the thread never recovers — the failure mode reported in #284.

This overrides `read()` in `OpenWikiLocalShellBackend` (same pattern as the existing `write`/`edit` overrides): when the underlying read returns binary content that isn't an image, return a `ReadResult` error instead of bytes. The deepagents tool layer turns backend errors into plain text blocks, so the agent gets actionable feedback ("skip this file, describe it from context") and the request stays valid. Image reads pass through untouched since image blocks are accepted in tool results, and text reads are completely unaffected.

## Notes from reproducing

- With the currently pinned deepagents (1.10.7), `.jl` and `Makefile` from the issue actually resolve to `text/plain` (unknown extensions fall back to text), so they read fine at HEAD — I verified this against a scratch repo. The still-crashing class today is extensions mapped to non-text types (`.pdf`, `.ppt`, `.pptx`), and any older bundled deepagents where the fallback differed. Guarding at the backend covers both.
- This prevents new threads from getting poisoned; a thread already bricked by a checkpointed `file` block still needs its thread/checkpoint deleted. Happy to look at that separately if wanted, but it seemed out of scope for one PR.

## Testing

Three new tests in `test/docs-only-backend.test.ts`: text-with-unknown-extension reads (`.jl`, `Makefile`), PDF read returns the error, PNG read passes through. `pnpm run format`, `pnpm run lint`, and `pnpm test` pass locally (the one `env-behavior` failure is the 0600-permissions test, which also fails on a clean checkout on Windows).

### DIFF
--- a/src/agent/docs-only-backend.ts
+++ b/src/agent/docs-only-backend.ts
@@ -2,6 +2,7 @@ import {
   LocalShellBackend,
   type EditResult,
   type LocalShellBackendOptions,
+  type ReadResult,
   type WriteResult,
 } from "deepagents";
 import { OPEN_WIKI_DIR } from "../constants.js";
@@ -20,6 +21,34 @@ export class OpenWikiLocalShellBackend extends LocalShellBackend {
     super(options);
     this.docsOnly = options.docsOnly === true;
     this.outputMode = options.outputMode ?? "repository";
+  }
+
+  override async read(
+    filePath: string,
+    offset?: number,
+    limit?: number,
+  ): Promise<ReadResult> {
+    const result = await super.read(filePath, offset, limit);
+
+    if (result.error !== undefined || typeof result.content === "string") {
+      return result;
+    }
+
+    // Binary reads that are not images become `file`/`audio`/`video` content
+    // blocks in the tool result, which Anthropic and OpenRouter reject with a
+    // 400 — and the checkpointed thread then replays the poisoned message on
+    // every retry. Return a readable error so the agent can move on instead.
+    const mimeType = result.mimeType ?? "application/octet-stream";
+    if (mimeType.startsWith("image/")) {
+      return result;
+    }
+
+    return {
+      error:
+        `Cannot read '${filePath}': binary content (${mimeType}) is not ` +
+        "supported. Skip this file and describe it from its path and " +
+        "surrounding context instead.",
+    };
   }
 
   override async write(
--- a/test/docs-only-backend.test.ts
+++ b/test/docs-only-backend.test.ts
@@ -1,4 +1,4 @@
-import { mkdtemp, readFile } from "node:fs/promises";
+import { mkdtemp, readFile, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { describe, expect, test } from "vitest";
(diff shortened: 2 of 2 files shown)


colifran:
Closing since this was resolved in an upstream deepagentsjs fix: https://github.com/langchain-ai/openwiki/issues/284#issuecomment-4958821634

PR #293, opened 2026-07-12
feat: add claude-code and codex-cli headless CLI providers

csy1204:
## What

Adds two providers — `claude-code` and `codex-cli` — that delegate the whole agent loop to locally installed headless CLIs (`claude -p --output-format stream-json` / `codex exec --json`) instead of API-backed deepagents:

- `src/constants.ts`: the two providers with `authMethod: "cli"` (no API key; the CLI's own login is the auth), model options (claude: `sonnet`/`opus`/`haiku` aliases; codex: the OpenAI model list), and a run-timeout env (`OPENWIKI_CLI_TIMEOUT_SECONDS`, default 1800s).
- `src/agent/cli-runner/`: engine adapters (argv builders + stream-event parsers mapped onto the existing `OpenWikiRunEvent` union, so the TUI is unchanged), a spawn/stream runner (stderr tail on failure, SIGTERM→SIGKILL timeout, auth-failure login hints, EPIPE-safe stdin), CLI-specific prompt builders (real relative paths instead of the virtual-root wording; no LangChain tool names), and a `threadId ↔ CLI session` store (`~/.openwiki/cli-sessions.json`) so chat follow-ups resume the CLI session (with fresh-session fallback when resume fails).
- `runOpenWikiAgent` routes CLI providers to the runner before any API-key checks; the update-noop gate, content snapshot, and `.last-update.json` metadata logic are shared with the API path (metadata records `claude-code/sonnet`-style model ids).
- Onboarding wizard lists both providers and skips the API-key step for them; the non-interactive startup gate exempts CLI-auth providers from the key requirement (empty-message validation still applies).
- Safety: claude runs with `--permission-mode acceptEdits` plus an allowlist (read tools, read-only git, exact plan-file cleanup commands — no network tools, no broad Bash); codex runs with `--sandbox workspace-write`. Since headless permission models cannot express "read repo-wide, write wiki-only", repository-mode runs also do a warn-only post-run `git status` check (baselined against the pre-run dirty state) that reports any out-of-wiki changes.

## Why

Bootstrapping repository docs (`openwiki code --init`/`--update`) is tool-call heavy, which makes the API path slow and expensive. Claude Code and Codex subscriptions already include agentic execution with native file/search/shell tools, so delegating the loop to them makes the initial analysis faster and removes per-token API cost. Existing API providers are untouched; the CLI providers are opt-in via `OPENWIKI_PROVIDER` only (never auto-selected by the key-based fallback chain).

## How I tested it

- **Unit/integration (Vitest): 251 passing** (234 pre-existing — all unchanged — plus 17 new across adapters, prompt builders, session store, runner, routing, startup gate, write guard). Parser fixtures are captured from real CLI runs (claude 2.1.207, codex 0.144.1). The runner is tested against stub CLI executables (real spawn/stdin/stream/exit paths, including a regression test that reproduced an uncaught `write EPIPE` crash pre-fix).
- **End-to-end against the real CLIs** on a fixture repo: wiki generation with both engines (correct `.last-update.json` model ids), noop-skip on unchanged re-run, chat follow-up resuming the same CLI session (verified conversational carry, no fallback), and the onboarding wizard flow (provider list → model selection, no key prompt).
- `pnpm run format`, `pnpm run lint`, `pnpm test` all clean.

## Scope note

This is one feature (CLI-auth providers) and every hunk serves it, but it is a large surface: provider config, a new `cli-runner` module, prompt-builder extraction (`createSystemPrompt` output is byte-identical — verified before/after across all command×outputMode combinations), wizard, and the startup gate. If you'd prefer it split (e.g. provider scaffolding → runner → wizard/gate), happy to do that — flagging per the contributing guide rather than deciding unilaterally.

### DIFF
--- a/src/agent/cli-runner/claude.ts
+++ b/src/agent/cli-runner/claude.ts
@@ -1,0 +1,180 @@
+import type { CliEngineAdapter, CliParsedEvent, CliRunSpec } from "./types.js";
+
+const CLAUDE_ALLOWED_TOOLS = [
+  "Read",
+  "Glob",
+  "Grep",
+  "LS",
+  "Write",
+  "Edit",
+  "MultiEdit",
+  "TodoWrite",
+  "Task",
+  "Bash(git log:*)",
+  "Bash(git show:*)",
+  "Bash(git diff:*)",
+  "Bash(git status:*)",
+  "Bash(git blame:*)",
+  "Bash(rg:*)",
+  "Bash(ls:*)",
+  // Exact plan-file cleanup commands referenced by the CLI prompt.
+  "Bash(rm -f openwiki/_plan.md)",
+  "Bash(rm -f _plan.md)",
+].join(",");
+
+export const claudeAdapter: CliEngineAdapter = {
+  cliCommand: "claude",
+  engine: "claude-code",
+
+  buildArgs(spec: CliRunSpec): string[] {
+    const args = [
+      "-p",
+      "--output-format",
+      "stream-json",
+      "--verbose",
+      "--model",
+      spec.modelId,
+      "--permission-mode",
+      "acceptEdits",
+      "--append-system-prompt",
+      spec.systemPrompt,
+      "--allowedTools",
+      CLAUDE_ALLOWED_TOOLS,
+    ];
+
+    if (spec.resumeSessionId) {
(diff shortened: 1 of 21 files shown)


colifran:

PR #303, opened 2026-07-13
fix: gate scheduled OpenWiki Update workflow to the origin repository

### LINKED ISSUE #302: Scheduled OpenWiki Update workflow has no fork guard — fails daily on every fork without OPENROUTER_API_KEY
## Problem

I forked this repo to contribute a provider PR.

Enabling Actions to see my own CI — something every contributor naturally does — also armed the scheduled `OpenWiki Update` workflow.

It has failed every day since:

```
OPENAI_API_KEY is required for non-interactive runs. Run openwiki in an interactive terminal to save credentials.
```

`.github/workflows/openwiki-update.yml` has no `if:` condition anywhere.

Any fork that enables Actions gets a daily job that is guaranteed to fail — unless the fork owner happens to know they need to configure a specific secret that nobody told them about (see the companion doc issue I'm about to open).

## Proposal

Guard the job to the origin repository, e.g.:

```yaml
jobs:
  update:
    if: github.repository == 'langchain-ai/openwiki'
```

Since `openwiki code --update` also *generates* this same workflow for user repos (`createCodeModeWorkflow` in `src/code-mode.ts`), it might also be worth considering whether the generated template should default to a similar guard, or a graceful skip, for repos where the required secret isn't set — rather than failing loudly by default.

Happy to send a PR for this repo's own workflow file if useful.

jyje:
Closes #302

## Summary

I forked this repo to contribute a provider PR.

Enabling Actions to see my own CI also armed the scheduled `OpenWiki Update` workflow — which has failed every day since. The workflow assumes an `OPENROUTER_API_KEY` repo secret, but nothing tells a new fork owner to set one up.

**Default (no variable set)**
The job only runs in the origin repository. Forking to contribute no longer silently arms a daily job that's guaranteed to fail.

**Opt-in (`OPENWIKI_ENABLE_SCHEDULED_UPDATE=true`)**
Fork owners who want scheduled updates on their own fork can set this repository variable (Settings → Actions → Variables) instead of editing the workflow file. The choice survives future upstream syncs instead of drifting out of a hand-edited copy. Deleting the variable (or setting it to anything other than `true`) reverts to the default.

## Testing

A schedule trigger can't be exercised on demand, so I verified both branches of the guard live on my fork via `workflow_dispatch`:

**Default (no variable set)**
Run's overall conclusion is `skipped`.
https://github.com/jyje/openwiki/actions/runs/29223553562

**Opt-in**
With `OPENWIKI_ENABLE_SCHEDULED_UPDATE=true` set, checkout / setup-node / install steps all succeed — confirming the guard let the job through. It then fails at the "Run OpenWiki" step because my fork has no `OPENROUTER_API_KEY` configured. That's the same symptom this issue reports, now happening on purpose — proof the opt-in works and nothing else changed.
https://github.com/jyje/openwiki/actions/runs/29223609215

No source files changed, so `pnpm test` / `typecheck` / `build` are unaffected. Ran `pnpm run format:check` to confirm the YAML is still Prettier-clean.

### DIFF
--- a/.github/workflows/openwiki-update.yml
+++ b/.github/workflows/openwiki-update.yml
@@ -11,6 +11,13 @@ permissions:
 
 jobs:
   update:
+    # Scheduled (and manually-dispatched) runs only fire in the origin repo by
+    # default, so forking to contribute doesn't silently arm a daily job that
+    # is guaranteed to fail without a configured provider secret. Fork owners
+    # who want scheduled updates on their own fork can opt in without editing
+    # this file by setting the OPENWIKI_ENABLE_SCHEDULED_UPDATE repository
+    # variable (Settings -> Actions -> Variables) to "true".
+    if: github.repository == 'langchain-ai/openwiki' || vars.OPENWIKI_ENABLE_SCHEDULED_UPDATE == 'true'
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository


colifran (approved):
Nice! Good catch and thanks for the PR @jyje!

PR #305, opened 2026-07-13
chore: update chatgpt login model wording

HwangJohn:
## Summary
- update the ChatGPT-login provider README wording so it does not list stale model IDs
- point readers to the shared OpenAI provider model list instead

## Testing
- `corepack pnpm exec prettier --check README.md`
- `corepack pnpm exec vitest run test/openai-chatgpt-provider.test.ts test/constants.test.ts`


colifran (approved):
Thanks @HwangJohn!

PR #307, opened 2026-07-13
fix: align cron schedule target help

HwangJohn:
## Summary
- make `openwiki cron pause|resume|delete` accept only `all`, matching the schedule mutation implementation
- update help text and examples so source-specific targets are not advertised
- add parser coverage for rejecting a source id

## Testing
- `corepack pnpm exec prettier --check src/commands.ts test/commands.test.ts`
- `corepack pnpm exec vitest run test/commands.test.ts`


colifran (approved):
Thanks - makes sense. Opened https://github.com/langchain-ai/openwiki/issues/399 to track adding actual per-connector schedules.

PR #309, opened 2026-07-13
fix: avoid OAuth callback server address race

### LINKED ISSUE #297: [Bug]: Callback server crashes with Cannot read properties of null (reading 'port') after successful OAuth — and skips connector registration
### Description

Gmail OAuth callback server crashes post-auth (null server.address()), leaving tokens saved but source not registered

### Steps to Reproduce

1. Run openwiki auth gmail
2. Complete the Google consent screen (click Allow)
3. Browser redirects to http://127.0.0.1:53682/callback

### Expected Behavior

 Either the server should ignore/ignore-gracefully-handle requests after the primary callback is processed, or guard the server.address() call with a null check. Additionally, connector source registration (writing to onboarding.json) should complete before the callback server is closed, or be resilient to this crash.

### Actual Behavior

CLI crashes immediately after redirect:
```
file:///C:/Users/.../npm-global/node_modules/openwiki/dist/auth/oauth.js:252
        const requestUrl = new URL(request.url ?? "/", `http://${CALLBACK_HOST}:${server.address().port}`);
                                                                                                  ^
TypeError: Cannot read properties of null (reading 'port')
    at Server.<anonymous> (file:///.../dist/auth/oauth.js:252:99)
    at Server.emit (node:events:518:28)
    at parserOnIncoming (node:_http_server:1150:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:117:17)
```

### Environment

- OS: Windows 11 (PowerShell)
- Node.js version: v22.11.0
- OpenWiki version: v0.1.1 

### Additional Context

**Consequence:**

 OPENWIKI_GMAIL_ACCESS_TOKEN and OPENWIKI_GMAIL_REFRESH_TOKEN ARE correctly written to ~/.openwiki/.env, but the connector is never registered — ~/.openwiki/onboarding.json shows empty sourceInstances: [] and sources: {}. Running openwiki ingest gmail afterward fails with:
No configured ingestion source matched gmail.

**Suspected root cause**: 

Race condition in the local callback HTTP server's shutdown — likely a trailing request (e.g. browser favicon fetch) arrives after server.close() has already nulled server.address(), and the handler at oauth.js:252 doesn't guard against this.

HwangJohn:
## Summary
- stop parsing callback request URLs with `server.address()` inside the request handler
- capture the callback base URL after listen succeeds and reuse it for later requests
- add a regression test for a trailing request after callback server close

Fixes #297.

## Testing
- `corepack pnpm exec prettier --check src/auth/oauth.ts test/oauth-callback.test.ts`
- `corepack pnpm exec vitest run test/oauth-callback.test.ts`

### DIFF
--- a/src/auth/oauth.ts
+++ b/src/auth/oauth.ts
@@ -1,7 +1,6 @@
 import { createHash, randomBytes } from "node:crypto";
 import { execFile } from "node:child_process";
 import http from "node:http";
-import { AddressInfo } from "node:net";
 import { loadOpenWikiEnv, saveOpenWikiEnv } from "../env.js";
 import { getAuthProvider } from "./providers.js";
 import type {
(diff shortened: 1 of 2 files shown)


HwangJohn:
Closing this as superseded by #311 / 4187eb0, which fixes the same OAuth callback shutdown race and adds a real callback-server regression test on main. Keeping this PR open would now conflict with that implementation.

PR #312, opened 2026-07-13
Opt-in OpenRouter hardening + subagent kill switch (redo of #50 against current main)

blue-az:
## Summary

This supersedes #50. That PR hardened the OpenRouter path against `main` as of early July; `main` has since landed ChatGPT/Codex OAuth login, the connector system, output modes, and a rewritten `docs-only-backend`, all of which touched every function #50 modified (`createModel`, the `createDeepAgent` call site, `createModeInstructions`). Rather than force a conflict resolution across that much drift, this re-implements the same three opt-in controls directly against current `main`.

All three are strictly opt-in — with nothing set, behavior is unchanged for every provider:

- **`OPENWIKI_OPENROUTER_MAX_INPUT_TOKENS`** — attaches a `maxInputTokens` profile to OpenRouter model instances (`Object.defineProperty`, so a frozen/non-extensible instance doesn't crash), so DeepAgents' summarization middleware triggers early enough to stay under OpenRouter's payload limit on constrained/budget models, instead of failing with an HTTP 500 on large transcripts. Suggested value for constrained models: `15000`.
- **`OPENWIKI_DISABLE_MODEL_FALLBACK=1`** — disables OpenRouter's `route: "fallback"` model list for deterministic single-model runs.
- **`OPENWIKI_DISABLE_SUBAGENTS=1`** — excludes the `task` tool from the model, enforced in two places that share one tool-name constant: a registered DeepAgents harness profile (`excludedTools`, `generalPurposeSubagent.enabled: false`) and runtime middleware (`wrapModelCall` filters `task` out of the tool list). The subagent-discipline prompt text is swapped to match, so the system prompt doesn't tell the model it can delegate when the tool has been removed.

Also: `openwiki.*` variant directories (testing/validation copies of generated docs) are excluded from repository discovery in the prompt, so they're never inspected or documented as part of a real run.

## Why bundle these three

They're independent switches but came from the same investigation (hardening a constrained-model OpenRouter run) and the original review thread on #50 already covered this scope; splitting them into separate PRs now would just re-litigate settled ground. Happy to split on request.

## Test plan

- [x] `pnpm run typecheck` — clean
- [x] `pnpm run lint:check` — clean
- [x] `pnpm run format:check` — clean
- [x] `pnpm run build` — clean
- [x] `pnpm test` — 213/213 passing across 20 files, including two new files:
  - `test/subagent-disable.test.ts` — `isSubagentDisabled` only trips on the literal `"1"`; `buildSubagentDisabledProfile()`'s shape; default behavior unchanged (no middleware, no tools removed) when unset; the `task` tool is filtered from the model call when disabled, all other tools pass through unchanged.
  - `test/openrouter-hardening.test.ts` — `isModelFallbackDisabled` only trips on `"1"`; non-OpenRouter providers never get a fallback list; OpenRouter gets the primary model plus fallback ids by default, without duplicating the primary when it's also a fallback id; disabled collapses to a single model; `resolveOpenRouterMaxInputTokens` parsing (unset → undefined, non-numeric → undefined, valid → parsed).
  - Extended `test/prompt.test.ts` for the `openwiki.*` exclusion bullet and both branches of the subagent-discipline text.

### DIFF
--- a/openwiki/operations/credentials-and-updates.md
+++ b/openwiki/operations/credentials-and-updates.md
@@ -39,6 +39,10 @@ The file stores provider configuration and API keys:
 - Optional OAuth callback settings: `OPENWIKI_OAUTH_CALLBACK_PORT` controls the
   local callback port, and `OPENWIKI_HTTPS_OAUTH_REDIRECT_URI` stores the
   Slack-only HTTPS callback URL created by `openwiki ngrok start`.
+- Run-hardening options (all opt-in; unset, behavior is unchanged):
+  - `OPENWIKI_OPENROUTER_MAX_INPUT_TOKENS` — sets a `maxInputTokens` profile on OpenRouter models so DeepAgents' summarization middleware triggers early enough to stay under OpenRouter's payload limit on constrained/budget models, instead of failing with an HTTP 500 on large transcripts. Suggested value for constrained models: `15000`.
+  - `OPENWIKI_DISABLE_MODEL_FALLBACK` — set to `1` to disable OpenRouter's `route: "fallback"` model list for deterministic single-model runs.
+  - `OPENWIKI_DISABLE_SUBAGENTS` — set to `1` to disable subagent task delegation entirely, for runs where transcript growth from subagent fan-out is the binding constraint.
 
 The loader merges those values into `process.env`, while preferring existing process-level values over file values. Deprecated keys (`OPENAI_BASE_URL`, `OPENAI_ORG_ID`, `OPENAI_PROJECT`) are skipped on load and removed on save.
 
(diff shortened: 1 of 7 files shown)


blue-az:
@rajpratham1 tagging you since you reviewed the original version of this on #50 — this is the same hardening (OpenRouter token-cap, model-fallback disable, subagent kill switch), redone against current main since #50's base had drifted too far to cleanly rebase. Test coverage for the subagent-disable path (your original ask) carried over, plus new coverage for the fallback-route and token-cap logic.

colifran:
Thanks a lot for this. The cross-vendor OpenRouter fallback here runs counter to a decision we already merged in #201, which deliberately pins routing rather than falling back across vendors, so taking this would reverse that. I'm going to close it to keep the queue current. If you're hitting a concrete problem that #201 doesn't cover, a fresh issue with the repro is very welcome, and thank you for digging in.

PR #318, opened 2026-07-13
fix: allow leading '@' in model ids for cloudflare workers ai

### LINKED ISSUE #315: [Bug]: isValidModelId rejects all Cloudflare Workers AI model IDs (leading "@" in "@cf/...")
### Description

The `openai-compatible` provider can be pointed at Cloudflare Workers AI's OpenAI-compatible endpoint (`https://api.cloudflare.com/client/v4/accounts/{account_id}/ai/v1`). All Workers AI model IDs start with `@cf/` (e.g. `@cf/moonshotai/kimi-k2.7-code`), and the endpoint only accepts that exact form.

However, `isValidModelId` in `dist/constants.js` requires the first character to be alphanumeric:

```js
/^[A-Za-z0-9][A-Za-z0-9._:/+-]*$/u
```

so every Workers AI model ID is rejected before any request is made. The CLI forbids the leading `@` while the endpoint requires it, which makes Workers AI unusable as an `openai-compatible` backend.

### Steps to Reproduce

1. `export OPENWIKI_PROVIDER=openai-compatible`
2. `export OPENAI_COMPATIBLE_BASE_URL=https://api.cloudflare.com/client/v4/accounts/<account_id>/ai/v1`
3. `export OPENAI_COMPATIBLE_API_KEY=<workers-ai-token>`
4. `export OPENWIKI_MODEL_ID=@cf/moonshotai/kimi-k2.7-code`
5. `openwiki code --update --print`

### Expected Behavior

The run proceeds — the model ID is valid for the configured endpoint.

### Actual Behavior

The CLI exits immediately, before any API call:

```
Invalid model ID configured in OPENWIKI_MODEL_ID.
```

### Environment

- OS: Debian 12 (official `node:26` Docker image)
- Node.js version: 26
- OpenWiki version: 0.1.1

### Additional Context

**Root cause:** `isValidModelId` (`dist/constants.js`) anchors the pattern to an alphanumeric first character, so the leading `@` fails validation before any request is made.

**Endpoint evidence (the prefix cannot be dropped on the Cloudflare side):** calling the endpoint with `cf/moonshotai/kimi-k2.7-code` returns `AiError: No such model` (code 5007); with `moonshotai/kimi-k2.7-code`, `Model not found` (code 7003).

**Tested fix:** allowing a leading `@` (`/^[@A-Za-z0-9][A-Za-z0-9._:/+-]*$/u`) was verified by patching the installed package; a full `openwiki code --update` run then completed successfully against Workers AI (21 requests, wiki generated), so the one-character change appears sufficient. A pull request with this change can be opened if the maintainers agree with the approach.

**Workaround (until fixed):** patch the installed package after install, e.g. `sed -i 's#\^\[A-Za-z0-9\]#^[@A-Za-z0-9]#' <global node_modules>/openwiki/dist/constants.js` — verified working, but a fragile stopgap that should be removed once a fixed version ships.

himanshu231204:
## Summary

The isValidModelId regex in src/constants.ts requires the first character to be alphanumeric, which rejects all Cloudflare Workers AI model IDs (e.g. @cf/meta/llama-3-8b-instruct). This makes Workers AI unusable as an openai-compatible backend.

## Fix

One-character change to the regex — add @ to the allowed first character set:

`diff
- /^[A-Za-z0-9][A-Za-z0-9._:/+-]*$/u
+ /^[@A-Za-z0-9][A-Za-z0-9._:/+-]*$/u
`

## Changes

- src/constants.ts: Updated regex to accept @ as a valid first character
- 	est/constants.test.ts: Added test cases for Cloudflare Workers AI model IDs

## Testing

All 26 tests pass (including 3 new ones for @cf/... model IDs).

Fixes #315

cc @colifran

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -655,7 +655,9 @@ export function isValidModelId(value: string): boolean {
   return (
     modelId.length > 0 &&
     modelId.length <= 120 &&
-    /^[A-Za-z0-9][A-Za-z0-9._:/@+-]*$/u.test(modelId) &&
+    // Leading @ for Cloudflare Workers AI ids (@cf/...); interior @ for
+    // Vertex AI @-versioned ids (e.g. claude-sonnet-4-5@20250929).
+    /^[@A-Za-z0-9][A-Za-z0-9._:/@+-]*$/u.test(modelId) &&
     !modelId.includes("://")
   );
 }
--- a/test/constants.test.ts
+++ b/test/constants.test.ts
@@ -38,6 +38,12 @@ describe("isValidModelId", () => {
     expect(isValidModelId("nvidia/nemotron-3-super-120b-a12b")).toBe(true);
   });
 
+  test("accepts Cloudflare Workers AI model ids with leading '@'", () => {
+    expect(isValidModelId("@cf/meta/llama-3-8b-instruct")).toBe(true);
+    expect(isValidModelId("@cf/moonshotai/kimi-k2.7-code")).toBe(true);
+    expect(isValidModelId("@cf/qwen/qwen1.5-14b-chat-awq")).toBe(true);
+  });
+
   test("rejects empty, whitespace-only, and over-long ids", () => {
     expect(isValidModelId("")).toBe(false);
     expect(isValidModelId("   ")).toBe(false);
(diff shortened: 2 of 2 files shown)


colifran (approved):
Great! Resolved merge conflicts on your branch. Ready to merge. Thanks for the PR!

PR #323, opened 2026-07-13
feat: OKF index.md files

bracesproul:


src/agent/prompt.ts
```diff
@@ -89,6 +89,16 @@ Planning discipline:
 - Before completing the run, delete ${output.planPath}. If there is no filesystem delete tool, use shell execute from the runtime root, for example ${output.removePlanCommand}.
 - Do not leave ${output.planPath} in the final wiki.
 
+Index finalization discipline:
+- OpenWiki tracks successful Markdown writes and edits during this run and exposes two tools for maintaining directory indexes. Do not write or edit index.md files manually.
+- \`openwiki_get_pending_indexes\` returns only index.md files that still need to be created or updated. Each result includes the index path, whether it exists, and its current description when one is set.
```
colifran: why tools here? couldn't this be done deterministically? shouldn't we make these behaviors a default non-optional part of the code?

src/agent/prompt.ts
```diff
@@ -89,6 +89,16 @@ Planning discipline:
...
+- \`openwiki_create_or_update_index\` accepts an index path and an optional one-sentence directory description. It deterministically lists the directory's current Markdown files and subdirectories, reads their titles and descriptions from OKF front matter, writes the index, and updates the immediate parent index when that parent already exists.
+- A new index, or an existing index without a description, requires you to supply a concise one-sentence description based on the directory's purpose and contents. Omit the description argument when the existing description remains accurate. Supply a replacement only when the directory's purpose has materially changed.
+- Wait until you are 100% finished with every non-index wiki write and edit in this run before starting index finalization. Index generation must be the final wiki-writing phase.
```
colifran: same here, feels like we should be the ones doing this

src/agent/indexing/create-or-update-index-tool.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const parentPath = getParentIndexPath(indexPath, outputMode);
+  if (parentPath && (await readTextIfExists(backend, parentPath))) {
+    await synchronize(backend, outputMode, parentPath, undefined, updated);
+  }
+  return { updated };
+}
+
+async function synchronize(
+  backend: IndexBackend,
```
colifran: why do we need an index backend? it looks like it is only selecting some of the methods on the primary backend but if we're writing the code for call edit and write then why does it need to be stripped down?

src/agent/indexing/get-pending-indexes-tool.ts
```diff
@@ -0,0 +1,74 @@
...
+          outputMode,
+          runtime.state.editedWikiPaths,
+        ),
+        null,
+        2,
+      ),
+    {
+      name: "openwiki_get_pending_indexes",
+      description:
+        "Return only index.md files that need creation or changes after this run's wiki writes/edits, including path, existence, and current description when set.",
```
colifran: seems pretty cheap to just always create or update index.md files even if we're recreating each time. latency isn't a critical concern imo. why not just make this a forced thing rather than introduce logic where we're checking old vs. new

src/agent/indexing/utils.ts
```diff
@@ -0,0 +1,268 @@
...
+  "INSTRUCTIONS.md",
+  INDEX_FILE_NAME,
+]);
+
+export type IndexBackend = Pick<
+  BackendProtocolV2,
+  "edit" | "ls" | "readRaw" | "write"
+>;
+
+export type IndexMetadata = {
```
colifran: nit: these types should be interfaces

colifran:
Why couldn't this whole thing be collapsed to:
1. Agent tries to generate OKF conformant wiki using prompting
2. Tree walk to validate frontmatter, yaml.parse, if fail fall back to deterministic frontmatter generation
3. Deterministically generate index.md everytime - regenerating is cheap, this isn't a design where latency is critical

src/agent/prompt.ts
```diff
@@ -139,7 +142,7 @@ Documentation goals:
 
 Front matter requirements (OKF):
 - Every Markdown file you create or update under ${output.docsLocation}, including the temporary ${output.planPath} file, MUST begin with OKF-compliant YAML front matter.
-- The front matter MUST follow the Google Knowledge Catalog OKF schema
+- The front matter MUST follow the Google Knowledge Catalog OKF schema: https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md
```
bracesproul:
```suggestion
- The front matter MUST follow the Google Knowledge Catalog OKF schema (see below)
```

src/agent/prompt.ts
bracesproul: stale

src/agent/prompt.ts
bracesproul: stale

src/agent/indexing/create-or-update-index-tool.ts
bracesproul: stale

src/agent/indexing/get-pending-indexes-tool.ts
bracesproul: stale

src/agent/indexing/utils.ts
bracesproul: was able to update two, but one needs to remain a type since it's just picking fields from another type and not adding any extra, so cant make that an interface

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+
+const INDEX_FILE = "index.md";
+const EXCLUDED_FILES = new Set([INDEX_FILE, "_plan.md", "INSTRUCTIONS.md"]);
+
+type IndexBackend = Pick<
+  BackendProtocolV2,
+  "edit" | "ls" | "readRaw" | "write"
+>;
+
+interface Directory {
```
colifran, line 10: Nice. Thanks for making these interfaces!

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+}
+
+/** Creates middleware that synchronizes deterministic wiki indexes after a run. */
+export function createOpenWikiIndexMiddleware(
+  backend: IndexBackend,
+  outputMode: OpenWikiOutputMode,
+) {
+  return createMiddleware({
+    name: "OpenWikiIndexMiddleware",
+    afterAgent: async () => {
```
colifran, line 27: good call - this is smart

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
+import type { BackendProtocolV2, FileInfo } from "deepagents";
+import { createMiddleware } from "langchain";
+import path from "node:path";
+import { parse } from "yaml";
+import type { OpenWikiOutputMode } from "./types.js";
+
+const INDEX_FILE = "index.md";
+const EXCLUDED_FILES = new Set([INDEX_FILE, "_plan.md", "INSTRUCTIONS.md"]);
+
+type IndexBackend = Pick<
```
colifran: Q: Why a subset of BackendProtocolV2? Is it failing the type checker or something?

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+  for (const directory of await collectDirectories(backend, root, true)) {
+    await synchronizeDirectory(backend, directory, root);
+  }
+}
+
+/** Recursively collects visible wiki directories and their entries. */
+async function collectDirectories(
+  backend: IndexBackend,
+  directoryPath: string,
+  allowMissing = false,
```
colifran, line 48: Q: what does this do? not blocking but if our only call site passes true should we just drop this and always allow missing?

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+  return `# ${heading}\n\n${items.join("\n")}`;
+}
+
+/** Parses the title and required description from YAML front matter. */
+function parseFrontmatter(
+  content: string,
+  filePath: string,
+): { description: string; title?: string } {
+  const block = /^---\r?\n([\s\S]*?)\r?\n---(?:\r?\n|$)/u.exec(content)?.[1];
+  if (!block) throw new Error(`${filePath} lacks YAML front matter.`);
```
colifran, line 164: Q: what happens in this case? does it crash? bubble up to the agent?

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  let fields: unknown;
+  try {
+    fields = parse(block, {
```
colifran, line 168: nice - clean

src/agent/index-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+    });
+  }
+
+  const indexPath = path.posix.join(directory.path, INDEX_FILE);
+  const title =
+    directory.path === root
+      ? "OpenWiki"
+      : titleFromSlug(path.posix.basename(directory.path));
+  const content = renderIndex(title, files, directories);
+  const existing = directory.entries.some(
```
colifran, line 113: i would almost even say at this point we've already generated the new index why not just write every time instead of doing a second read for existing and comparing.

colifran (approved):
Super, super clean. Really nice.

src/agent/index-middleware.ts
bracesproul: good catch this is ai verbosity

src/agent/index-middleware.ts
bracesproul: we pass both, true on L39 and false (not passed, uses default) on L62

src/agent/index-middleware.ts
bracesproul: kills the agent

src/agent/index-middleware.ts
bracesproul: write() fails for existing files, while edit() requires the existing contents, so we need this read. Since we already have the contents, the comparison prevents unnecessary edits.

PR #324, opened 2026-07-14
feat: Add OKF front matter validator

bracesproul:


src/agent/indexing/frontmatter-validator.ts
```diff
@@ -0,0 +1,226 @@
...
+  line?: number;
+  message: string;
+};
+
+export type FrontmatterValidation =
+  { valid: true } | { issues: FrontmatterIssue[]; valid: false };
+
+type ReadBackend = Pick<BackendProtocolV2, "readRaw">;
+
+export function validateOkfFrontmatter(content: string): FrontmatterValidation {
```
colifran: all of this handle rolled yaml parsing seems super heavy and error prone for something we could probably just validate with yaml.parse or fallback to a deterministic frontmatter generation

src/agent/indexing/state.ts
```diff
@@ -31,13 +32,24 @@ export const OpenWikiIndexStateSchema = new StateSchema({
...
+) {
   return createMiddleware({
     name: "OpenWikiIndexState",
     stateSchema: OpenWikiIndexStateSchema,
     beforeAgent: () => ({ editedWikiPaths: [] }),
-    wrapToolCall: async (request, handler) =>
-      addEditedPathUpdate(await handler(request)),
+    wrapToolCall: async (request, handler) => {
+      const result = await handler(request);
+      await addFrontmatterWarning(
```
colifran: why even go back to the model? if it got it wrong we can just generate ourselves and move on, right?

src/agent/indexing/frontmatter-validator.ts
bracesproul: updated

src/agent/indexing/state.ts
bracesproul: resolved in slack

src/agent/frontmatter-validator.ts
```diff
@@ -0,0 +1,211 @@
...
+interface ValidFrontmatterValidation {
+  valid: true;
+}
+
+interface InvalidFrontmatterValidation {
+  issues: FrontmatterIssue[];
+  valid: false;
+}
+
+export type FrontmatterValidation =
```
colifran, line 18: sorry if i misled you here. when you have an explicit value it is probably better as a type. `FrontmatterIssue` is correct as an interface. Iw ould recommend this for FrontmatterValidation:
```ts
export type FrontmatterValidation =
  | { valid: true }
  | { valid: false; issues: FrontmatterIssue[] };
```

src/agent/frontmatter-validator.ts
```diff
@@ -0,0 +1,211 @@
...
+  ValidFrontmatterValidation | InvalidFrontmatterValidation;
+
+type ReadBackend = Pick<BackendProtocolV2, "readRaw">;
```
colifran: Could we just pass `BackendProtocolV2`

colifran (approved):
Nice!

PR #327, opened 2026-07-14
feat: add aws bedrock to the model provider list

WiFisunset:
## Summary
- Add `bedrock` as a selectable model provider, authenticated with AWS IAM
  credentials (access key ID, secret access key, region) rather than a
  single vendor API key
- Wire `ChatBedrockConverse` (`@langchain/aws`) into model construction,
  with credential/region validation before a run starts
- Extend the interactive onboarding wizard with new "secret key" and
  "region" steps, since Bedrock needs two credential fields where every
  other provider needs one
- Redact the new secret access key from diagnostic/error output
- Document the provider in the README, including a note on cross-region
  inference profiles (see Test plan)

## Why
Lets OpenWiki route inference through Bedrock using existing AWS IAM
credentials, for teams already running in AWS rather than provisioning a
separate vendor API key.

## Test plan
- [x] `pnpm run format`
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm run test` (229 passing, incl. new coverage for the provider
      config and secret redaction)
- [x] `pnpm run build`
- [x] Ran a full end-to-end `openwiki --update` against a real AWS
      Bedrock account (IAM access key/secret + `openai.gpt-oss-120b-1:0`),
      confirming real IAM credential resolution, `ChatBedrockConverse`
      model construction, and the full DeepAgents tool-calling loop all
      work through this provider — verified by inspecting the generated
      wiki output and `.last-update.json`, not just a clean exit code.
  - Note for reviewers: some smaller models on Bedrock (tested:
    `openai.gpt-oss-20b-1:0`) can emit malformed tool-call names that
    Bedrock's Converse API rejects on the next turn — this is a
    model-capability issue, not specific to this integration; larger/more
    disciplined models (Anthropic Claude, `gpt-oss-120b`) don't hit it.
  - Also surfaced and fixed along the way: the README's example model ID
    was fabricated and didn't match AWS's real catalog — corrected to a
    verified ID, plus added a note on the inference-profile requirement
    some newer models have.

Co-Authored with [Claude Code](https://claude.com/claude-code)


colifran:
@WiFisunset thanks for this PR! we need this. i'm working on a review. in the meantime, can you resolve merge conflicts?

src/constants.ts
```diff
@@ -134,6 +156,18 @@ export const PROVIDER_CONFIGS: Record<OpenWikiProvider, ProviderConfig> = {
...
     ],
   },
+  bedrock: {
+    apiKeyEnvKey: BEDROCK_AWS_ACCESS_KEY_ID_ENV_KEY,
+    label: "AWS Bedrock",
+    // Available model IDs are account- and region-specific (they depend on
+    // which foundation models are enabled in Bedrock), so there is no safe
+    // preset list here; paste the Bedrock model ID directly, for example
+    // anthropic.claude-sonnet-5-20260101-v1:0.
+    modelOptions: [],
```
colifran, line 170: nice catch, good call out

src/constants.ts
```diff
@@ -111,6 +116,22 @@ type ProviderConfig = {
   requiresBaseUrl?: boolean;
   label: string;
   modelOptions: ProviderModelOption[];
+  /**
+   * Environment variable holding a second required secret (e.g. an AWS secret
+   * access key paired with {@link ProviderConfig.apiKeyEnvKey} as an access key
+   * ID). Omitted for providers authenticated by a single API key.
+   */
+  secretKeyEnvKey?: string;
```
colifran, line 127: i think fine for now but seems like we should do a better job with separation of concerns here (the internal team). i don't want provider config to become a dumping ground and these new props are bedrock provider specific.

src/constants.ts
```diff
@@ -272,6 +306,41 @@ export function providerRequiresBaseUrl(provider: OpenWikiProvider): boolean {
...
+}
+
+export function providerRequiresRegion(provider: OpenWikiProvider): boolean {
+  return getProviderConfig(provider).requiresRegion === true;
+}
+
+/**
+ * Resolves the configured region for a provider from its region environment
+ * variable. Returns `undefined` when unset, so callers fall back to the SDK's
+ * own region resolution (e.g. `~/.aws/config`).
```
colifran, line 342: awesome

src/constants.ts
colifran: one callout is in the cli i see `gpt-5.6-terra` as the default model. is there some way to fix that? if you think that is a pre-existing bug (which it might be) then we can take it as a follow-up

Image

colifran (commented):
Looks good overall! Thanks for the PR. Left a few comments and maybe one thing to look into. Let me know your thoughts. If we can get the merge conflicts resolved I think this should be basically ready to go.

src/constants.ts
WiFisunset: Thanks! I took a look at the code, and the default fallback is pre-existing and the bedrock impl borrows it. I think it's better if bedrock doesn't rely on that fallback though, adjusting that now.

src/constants.ts
WiFisunset: Thanks!

src/constants.ts
WiFisunset: That makes sense. I'll think about the best approach to adjust that and open up another pr for it.

colifran:
As a heads up I have a fix for the audit check in ci: https://github.com/langchain-ai/openwiki/pull/338. Just in case you're wondering why that is failing.

src/constants.ts
WiFisunset: Got the fix pushed up for this. Leaving this convo open for more feedback or to be closed out if all looks good.

colifran (approved):
Very nice. Thank you @WiFisunset!

src/constants.ts
colifran: Looks good. Thanks!!

PR #330, opened 2026-07-14
fix: require complete chatgpt oauth startup tokens

HwangJohn:
## Summary
- make non-interactive startup validate the complete ChatGPT OAuth token set
- keep API-key provider startup behavior unchanged
- leave expired but complete ChatGPT tokens to the existing agent refresh path

## Testing
- `corepack pnpm exec prettier --check src/startup.ts test/startup.test.ts`
- `corepack pnpm exec vitest run test/startup.test.ts`


colifran (approved):
Nice catch!

PR #331, opened 2026-07-14
fix: tolerate malformed gmail array config

HwangJohn:
## Summary
- Treat Gmail `labelIds` and `metadataHeaders` as untrusted config values.
- Ignore malformed non-array values instead of throwing during query parameter normalization.
- Add a regression test that stubs Gmail fetch and verifies malformed values are omitted.

## Testing
- `corepack pnpm exec prettier --write src/connectors/sources/gmail.ts test/connector-config-overrides.test.ts`
- `corepack pnpm exec vitest run test/connector-config-overrides.test.ts`

### DIFF
--- a/src/connectors/sources/gmail.ts
+++ b/src/connectors/sources/gmail.ts
@@ -25,9 +25,9 @@ type GmailConfig = {
   enabled?: boolean;
   format?: GmailMessageFormat;
   includeSpamTrash?: boolean;
-  labelIds?: string[];
+  labelIds?: unknown;
   maxMessages?: number;
-  metadataHeaders?: string[];
+  metadataHeaders?: unknown;
   pageSize?: number;
   query?: string;
   readOnlyOperations?: unknown[];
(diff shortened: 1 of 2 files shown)


src/connectors/sources/gmail.ts
```diff
@@ -402,10 +402,15 @@ function clamp(value: number | undefined, min: number, max: number): number {
   return Math.max(min, Math.min(max, Math.trunc(value ?? min)));
 }
 
-function normalizeStringArray(values: string[] | undefined): string[] {
-  return (values ?? [])
-    .map((value) => value.trim())
-    .filter((value) => value.length > 0);
+function normalizeStringArray(value: unknown): string[] {
+  return Array.isArray(value)
```
colifran, line 406: very nice fix

colifran (approved):
Very nice. Thanks @HwangJohn! Pushed a small linting fix to your branch.

PR #332, opened 2026-07-14
feat: support openrouter provider pinning

### LINKED ISSUE #39: OpenRouter - Allow pinning of providers of a model
I setup openwiki mostly using defaults. GLM 5.2 & OpenRouter with LangSmith for tracing.

I was messing around attempting to start getting the wiki going.

My issue? One of the providers OpenRouter was choosing for GLM 5.2 is bugging and failing and sending back truncated requests which cause this error in OpenWiki

```
* Error Error invoking tool 'write_file' with kwargs {} with error: Error: Received tool input did not match expected schema

✖ Invalid input: expected string, received undefined
  → at file_path
    at DynamicStructuredTool.call (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/@langchain/core/dist/tools/index.js:113:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async baseHandler (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/langchain/dist/agents/nodes/ToolNode.js:211:20)
    at async ToolNode.wrappedHandler [as wrapToolCall] (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/langchain/dist/agents/utils.js:363:20)
    at async ToolNode.runTool (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/langchain/dist/agents/nodes/ToolNode.js:257:11)
    at async ToolNode.run (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/langchain/dist/agents/nodes/ToolNode.js:292:15)
    at async ToolNode.invoke (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/langchain/dist/agents/RunnableCallable.js:34:23)
    at async RunnableSequence.invoke (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/@langchain/core/dist/runnables/base.js:934:69)
    at async _runWithRetry (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/@langchain/langgraph/dist/pregel/retry.js:48:18)
    at async PregelRunner._executeTasksWithRetry (file:///home/user/.nvm/versions/node/v24.7.0/lib/node_modules/openwiki/node_modules/@langchain/langgraph/dist/pregel/runner.js:147:24)
 Please fix the error and try again.
```
Debugging this I found that OpenRouter is sending back truncated requests which are timing out, but they're concentrated on a specific provider. So I'd like to "pin" my provider to avoid this issue. 

Problem? I'm unable to choose a specific provider in OpenRouter to avoid this issue.

Solution? Allow the users to choose a specific provider when using OpenRouter

---
Debugging from Claude Trace

<h1 id="glm-52-via-openrouter-an-intermittent-upstream-provider-stall-truncates-a-tool-call-causing-a-fatal-write_file-andquotinput-did-not-match-expected-schemaandquot-crash" class="atx">GLM 5.2 via OpenRouter: an intermittent upstream provider stall truncates a tool call, causing a fatal <code>write_file</code> &quot;input did not match expected schema&quot; crash</h1>
<h2 id="summary" class="atx">Summary</h2>
<p>When running OpenWiki with <strong><code>z-ai/glm-5.2</code> through the OpenRouter provider</strong> (the default), documentation runs intermittently crash with a fatal error while the agent is writing a file:</p>
<pre><code class="fenced-code-block">Error invoking tool &#39;write_file&#39; with kwargs {} with error: Error: Received tool input did not match expected schema

✖ Invalid input: expected string, received undefined
  → at file_path
    at DynamicStructuredTool.call (.../node_modules/@langchain/core/dist/tools/index.js:113:10)
    at async ToolNode.runTool (.../node_modules/langchain/dist/agents/nodes/ToolNode.js:257:11)
    ...</code></pre>
<p>On the surface this reads like the model produced an empty tool call. It did not. After tracing it, the actual cause is an <strong>intermittent stall/drop by the upstream inference provider that OpenRouter routed the request to</strong>, which truncates the streamed <code>write_file</code> tool call. That truncation is then silently degraded to an empty arguments object <code>{}</code> by the streaming layer, and finally rejected by the tool&#39;s schema. The whole run aborts with no retry.</p>
<p>This report focuses on <em>why the request fails</em>. It walks through the evidence so it can be understood without access to our traces.</p>
<h2 id="environment" class="atx">Environment</h2>
<ul>
<li>OpenWiki <code>0.0.1</code></li>
<li>Provider: <strong>OpenRouter</strong> (<code>OPENWIKI_PROVIDER=openrouter</code>), model <code>OPENWIKI_MODEL_ID=z-ai/glm-5.2</code></li>
<li>OpenRouter served the request from the model snapshot <code>z-ai/glm-5.2-20260616</code></li>
<li>Relevant dependency versions:<ul>
<li><code>deepagents</code> 1.10.5</li>
<li><code>langchain</code> 1.5.1</li>
<li><code>@langchain/core</code> 1.2.1</li>
<li><code>@langchain/openrouter</code> 0.4.3</li>
</ul>
</li>
<li>Node 24.x</li>
</ul>
<h2 id="what-the-agent-is-doing-when-it-fails" class="atx">What the agent is doing when it fails</h2>
<p>OpenWiki runs a DeepAgents agent that uses the built-in filesystem tools to write documentation. The <code>write_file</code> tool is a LangChain structured tool whose input schema requires two string fields:</p>
<pre><code class="fenced-code-block language-ts">schema: z.object({
  file_path: z.string().describe(&quot;Absolute path to the file to write&quot;),
  content: z.string().default(&quot;&quot;).describe(&quot;Content to write to the file&quot;),
})</code></pre>
<p>The model is invoked in <strong>streaming</strong> mode (the agent is streamed so the CLI can render output as it arrives). Because OpenWiki generates docs, the <code>content</code> argument to <code>write_file</code> is frequently large (multiple KB of markdown), so these tool calls take a while to stream.</p>
<h2 id="root-cause" class="atx">Root cause</h2>
<p>The failure chain, in order:</p>
<ol>
<li>OpenWiki sends a streaming chat-completion request for <code>z-ai/glm-5.2</code> to OpenRouter. OpenWiki does not specify an upstream provider, so <strong>OpenRouter auto-selects one</strong> from the ~28 providers that serve this model.</li>
<li>The selected upstream provider begins streaming the assistant&#39;s <code>write_file</code> tool call, emits the very beginning of the arguments — literally the string <code>{&quot;content&quot;: </code> — and then <strong>stalls and drops the stream</strong> without ever sending a finish reason. This is confirmed by OpenRouter&#39;s own generation records: <code>native_finish_reason: null</code>, <code>cancelled: false</code> (i.e. <em>the client did not abort</em>), and a <code>generation_time</code> of ~2 minutes for ~0 useful output.</li>
<li>In streaming mode, <code>@langchain/openrouter</code>&#39;s <code>ChatOpenRouter</code> <strong>does not throw</strong> when a stream ends this way. It simply stops iterating and hands back the partial message it accumulated.</li>
<li>LangChain reassembles the streamed tool-call arguments. The accumulated argument string is the truncated, unparseable <code>{&quot;content&quot;: </code>. LangChain&#39;s partial-JSON parser turns that into an <strong>empty object <code>{}</code></strong> rather than raising.</li>
<li>The tool call — now with arguments <code>{}</code> — is dispatched to <code>write_file</code>. The zod schema requires <code>file_path: z.string()</code>, but it is <code>undefined</code>, so validation fails:
<code>Invalid input: expected string, received undefined → at file_path</code>.</li>
<li>This error is <strong>fatal to the run</strong>. OpenWiki&#39;s model-fallback/retry only re-runs on OpenRouter HTTP 5xx responses; a mid-stream stall returns HTTP 200, so nothing retries, and the raw error surfaces to the user.</li>
</ol>
<p>So the request &quot;fails&quot; not because the model misbehaved, and not because of a client timeout, but because an <strong>intermittent upstream stall truncates the tool call</strong>, and every layer between the provider and the tool <strong>silently degrades that truncation</strong> (drop → empty stream → <code>{}</code> → schema error) instead of surfacing a clear, retryable failure.</p>
<h2 id="evidence" class="atx">Evidence</h2>
<p>We reproduced the crash and then inspected it two ways: LangSmith traces (OpenWiki writes traces to a project when a LangSmith key is configured) and OpenRouter&#39;s per-generation metadata endpoint (<code>GET /api/v1/generation?id=&lt;gen id&gt;</code>). Anyone can gather the same data by enabling LangSmith tracing and looking up the generation id (the response <code>id</code>, e.g. <code>gen-...</code>) for a failing run.</p>
<p><strong>The two failing model calls both look like this</strong> (the model/LLM run that produced the crashing tool call):</p>

Field (from the model run) | Failing run A | Failing run B
-- | -- | --
OpenRouter upstream provider | Novita | Novita
generation_time | 124.2 s | 280.1 s
time to first token (latency) | ~4.1 s (normal) | ~4.3 s (normal)
completion tokens produced | 8 | 3789 (+4376 reasoning)
native_finish_reason (from OpenRouter) | null | null
cancelled (client aborted?) | false | false
LangChain finish_reason on the run | error | error
raw streamed tool-call arguments | {"content":  (12 chars) | {"content":  (12 chars)
resulting parsed tool args | {} | {}

<p>The same provider serves complete, valid tool calls most of the time (we observed successful <code>write_file</code> calls with 5–20 KB of arguments). So this is <strong>tail behavior</strong>, not a permanent outage — which is why it&#39;s intermittent and hard to catch. (For reference, that provider&#39;s aggregate 30-minute uptime as reported by OpenRouter was ~99.6% at the time.)</p>
<h2 id="why-itand39s-intermittent--when-it-triggers" class="atx">Why it&#39;s intermittent / when it triggers</h2>
<ul>
<li>It requires the upstream provider to stall <em>during</em> a generation. This appears more likely on the long tool-call generations OpenWiki produces (large <code>content</code>), but it is not strictly size-bound — it happened after only 8 tokens in one case.</li>
<li>Because OpenRouter auto-selects the provider per request, whether you hit a stall-prone provider varies run to run.</li>
<li>The truncation always lands the same way: the arguments are cut off after the first key opens (<code>{&quot;content&quot;: </code>), which parses to <code>{}</code>, which is missing the required <code>file_path</code>.</li>
</ul>
<h2 id="impact" class="atx">Impact</h2>
<ul>
<li>Documentation runs (<code>--init</code> / <code>--update</code> / interactive) crash partway through with a confusing error that appears to blame the model or the tool schema, when the real cause is an upstream stream drop.</li>
<li>There is no automatic recovery: the crash is fatal and not retried, so an intermittent provider hiccup wastes the whole run (including the tokens already billed for the large prompt).</li>
<li>The error message is misleading for debugging: &quot;<code>write_file</code> with kwargs <code>{}</code> … expected string, received undefined at file_path&quot; gives no hint that the underlying event was an upstream provider stall and a truncated stream.</li>
</ul>
<h2 id="how-to-reproduce" class="atx">How to reproduce</h2>
<ol>
<li>Configure OpenWiki with OpenRouter and <code>z-ai/glm-5.2</code>.</li>
<li>Run documentation generation on a repository large/complex enough that the agent writes sizable files (so tool-call arguments are large and take time to stream).</li>
<li>Re-run as needed; intermittently a run will crash with the <code>write_file … kwargs {} … expected string, received undefined → at file_path</code> error.</li>
<li>To confirm the cause on a failing run: enable LangSmith tracing, open the model run that produced the failing <code>write_file</code>, and note <code>finish_reason: error</code> with the tool-call arguments truncated to <code>{&quot;content&quot;: </code>. Cross-reference the response id (<code>gen-...</code>) via OpenRouter&#39;s <code>GET /api/v1/generation?id=...</code> — you&#39;ll see <code>native_finish_reason: null</code>, <code>cancelled: false</code>, and an unusually long <code>generation_time</code> relative to tokens produced.</li>
</ol>
<h2 id="notes-for-maintainers-diagnostic-not-prescriptive" class="atx">Notes for maintainers (diagnostic, not prescriptive)</h2>
<p>The single upstream event (a provider stalling/dropping a stream) becomes a fatal, cryptic error because it is silently degraded at three layers:</p>
<ul>
<li><code>ChatOpenRouter</code> streaming does not raise when a stream ends with a provider drop / <code>finish_reason: error</code>;</li>
<li>LangChain coerces the truncated tool-call JSON (<code>{&quot;content&quot;: </code>) into <code>{}</code> rather than flagging it as invalid;</li>
<li>the empty <code>{}</code> reaches <code>write_file</code> and only then fails, as a schema error, with no retry (OpenWiki&#39;s existing retry only covers HTTP 5xx).</li>
</ul>
<p>Any of these could reasonably be where the condition is detected and surfaced/handled.</p>

HwangJohn:
## Summary
- Add `OPENWIKI_OPENROUTER_PROVIDER_ONLY` to pin OpenRouter routing to one upstream provider or a comma-separated allowlist.
- Pass the configured values to `ChatOpenRouter` as `provider.only`.
- Persist and surface the new routing key in OpenWiki env diagnostics, and document the setting in the README.

Fixes #39.

## Testing
- `corepack pnpm exec prettier --write README.md src/constants.ts src/env.ts src/agent/index.ts test/constants.test.ts`
- `corepack pnpm exec vitest run test/constants.test.ts test/env.test.ts` — passed
- `corepack pnpm run typecheck` — passed
- `corepack pnpm exec vitest run test/constants.test.ts test/env.test.ts test/env-behavior.test.ts` — only failed the existing Windows chmod assertion in `test/env-behavior.test.ts > writes the env file with 0600 permissions`


colifran (approved):
Thanks @HwangJohn! Looks good to me. As a follow-up we should probably surface this as a configurable option as part of openwiki --init.

PR #334, opened 2026-07-14
fix: stop env-behavior tests from overwriting the real ~/.openwiki/.env

omkamal:
## What

`test/env-behavior.test.ts` imports `../src/env.ts` statically, but `src/env.ts` resolves `openWikiEnvPath` from `os.homedir()` **at module load time**. The `beforeEach` hook's `HOME` swap therefore happens *after* the path is already bound to the developer's real home directory — so every `saveOpenWikiEnv()` call in the file writes test fixtures (`sk-secret-12345`, `bad model id`, `https://gateway.example.com/anthropic`, …) straight into the developer's **real `~/.openwiki/.env`**, silently destroying their saved provider credentials and connector tokens. The file's own header comment claims the tests are "fully isolated from the developer's real credentials" — this PR makes that true.

This bit me for real: one `pnpm test` run replaced my configured provider, model ID, Google OAuth client, and Gmail tokens with fixtures, and I had to re-authenticate everything.

## Why this fix

`test/onboarding.test.ts` already solves the identical problem correctly: `vi.resetModules()` + set `HOME` + dynamically `import()` the module under test. This PR applies the same pattern to `env-behavior.test.ts`, and adds a regression test asserting that the resolved `openWikiEnvPath` lives inside the test's temp `HOME` — so a future return to static imports fails loudly instead of silently clobbering real credentials.

One PR = one change: only the test file is touched. (A follow-up could make `src/env.ts` / `src/openwiki-home.ts` resolve paths lazily so the footgun disappears at the source — happy to open that separately if wanted.)

## How I tested it

Repro before the fix (on `main`):

```sh
sha256sum ~/.openwiki/.env          # note hash
pnpm exec vitest run test/env-behavior.test.ts
sha256sum ~/.openwiki/.env          # hash CHANGED; file now contains test fixtures
```

After the fix:

- `pnpm exec vitest run test/env-behavior.test.ts` — 15/15 pass (14 existing + 1 new regression test)
- `pnpm test` — full suite 224/224 pass
- `sha256sum ~/.openwiki/.env` before/after the full suite — **unchanged**
- `pnpm run lint:check` / `pnpm run format:check` — clean on the changed file

### DIFF
--- a/test/env-behavior.test.ts
+++ b/test/env-behavior.test.ts
@@ -8,13 +8,7 @@ import {
 } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
-import { afterEach, beforeEach, describe, expect, test } from "vitest";
-import {
-  getCredentialDiagnostics,
-  loadOpenWikiEnv,
-  openWikiEnvPath,
-  saveOpenWikiEnv,
-} from "../src/env.ts";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 import {
   ANTHROPIC_API_KEY_ENV_KEY,
   ANTHROPIC_BASE_URL_ENV_KEY,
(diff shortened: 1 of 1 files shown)


colifran:
Noticed this too! Thank you for opening this PR!

colifran (approved):
This is a nasty bug. Thank you @omkamal!

PR #336, opened 2026-07-14
fix: redirect conversation_history offloads out of target repo

### LINKED ISSUE #49: Summarization offloads write conversation_history/ into the target repository

blue-az:
Summarization middleware offloads write to /conversation_history,
which the OpenWikiLocalShellBackend resolved under the target
repo's cwd, polluting the workspace being documented. Redirect
those virtual paths to ~/.openwiki/conversation_history instead.

Fixes #49

### DIFF
--- a/src/agent/docs-only-backend.ts
+++ b/src/agent/docs-only-backend.ts
@@ -1,3 +1,4 @@
+import path from "node:path";
 import {
   LocalShellBackend,
   type EditResult,
@@ -5,6 +6,7 @@ import {
   type WriteResult,
 } from "deepagents";
 import { OPEN_WIKI_DIR } from "../constants.js";
+import { openWikiHistoryDir } from "../openwiki-home.js";
 import type { OpenWikiOutputMode } from "./types.js";
 
 type OpenWikiBackendOptions = LocalShellBackendOptions & {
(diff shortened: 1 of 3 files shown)


colifran:
This was fixed in #504

PR #342, opened 2026-07-15
fix: name the code-mode update job after the detected Git host

### LINKED ISSUE #341: code mode is hardcoded to GitHub (workflow file + AGENTS.md wording) on non-GitHub repos
## Summary

`openwiki code` is hardcoded to GitHub in two places, even when the repository uses a different git host (GitLab, Bitbucket, etc.). This makes it awkward to use OpenWiki in non-GitHub repos, because every run reintroduces GitHub-specific artifacts/wording that then have to be scrubbed in CI.

There are two related problems.

### 1. A GitHub Actions workflow is written on every run, unconditionally

`ensureCodeModeRepoSetup()` calls `writeCodeModeWorkflow()`, which always creates/overwrites `.github/workflows/openwiki-update.yml` — regardless of the repo's git provider. It runs on both the interactive and the `--print` (CI) paths (`src/code-mode.ts`, called from `src/cli.tsx`).

For a Bitbucket- or GitLab-only repo this means:
- A `.github/workflows/openwiki-update.yml` reappears on every `openwiki code --update` run.
- CI pipelines have to `rm -rf .github` (or otherwise exclude it) before committing, so the unused workflow doesn't leak into the docs PR.

### 2. The `AGENTS.md` / `CLAUDE.md` OPENWIKI block hardcodes "GitHub Actions workflow"

`writeCodeModeAgentSnippets()` rewrites the `<!-- OPENWIKI:START -->…<!-- OPENWIKI:END -->` block in `AGENTS.md` and `CLAUDE.md` on every run, and the template contains:

> The scheduled OpenWiki **GitHub Actions workflow** refreshes the repository wiki.

On a Bitbucket/GitLab repo this sentence is factually wrong, and because the block is regenerated each run, any manual correction is reverted the next time OpenWiki runs — producing a recurring, misleading diff. In our Bitbucket setup we currently patch it back with a `sed` step after every run, which is a workaround, not a fix.

## Steps to reproduce

1. In a repo that is **not** hosted on GitHub (e.g. Bitbucket), run `openwiki code --update --print` with a provider configured.
2. Observe:
   - `.github/workflows/openwiki-update.yml` is (re)created.
   - The OPENWIKI block in `AGENTS.md`/`CLAUDE.md` says "GitHub Actions workflow".

## Expected

OpenWiki should be provider-aware, or at least not emit GitHub-specific content when it isn't wanted. Some options:

- Detect the git remote host and skip `writeCodeModeWorkflow()` / emit the matching CI file (there are already GitLab and Bitbucket examples under `examples/`).
- Add a flag / config (e.g. `--ci-provider none|github|gitlab|bitbucket`, or a field in the OpenWiki config) to control which CI workflow is written, defaulting to none in `--print`/CI mode.
- Make the OPENWIKI block wording provider-neutral (e.g. "the scheduled OpenWiki update job refreshes the repository wiki") so it's accurate everywhere, or template it from the detected/configured provider.

## Environment

- OpenWiki installed via `npm install --global openwiki`
- Used in CI on Bitbucket Pipelines (`openwiki code --update --print`)

Happy to send a PR if a preferred direction is decided (provider detection vs. an explicit flag).

victor-mico:
## What and why

Fixes #341. Rebased onto current `main` and narrowed to what `main` does not already cover.

`main` has since fixed most of the workflow half by another route: `createWorkflow` is now gated on `mode === "init"`, and `ensureCodeModeWorkflow` returns early when the file exists, so `code --update` no longer reintroduces `.github/workflows/openwiki-update.yml`. What is left is what this PR fixes:

1. `code --init` still writes the GitHub Actions workflow on a GitLab or Bitbucket repository, which cannot run it.
2. The managed `<!-- OPENWIKI:START -->…<!-- OPENWIKI:END -->` block still says "The scheduled OpenWiki **GitHub Actions workflow** refreshes the repository wiki" on every host. On a non-GitHub repo that is false, and since the block is rewritten each run, a manual correction never survives (also reported independently in #820).

## What changed

- **`src/ingestion/code-mode.ts`**: resolve the host from the repository remote (`origin`, else any remote), overridable with `OPENWIKI_CI_PROVIDER=github|gitlab|bitbucket|none`.
  - Create the GitHub Actions workflow only when the host is GitHub — `--init` on GitLab/Bitbucket now leaves `.github/` alone.
  - Name the recurring job after the detected host: GitHub Actions workflow / GitLab pipeline / Bitbucket pipeline / a neutral "scheduled OpenWiki update job".
  - Detection reads the remote's **hostname only**, so `git@github.com:acme/gitlab-importer.git` stays GitHub while `gitlab.acme.com` is still recognized. An unrecognized host, an absent remote, an unrecognized override value, and a failed or timed-out `git` all fall back to GitHub — the behavior every existing repo was set up under. Detection never throws and is bounded by a 10s timeout.
- **`examples/`**: the GitLab and Bitbucket pipelines set `OPENWIKI_CI_PROVIDER` explicitly and stop tracking the GitHub workflow path OpenWiki no longer writes for them.
- **`README.md`**: documents detection and the override.

## Review comments from the previous round

- *"is it safe to match on the entire url?"* — no, and it is now fixed: `parseRemoteHost` extracts the hostname (URL form and scp-like `git@host:path` form) and only that is matched. `git@github.com:acme/gitlab-importer.git` is covered by a test.
- *"does this need a default case?"* — added. `describeCodeModeUpdateJob` falls through to the provider-neutral phrasing for `none` and for any host added later without wording of its own, so a new provider can never inherit the wrong name.

## Relationship to #835 and #742

#835 proposes **omitting** the sentence unless a `.github` workflow with an `on.schedule` trigger can be confirmed. That is strictly better than asserting something false, but a Bitbucket repo driven by a scheduled pipeline does have a schedule — it is just not visible under `.github` — so it ends up with no statement rather than a correct one. This PR keeps the claim and makes it true. The two are compatible: host detection here can render the sentence, with #835's workflow check deciding whether to state it at all, and either can fold into #742's persistent config shape. Happy to converge on whichever direction a maintainer prefers.

## Testing

- `test/ingestion/code-mode.test.ts`: workflow skipped and matching wording for gitlab/bitbucket/none; detection from a real git remote; hostname-not-path matching; self-hosted `gitlab.*` host; env override beating detection; unrecognized override falling through to detection; unknown host defaulting to GitHub.
- `pnpm run format:check`, `pnpm run lint:check`, `pnpm run typecheck` all clean.
- `pnpm test`: 3041 passed. The 6 failures are pre-existing on `main` in this environment (mermaid render timeouts and the openai-compatible transport tests) and reproduce with the branch's changes reverted.

### DIFF
--- a/.changeset/host-aware-code-mode-setup.md
+++ b/.changeset/host-aware-code-mode-setup.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: name the code-mode update job after the detected Git host instead of always claiming GitHub Actions
--- a/README.md
+++ b/README.md
@@ -66,6 +66,8 @@ Keep it current automatically by adding a scheduled CI job that opens a docs PR 
 - **GitLab CI:** copy [`openwiki-update.gitlab-ci.yml`](./examples/openwiki-update.gitlab-ci.yml) into `.gitlab-ci.yml` or include it from your pipeline.
 - **Bitbucket Pipelines:** copy [`openwiki-update.bitbucket-pipelines.yml`](./examples/openwiki-update.bitbucket-pipelines.yml) into `bitbucket-pipelines.yml`, then schedule the `openwiki-update` pipeline.
 
+`openwiki --init` writes the GitHub Actions workflow for you when the repository's remote is a GitHub host, and leaves a GitLab or Bitbucket repository to the examples above. The managed `AGENTS.md` block names the job for the detected host, so it never tells your agents about a GitHub Actions run the repository cannot perform. Set `OPENWIKI_CI_PROVIDER=github|gitlab|bitbucket|none` to override detection on a self-hosted instance whose hostname does not name its provider, or to `none` to keep OpenWiki from writing any workflow.
+
 > [!NOTE]
 > On Windows, install with a Node.js package manager (`npm install -g openwiki` or `pnpm add -g openwiki`). Installing with `bun` can fall back to compiling the `better-sqlite3` native dependency, which needs Visual Studio Build Tools with the Desktop development with C++ workload.
 
@@ -234,7 +236,7 @@ Locally the setup wizard saves this to `~/.openwiki/.env`. In CI, set it as a re
 
 Your wiki stays in the repository as plain Markdown you own, with OpenWiki-managed grounding and run metadata versioned alongside it.
 
-- **Agents read it as memory.** On each `code` run, OpenWiki maintains an `AGENTS.md` and `CLAUDE.md` at the repo root that point your coding agent at the wiki. It only rewrites its own `<!-- OPENWIKI:START -->…<!-- OPENWIKI:END -->` block and leaves the rest of each file untouched.
+- **Agents read it as memory.** On each `code` run, OpenWiki maintains an `AGENTS.md` and `CLAUDE.md` at the repo root that point your coding agent at the wiki. It only rewrites its own `<!-- OPENWIKI:START -->…<!-- OPENWIKI:END -->` block and leaves the rest of each file untouched, and names the scheduled update job after your detected Git host.
 - **Grounding stays with the wiki.** Versioned claim sidecars under `openwiki/.claims/` travel with the Markdown, so the evidence needed to maintain factual pages is inspectable and reviewable.
 - **You set the brief.** Repository-specific instructions live in `openwiki/INSTRUCTIONS.md`, a user-authored file OpenWiki reads for scope and priorities but never rewrites during normal runs.
 - **No-op runs do not churn docs.** A clean update skips model work and leaves wiki content untouched while refreshing `.last-update.json` to record that the check ran.
--- a/examples/openwiki-update.bitbucket-pipelines.yml
+++ b/examples/openwiki-update.bitbucket-pipelines.yml
@@ -18,15 +18,20 @@ pipelines:
             - npm install --global openwiki mermaid@11.16.0 jsdom@29.1.1
             - git config user.name "${BITBUCKET_USER_NAME:-OpenWiki Bot}"
             - git config user.email "${BITBUCKET_USER_EMAIL:-openwiki@example.com}"
+            # Keep the managed AGENTS.md block naming this pipeline instead
+            # of a GitHub Actions workflow, and keep `openwiki --init` from
+            # writing one. Detection from the git remote covers bitbucket.org;
+            # set it explicitly for a self-hosted host that is not named for it.
+            - export OPENWIKI_CI_PROVIDER=bitbucket
             - openwiki code --update --print
             - |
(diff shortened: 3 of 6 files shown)


src/code-mode.ts
```diff
@@ -14,8 +28,95 @@ export async function ensureCodeModeRepoSetup(
...
+// instances typically keep the provider name in their hostname, so a substring
+// match is deliberate. Unknown hosts (and missing remotes) fall back to GitHub
+// to preserve the historical default; use OPENWIKI_CI_PROVIDER to override.
+function classifyRemoteHost(remoteUrl: string | null): CiProvider {
+  if (!remoteUrl) {
+    return "github";
+  }
+
+  const lower = remoteUrl.toLowerCase();
+  if (lower.includes("gitlab")) {
```
colifran: is it safe to match on the entire url? should we just be matching on the host part? for example, what if someone has something like:
```
git@github.com:acme/gitlab-importer.git
```

src/code-mode.ts
```diff
@@ -121,14 +225,29 @@ jobs:
...
 
-The scheduled OpenWiki GitHub Actions workflow refreshes the repository wiki. Do not hand-edit generated OpenWiki pages unless explicitly asked; prefer updating source code/docs and letting OpenWiki regenerate.
+The ${describeCodeModeUpdateJob(provider)} refreshes the repository wiki. Do not hand-edit generated OpenWiki pages unless explicitly asked; prefer updating source code/docs and letting OpenWiki regenerate.
 
 ${OPENWIKI_AGENTS_SNIPPET_END}`;
 }
+
+// Provider-aware phrasing for the OPENWIKI agent-file block so the sentence is
+// accurate on non-GitHub hosts instead of always claiming a GitHub Actions run.
+function describeCodeModeUpdateJob(provider: CiProvider): string {
```
colifran: does this need a default case?

colifran (commented):
Thanks for the PR @victor-mico! Left a few comments but otherwise looks great.

src/code-mode.ts
victor-mico: Good catch — matching the whole URL was wrong, and `git@github.com:acme/gitlab-importer.git` is exactly the case it broke on. Fixed in the rebase: `parseRemoteHost` now extracts the hostname (both the URL form and the scp-like `git@host:path` form, which is not a parseable URL) and only the hostname is matched, so a repository *named* after another host stays on its real one while a self-hosted `gitlab.acme.com` is still recognized. Both are covered by tests.

src/code-mode.ts
victor-mico: Added. `describeCodeModeUpdateJob` now falls through to the provider-neutral "scheduled OpenWiki update job" for `none` and for any host added later without wording of its own, so a new provider can never silently inherit another host's name. `normalizeCiProvider` has a default too: an unrecognized `OPENWIKI_CI_PROVIDER` value falls through to detection rather than failing setup, so a typo in a pipeline variable does not break the docs run.

PR #344, opened 2026-07-15
feat(connectors): add IMA OpenAPI connector

DuanZGit:
Adds a built-in IMA connector that uses the IMA OpenAPI to discover knowledge bases and fetch media info. Validated locally against the live IMA API and with an end-to-end  run on a personal OpenWiki instance.


colifran:
Thanks for the PR - unfortunately I don't think we want to support this at the moment.

PR #280, opened 2026-07-11
feat: add Grok Build subscription CLI as a provider

sanjay3290:
## Summary

Adds a `grok-build` provider that runs OpenWiki documentation jobs through the local **Grok Build CLI** (`grok`) using a subscription login instead of a metered API key. OpenWiki spawns the binary headlessly with `--always-approve --output-format streaming-json`, streams and coalesces its NDJSON events, and reuses the CLI's own session for interactive follow-ups.

Intended for local/subscription use on a machine already logged into Grok Build interactively — not a drop-in for the metered CI examples.

## What changed

- **Provider config becomes a discriminated union** (`api` | `agent-cli`). Agent-CLI providers carry a binary path + install hint instead of an API key, and onboarding, credential setup, and the non-interactive startup check all skip key collection for them.
- **New `src/agent/engines/` module**: a generic agent-CLI runner (temp prompt file to avoid ARG_MAX, detached process group with SIGTERM→SIGKILL timeout, NDJSON stream parsing) plus a Grok Build adapter and streaming-json parser that coalesces partial text tokens into a single final answer.
- **Shared run scaffolding**: extracted `prepareAgentRun` / `finalizeAgentRun` so the CLI and DeepAgents paths share snapshot + `.last-update.json` metadata handling.
- **Docs**: README section covering prerequisites (`grok login`), env overrides (`OPENWIKI_GROK_BUILD_BINARY`, `OPENWIKI_GROK_BUILD_MAX_TURNS`, `OPENWIKI_AGENT_CLI_TIMEOUT_SECONDS`), and the local-only caveat.

## Config

| Env var | Purpose |
|---|---|
| `OPENWIKI_PROVIDER=grok-build` | Select the provider |
| `OPENWIKI_GROK_BUILD_BINARY` | Binary path when `grok` is not on `PATH` |
| `OPENWIKI_MODEL_ID` | Model (default `grok-4.5`) |
| `OPENWIKI_GROK_BUILD_MAX_TURNS` | Max agent turns (default 50) |
| `OPENWIKI_AGENT_CLI_TIMEOUT_SECONDS` | Run timeout (default 1800) |

## Testing

- `tsc --noEmit` — clean
- `eslint` — clean
- Full suite: **200/200 tests pass**, including new coverage for the adapter, provider integration, and CLI runner.

PR #356, opened 2026-07-16
feat: add Antigravity subscription CLI as a provider

sanjay3290:
## Summary

Adds an `antigravity` provider that runs OpenWiki documentation jobs through the local **Antigravity CLI** (`agy`) using a subscription/session login instead of a metered API key.

This builds on the agent-CLI foundation introduced for Grok Build (see also #280): OpenWiki spawns the vendor binary headlessly, streams stdout, and reuses the CLI session for interactive follow-ups.

## What changed

- **`antigravity` agent-cli provider** (`agy` binary, override via `OPENWIKI_ANTIGRAVITY_BINARY`)
- **Plain-text stream format** on the shared agent-CLI runner (Antigravity print mode is not NDJSON)
- **Session resume** via `--conversation <id>`, recovered from a per-run `--log-file`
- **Model presets** use exact `agy models` display strings (e.g. `Gemini 3.5 Flash (Medium)`); `isValidModelId` now allows spaces/parentheses for those names
- **Docs** + tests for adapter, provider config, and text-stream runner path

## Config

| Env var | Purpose |
|---|---|
| `OPENWIKI_PROVIDER=antigravity` | Select the provider |
| `OPENWIKI_ANTIGRAVITY_BINARY` | Binary path when `agy` is not on `PATH` |
| `OPENWIKI_MODEL_ID` | Exact `agy models` display name (default `Gemini 3.5 Flash (Medium)`) |
| `OPENWIKI_AGENT_CLI_TIMEOUT_SECONDS` | Run / print-timeout budget (default 1800) |

## Runtime flags

```text
agy -p <prompt> --dangerously-skip-permissions --mode accept-edits \
  --print-timeout <n>m<n>s --log-file <tmp> --add-dir <cwd> \
  [--model <display name>] [--conversation <id>]
```

## Testing

- `pnpm run typecheck` — clean
- `pnpm run lint:check` — clean
- Full suite: **296/296 tests pass**

## Notes

- Local/subscription only — not a drop-in for CI examples that expect API keys
- Model ids must match `agy models` verbatim; near-misses can no-op with empty stdout
- Includes the agent-CLI infrastructure also present in #280 (Grok Build). Happy to re-split or rebase if preferred.

## Review

@bracesproul @colifran — would appreciate a review when you have a chance.


sanjay3290:
Addressed both Corridor findings in 16a452e:

1. **Env scrubbing** (`src/agent/engines/child-env.ts` + `runner.ts`): vendor agent CLIs are now spawned with an explicit allowlisted `env` (PATH/HOME/tmp/locale/proxy/CA only). All `MANAGED_ENV_KEYS`, secret-like keys, and `OPENWIKI_*` / `LANGCHAIN_*` / `LANGSMITH_*` vars are dropped so credentials loaded by `loadOpenWikiEnv()` never reach the child.

2. **Docs-only write boundary** (`src/agent/engines/write-boundary.ts` + `runner.ts`): repository `init`/`update` runs set `writeBoundary: "docs-only"`. After a successful agent-CLI exit, OpenWiki scans the cwd for files mtime-touched during the run and fails the run if any path is outside `openwiki/` (plus root `AGENTS.md` / `CLAUDE.md`). Chat and local-wiki runs keep `writeBoundary: "none"` (cwd is already the wiki root for local-wiki).

Tests: `test/agent-cli-security.test.ts` covers scrubbing + allowed/disallowed writes. Full suite green (303 tests).

@corridor-security — treating both findings as valid; mitigated as above (post-run mtime boundary rather than full OS sandbox, which is the practical control for a local CLI integration).

sanjay3290:
Also landed the `.git` write-boundary scan fix here (same as #280): VCS metadata trees are no longer skipped, so planted hooks under `.git/hooks/` fail docs-only runs.

colifran:
Thank you for building this, genuinely. We've decided not to delegate OpenWiki's agent loop to an external CLI, because that runs tool and shell execution outside the docs-only sandbox the project relies on.

PR #365, opened 2026-07-16
fix: record interrupted runs so the update no-op check retries them

### LINKED ISSUE #364: bug: an interrupted update run self-reports as fresh and the no-op check then skips every retry
### What happens

Since #277 (merged), a run that fails mid-stream still persists `openwiki/.last-update.json` — deliberately, so content that was already generated stays diffable. But the metadata schema has no notion of how the run ended (`updatedAt`, `command`, `gitHead`, `model` only), so a crashed half-update is indistinguishable from a completed one.

The next `openwiki --update` (no user message) then runs `getUpdateNoopStatus`: clean worktree + `HEAD === lastUpdate.gitHead` → *"No repository changes detected since the last OpenWiki update; skipping agent run."* The retry is silently skipped, and keeps being skipped until an unrelated commit lands in the repo.

### Why it matters

A possibly **partial wiki self-reports as fresh and locks itself in**:

1. Scheduled `openwiki code --update --print` crashes late (provider error, timeout) after writing some pages — metadata is persisted (per #277).
2. Every subsequent scheduled run is a no-op skip.
3. Consumers (coding agents pointed at the wiki via AGENTS.md/CLAUDE.md) read a half-written wiki whose `.last-update.json` carries a fresh `updatedAt`.

### Expected

An interrupted run should be recorded as such, and the update no-op check should not skip while the last run is interrupted — while metadata written by older versions (no status field) keeps behaving as a completed run, so upgrades don't force spurious re-runs.

I have a tightly-scoped fix with regression tests ready — PR incoming.

WenyuChiou:
### What

Follow-up to #277. Update metadata now records how the run ended, and the update no-op check retries interrupted runs instead of locking them in:

- `UpdateMetadata` gains an optional `status: "complete" | "interrupted"`.
- The failure path in `runOpenWikiAgentCore` (the #277 catch block) persists metadata with `status: "interrupted"`; successful runs record `"complete"`.
- `getUpdateNoopStatus` returns `shouldSkip: false` (reason: `previous update was interrupted`) when the last run was interrupted, so the next scheduled/manual `--update` actually retries.
- **Back-compat:** metadata written by older versions has no `status` field and is normalized to `"complete"` — upgrading does not force a spurious re-run.
- **Recovery:** a completed retry that ends up changing no content still rewrites the metadata once to clear a leftover `interrupted` status — otherwise the no-op skip would never engage again and every scheduled run would pay a full agent pass (the #44/#51 cost class).

### Why

#277 correctly keeps partially-generated content diffable, but it left a crashed run indistinguishable from a complete one. A half-written wiki then self-reports as fresh (`updatedAt` is new), and the no-op gate skips every retry until an unrelated commit lands — bad for both correctness (agents consume a partial wiki that claims to be current) and scheduled-CI semantics.

### How tested

- 5 new regression tests across `test/update-noop.test.ts` and `test/run-metadata.test.ts`, mirroring the existing temp-repo patterns: interrupted metadata → no skip with the exact reason; pre-status-field metadata still skips (back-compat guard); `persistRunMetadataIfChanged` records `complete`/`interrupted`; a no-change completed retry clears the interrupted flag; a no-change retry after a complete run still writes nothing. **3 of the new assertions fail against the previous code** — verified before applying the fix.
- `pnpm run format`, `pnpm run lint`, `pnpm test`: 300/302 pass on my machine (Windows). The 2 failures are the pre-existing win32 `test/env-behavior.test.ts` platform checks (0600 permissions / HOME resolution), which reproduce on pristine `main` — CI (ubuntu) should be fully green.

Fixes #364


colifran (approved):
Awesome. Very nice fix! Thanks @WenyuChiou!

PR #367, opened 2026-07-16
fix: restrict ~/.openwiki ACLs on Windows where chmod is a no-op

### LINKED ISSUE #366: bug: ~/.openwiki permissions are a no-op on Windows — chmod 0o700/0o600 leaves the credential store on inherited ACLs
### What happens

OpenWiki carefully sets POSIX permissions on its home directory everywhere it is ensured — `mkdir(..., { mode: 0o700 })` + `chmod(0o700)` in `ensureOpenWikiHome` (`src/openwiki-home.ts`) and `saveOpenWikiEnv` (`src/env.ts`), and `0o600` on `.env` and the sqlite checkpoint. On Windows, however, Node's `fs.chmod` only toggles the read-only attribute ([documented Node caveat](https://nodejs.org/api/fs.html#fspromiseschmodpath-mode)) and the `mode` option on `mkdir`/`writeFile` is ignored — so none of these calls restrict anything. `~/.openwiki` just inherits whatever ACL its parent directory grants.

### Why it matters

`~/.openwiki` is the highest-value directory OpenWiki manages:

- `.env` holds every provider API key and OAuth token in plaintext — including the ChatGPT **refresh token** the docs say to treat like a password,
- `openwiki.sqlite` holds full chat checkpoint state,
- `connectors/<id>/raw/` holds raw Gmail/Slack/Notion content.

On a default `%USERPROFILE%` the inherited ACL is usually owner+SYSTEM+Administrators, but nothing guarantees that: any broader ACE on the profile (or a relocated home) flows straight down to the credential store. Verified on a real Windows 11 machine — a directory created under `%TEMP%` inherited a dozen modify-rights ACEs from its parent, and the existing `chmod(0o700)` left all of them in place.

### Expected

On Windows, the existing 0o700 owner-only intent should be mirrored with an ACL: full control for the current user and SYSTEM only, inheritance removed — best-effort, exactly like the surrounding chmod calls (never failing a run because ACL tooling is unavailable).

I have a tightly-scoped, dependency-free fix (two `icacls` invocations, grant-before-inheritance-reset so a failed grant can never lock the user out) with tests ready — PR incoming.

WenyuChiou:
### What

Adds `restrictDirToCurrentUser` (`src/windows-acl.ts`) and wires it into the two places that express the home directory's 0o700 owner-only intent — `ensureOpenWikiHome` and `saveOpenWikiEnv`:

- On `win32`, runs `icacls <dir> /grant:r <user>:(OI)(CI)F *S-1-5-18:(OI)(CI)F`, then `icacls <dir> /inheritance:r` — full control for the current user and SYSTEM (well-known SID, language-independent), inheritable so newly created children (`.env`, `openwiki.sqlite`, `connectors/**`) are covered, inherited ACEs removed.
- **Fail-safe ordering:** the grant runs before the inheritance reset, so a failed grant can never lock the user out of the directory.
- **Best-effort by design:** returns `false` instead of throwing (mirroring the surrounding `chmodIfExists` semantics) — a machine without working `icacls` never fails a run.
- No new dependencies; no-op on non-Windows platforms, so POSIX behavior is completely unchanged.

### Why

On Windows, Node's `fs.chmod` only toggles the read-only attribute and the `mode` option on `mkdir`/`writeFile` is ignored — so the existing 0o700/0o600 calls protect nothing there, and `~/.openwiki` (plaintext `.env` with every provider key and the ChatGPT refresh token, sqlite chat checkpoints, raw connector dumps) simply inherits its parent's ACL.

Verified on a real Windows 11 machine (machine name redacted). A fresh directory before/after the helper:

```text
BEFORE (inherited from parent):
  ... a dozen inherited ACEs with (M) modify rights, including sandbox users ...
  NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
  BUILTIN\Administrators:(I)(OI)(CI)(F)
  MACHINE\user:(I)(OI)(CI)(F)

AFTER restrictDirToCurrentUser:
  NT AUTHORITY\SYSTEM:(OI)(CI)(F)
  MACHINE\user:(OI)(CI)(F)
```

Scope notes, stated explicitly: pre-existing files keep the ACEs they already inherited (no recursive `/T` pass — new children inherit the restricted ACL; a recursive apply on a large `connectors/raw` tree could be slow and is easy to add later if wanted). Administrators lose their inherited ACE on the directory but can still take ownership — the same way root bypasses 0o700 on POSIX.

### How tested

- New `test/windows-acl.test.ts` (3 tests, mocked `node:child_process`, runs on any platform incl. ubuntu CI): non-win32 no-op; exact grant + inheritance-reset argument sequence with grant ordered first; **grant failure short-circuits without removing inheritance** (the lockout guard).
- Real-machine integration check on Windows 11: `icacls` output before/after shown above.
- `pnpm run format`, `pnpm run lint`, `pnpm test`: 298/300 pass locally on Windows; the 2 failures are the pre-existing win32 `test/env-behavior.test.ts` platform checks (they reproduce on pristine `main`) — notably, this PR is the change that makes the *intent* of those 0600 expectations real on Windows at the ACL layer.

Fixes #366

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -149,8 +149,13 @@ jobs:
           exit-code: "1"
           ignore-unfixed: false
 
+      # Populates the Security tab. Fork PRs run with a read-only GITHUB_TOKEN
+      # that cannot write security events, so this upload fails there ("No
+      # server is currently available"). Skip it on fork PRs only; same-repo PRs
+      # and pushes to main still upload and stay strict. The real gate is Trivy's
+      # `exit-code: 1` above, which runs regardless.
       - name: Upload Trivy results to GitHub Security
-        if: always()
+        if: always() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork == false)
         uses: github/codeql-action/upload-sarif@v4
         with:
           sarif_file: trivy-results.sarif
--- a/src/env.ts
+++ b/src/env.ts
@@ -52,6 +52,7 @@ import {
   resolveProviderRetryAttempts,
 } from "./constants.js";
 import { isFileNotFoundError } from "./fs-errors.js";
+import { restrictDirToCurrentUser } from "./windows-acl.js";
 
 export const openWikiEnvDir = path.join(os.homedir(), ".openwiki");
 export const openWikiEnvPath = path.join(openWikiEnvDir, ".env");
@@ -208,6 +209,7 @@ export async function saveOpenWikiEnv(updates: EnvMap): Promise<void> {
     mode: 0o700,
   });
   await chmod(openWikiEnvDir, 0o700);
+  await restrictDirToCurrentUser(openWikiEnvDir);
 
   await writeFile(openWikiEnvPath, formatEnv(nextEnv), {
     encoding: "utf8",
--- a/src/openwiki-home.ts
+++ b/src/openwiki-home.ts
@@ -1,6 +1,7 @@
 import { chmod, mkdir } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
+import { restrictDirToCurrentUser } from "./windows-acl.js";
 
 export const openWikiHomeDir = path.join(os.homedir(), ".openwiki");
 export const openWikiConnectorsDir = path.join(openWikiHomeDir, "connectors");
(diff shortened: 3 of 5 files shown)


colifran (approved):
LGTM! Thanks @WenyuChiou!

PR #370, opened 2026-07-16
fix: keep scheduled OpenWiki updates opt-in on forks

### LINKED ISSUE #369: Scheduled OpenWiki updates run and fail on forks again after #179
## Summary

#302 reported that enabling GitHub Actions on a fork also enabled the scheduled `OpenWiki Update` workflow.

Contributor forks usually do not have an `OPENROUTER_API_KEY` repository secret. As a result, the scheduled job repeatedly failed even though the fork owner had not opted in to automated wiki updates.

#303 addressed this by restricting the update job to the upstream repository by default.

Fork owners could opt in by setting the `OPENWIKI_ENABLE_SCHEDULED_UPDATE` repository variable to `true`.

After #179 was merged, that guard disappeared from `.github/workflows/openwiki-update.yml`.

Scheduled OpenWiki updates can therefore run and fail on contributor forks again.

## Regression

#179 removed the condition introduced by #303:

```yaml
if: github.repository == 'langchain-ai/openwiki' || vars.OPENWIKI_ENABLE_SCHEDULED_UPDATE == 'true'
```

It also removed the accompanying explanation of the fork behavior.

I could not find any discussion in #179 about changing the scheduled-update policy.

Since that PR is scoped to adding the Google Vertex AI provider, this appears to be an unrelated regression introduced while the branch was rebased or the workflow was regenerated.

## Reproduction

Rather than waiting for the next cron invocation, I manually dispatched the same workflow on my fork:

https://github.com/jyje/openwiki/actions/runs/29509927346

The manual dispatch itself is expected behavior and is not the issue. It was used to reproduce the execution path immediately.

The run shows what happens when the shared `update` job executes on a contributor fork without an `OPENROUTER_API_KEY` secret:

- Checkout succeeded.
- Node.js setup succeeded.
- OpenWiki installation succeeded.
- `Run OpenWiki` failed.

A scheduled invocation uses the same job, environment, and steps.

Without a fork guard, the scheduled run follows the same failing path.

## Expected behavior

Scheduled updates should behave as follows:

- The upstream `langchain-ai/openwiki` repository runs the scheduled update normally.
- Forks do not run scheduled updates by default.
- Fork owners can opt in by setting `OPENWIKI_ENABLE_SCHEDULED_UPDATE=true`.
- Manually dispatched runs remain available, including on forks.

This keeps scheduled automation opt-in for fork owners without preventing intentional manual testing.

## Suggested direction

Restore the protection intended by #303 for scheduled runs while explicitly allowing `workflow_dispatch`.

For example:

```yaml
if: >
  github.event_name == 'workflow_dispatch' ||
  github.repository == 'langchain-ai/openwiki' ||
  vars.OPENWIKI_ENABLE_SCHEDULED_UPDATE == 'true'
```

This would result in:

| Trigger | Repository | Opt-in variable | Result |
| --- | --- | --- | --- |
| `schedule` | `langchain-ai/openwiki` | not required | Run |
| `schedule` | fork | not set | Skip |
| `schedule` | fork | `true` | Run |
| `workflow_dispatch` | upstream or fork | not required | Run |

There may also be an underlying regeneration issue to review separately.

`openwiki code --update` appears able to regenerate `.github/workflows/openwiki-update.yml` from the generic workflow template. Repository-specific workflow configuration should not be silently removed during that process.

The immediate scope of this issue is to restore safe scheduled behavior for forks while preserving intentional manual dispatch.

jyje:
Closes #369

## Summary

Restore the fork protection for scheduled OpenWiki updates that was introduced in #303 and removed by #179.

Scheduled runs now:

- Run normally in `langchain-ai/openwiki`.
- Skip on forks by default.
- Run on forks when `OPENWIKI_ENABLE_SCHEDULED_UPDATE=true`.

Manual `workflow_dispatch` runs remain available on both the upstream repository and forks.

## Why

Contributor forks usually do not have the `OPENROUTER_API_KEY` secret required by this workflow.

Without the guard, enabling Actions for contribution CI can also enable a daily OpenWiki update that repeatedly fails, even though the fork owner did not opt in to scheduled updates.

A manual run reproduced the execution path on my fork:

https://github.com/jyje/openwiki/actions/runs/29509927346

The manual dispatch itself is expected behavior. It demonstrates the failure that the scheduled job reaches on a fork without the required provider secret.

## Testing

- `actionlint .github/workflows/openwiki-update.yml`
- Prettier formatting check
- `git diff --check`

If explicitly allowing manual runs with `github.event_name == 'workflow_dispatch' ||` is not the preferred approach, please let me know and I will adjust it.

### DIFF
--- a/.github/workflows/openwiki-update.yml
+++ b/.github/workflows/openwiki-update.yml
@@ -11,6 +11,12 @@ permissions:
 
 jobs:
   update:
+    # Scheduled runs only fire in the origin repo by default, so contributor
+    # forks do not silently arm a daily job that requires a provider secret.
+    # Fork owners can opt in without editing this file by setting the
+    # OPENWIKI_ENABLE_SCHEDULED_UPDATE repository variable to "true". Manual
+    # dispatch remains available for intentional runs and testing.
+    if: github.event_name == 'workflow_dispatch' || github.repository == 'langchain-ai/openwiki' || vars.OPENWIKI_ENABLE_SCHEDULED_UPDATE == 'true'
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository


colifran (approved):
thanks @jyje! sorry that slipped through!

jyje:
No worries at all — it's totally fine. Thanks for the kind review and quick merge! :D

PR #371, opened 2026-07-16
feat: validate wiki internal links after generation

### LINKED ISSUE #358: Post-generation consistency pass: dead internal links/anchors and cross-page drift in regenerated wikis
## The gap

The wiki-as-memory loop this tool implements is canonically *ingest → query → lint → write back* — but openwiki currently ships without the lint step. The synthesis agent updates pages locally without re-verifying the wiki's cross-reference graph, so regenerations reliably ship a class of internal-consistency drift that review bots then catch one item per round:

- **Stale counts/enumerations**: a page opening still said "the five existing dimensions" two paragraphs above freshly regenerated content documenting the sixth; a "planned but not yet authored" list contradicted by the same run's own additions.
- **Dead internal anchors**: a heading edited during regeneration left links in sibling pages pointing at the old anchor.
- **Repo-state claims that drift from the committed tree**: generated pages describing files that exist only in the working tree (e.g., the generator's own uncommitted CI scaffold), which reads as broken guidance to anyone at the committed tree.

Observed across several 0.1.2 code-mode wikis (both `--update` and `--init` paths) and equally applicable to personal-brain wikis.

## Proposal

Two tiers, cheapest first:

1. **Deterministic validator (no LLM)** — after generation completes, before `.last-update.json` is written and before the generated CI workflow opens its PR: resolve every relative `[text](path.md#anchor)` link against files that exist and anchors GitHub actually generates (lowercase → strip non `[\w\- ]` → each space becomes one hyphen, so `"a + b"` → `a--b`; duplicate headings get `-1`, `-2` suffixes; root-relative `/path.md` resolves from the repo root). ~60 lines of stdlib Python/TS; fail the run or surface the list to the agent for a self-repair turn.

2. **LLM lint pass (the missing loop step)** — a cheap final turn over the diff'd pages asking specifically for cross-page contradictions the local edits introduced: counts/enumerations that no longer match, claims about repo state, "planned" lists contradicted by shipped content. This is much cheaper as a dedicated pass than as N rounds of external review-bot findings.

Happy to contribute the tier-1 validator as a PR if there's interest — we're running exactly this as a post-generation gate in our repos' openwiki CI workflows and it took the review-round churn to zero.

DecentralizedJM:
## Summary
- Add a deterministic post-generation validator that checks relative markdown links and GitHub-style heading anchors across the generated wiki.
- Fail init/update runs when broken internal links are detected, before `.last-update.json` is written.
- Include regression tests plus a dogfood check against this repo's checked-in `openwiki/` tree.

Fixes #358

## Test plan
- [x] `pnpm test` (341 tests pass)
- [x] `pnpm run lint:check`
- [x] Dogfood test validates the repository's own `openwiki/` links

### DIFF
--- a/src/agent/wiki-link-validator.ts
+++ b/src/agent/wiki-link-validator.ts
@@ -1,0 +1,398 @@
+import type { BackendProtocolV2, FileInfo } from "deepagents";
+import path from "node:path";
+import type { OpenWikiOutputMode } from "./types.js";
+
+/**
+ * Reserved or control files that never carry agent-authored concept links.
+ */
+const EXCLUDED_FILES = new Set([
+  "index.md",
+  "log.md",
+  "_plan.md",
+  "INSTRUCTIONS.md",
+]);
+
+const MARKDOWN_LINK_PATTERN = /\[([^\]]*)\]\(([^)]+)\)/gu;
+const HEADING_PATTERN = /^(#{1,6})\s+(.+?)\s*#*\s*$/u;
+const BROKEN_LINK_STAMP_PATTERN =
+  /^\s*<!--\s*openwiki:\s*broken internal link\b.*?-->\s*$/u;
+
+export interface WikiLinkIssue {
+  href: string;
+  line: number;
+  message: string;
+  sourcePath: string;
+}
+
+/**
+ * Summary of one internal-link validation pass over a generated wiki.
+ */
+export interface WikiLinkReport {
+  /**
+   * How many Markdown files were scanned.
+   */
+  filesScanned: number;
+
+  /**
+   * How many relative internal links were checked.
+   */
+  linksChecked: number;
+
+  /**
+   * How many broken links were found (and stamped).
+   */
+  issuesFound: number;
+
+  /**
+   * Wiki-root-relative paths of files that were rewritten with stamps.
+   */
+  stampedFiles: string[];
+}
+
+/**
+ * Validates relative wiki links and GitHub-style heading anchors after
+ * generation, stamping broken links in place instead of failing the run.
+ *
+ * Each broken link is preceded by an HTML comment so a later update run can
+ * find it inline and repair the href. Existing stamps are cleared first, so a
+ * fixed link leaves no residual comment.
+ */
+export async function validateWikiInternalLinks(
+  backend: BackendProtocolV2,
+  outputMode: OpenWikiOutputMode,
+): Promise<WikiLinkReport> {
+  const wikiRoot = outputMode === "local-wiki" ? "/" : "/openwiki";
+  const report: WikiLinkReport = {
+    filesScanned: 0,
+    linksChecked: 0,
+    issuesFound: 0,
+    stampedFiles: [],
+  };
+
+  for (const sourcePath of await collectMarkdownFiles(backend, wikiRoot)) {
+    report.filesScanned += 1;
+    const original = await readText(backend, sourcePath);
+    const cleaned = stripBrokenLinkStamps(original);
+    const headingAnchors = buildHeadingAnchors(extractHeadings(cleaned));
+    const issues: WikiLinkIssue[] = [];
+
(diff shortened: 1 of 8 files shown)


src/agent/index-middleware.ts
```diff
@@ -40,6 +44,10 @@ export function createOpenWikiIndexMiddleware(
       ),
     afterAgent: async () => {
       await synchronizeWikiIndexes(backend, outputMode);
+      const linkIssues = await validateWikiInternalLinks(backend, outputMode);
+      if (linkIssues.length > 0) {
+        throw new Error(formatWikiLinkIssues(linkIssues));
```
colifran: i like the general concept - my concern is that this would crash the users run. i think a better approach is to find a way to gracefully flag these in a way that a future run can make the agent aware of the invalid links.

for reference take a look at some of the work i did in #199. instead of failing on a bad mermaid diagram i degraded to text with a stamp that lets the agent self-correct an a future update run:
```
<!-- openwiki: mermaid parse failed and this diagram was converted to a text fence so it does not break rendering. Fix the diagram source and restore the mermaid fence. Parser error: Parse error on line 5: ...leted]  notify -> end[Response] -->
```

thoughts on that?

colifran (commented):
thanks for the PR! i think this is really useful. my immediate concern is about crashing on invalid links. take a look at my comment and let me know your thoughts!

DecentralizedJM:
Thanks for the review — agreed that failing the run was too harsh.

Updated to follow the mermaid pattern from #199:

- Broken internal links/anchors are left in place and stamped with an HTML comment:
  `<!-- openwiki: broken internal link [...] ... Fix the href or restore the target, then delete this comment. -->`
- The init/update run completes normally (no throw).
- Prompt guidance teaches the agent to find those stamps and self-correct on a later update.
- Stamps are cleared automatically once the link validates again.

Also rebased onto current `main` and wired through `okf-middleware` after the mermaid + index passes.

src/agent/index-middleware.ts
DecentralizedJM: Thanks — that makes sense. Switched away from throwing: broken links are stamped inline with `<!-- openwiki: broken internal link ... -->` (same self-repair idea as the mermaid degrade path in #199), the run completes, and the prompt now teaches the agent to fix those on a later update.

src/agent/wiki-link-validator.ts
```diff
@@ -0,0 +1,398 @@
+import type { BackendProtocolV2, FileInfo } from "deepagents";
```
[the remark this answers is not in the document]
DecentralizedJM: Thanks @corridor-security — valid finding.

Updated in the latest push:

1. **Containment** — after resolving a link, we now require the path to stay under the wiki root (`isPathUnderWikiRoot` via `path.relative`). Escaping hrefs (e.g. `../../../../etc/passwd`) are stamped as broken **without** calling `readRaw`/`ls` outside `/openwiki`.
2. **No path disclosure in stamps** — missing-anchor messages cite the original `linkPath` / href instead of the resolved absolute target.

Added a regression test that places a readable file outside the wiki root and asserts we neither read it nor leak its resolved path in the stamp.

cc @colifran

colifran:
@DecentralizedJM sorry for the delay. reviewing this one today!

DecentralizedJM:
Thanks @colifran — no worries on the timing!

Just rebased onto latest `main` and resolved the merge conflicts in `okf-middleware.ts` / `prompt.ts` (link validation now runs inside the existing `inStage("finalize", ...)` passes alongside mermaid + index sync).

Current state for review:
- Broken links are **stamped** inline (mermaid degrade pattern from #199), not thrown.
- **Wiki-root containment** added per corridor-security feedback — escaping hrefs are stamped without reading outside `/openwiki`.
- `pnpm test` passes locally (896 tests).

Should be mergeable once CI finishes — let me know if anything else looks off.

colifran (approved):
Thanks for your work on this @DecentralizedJM!

Issue #84, opened 2026-07-04
Add optional Open Knowledge Format (OKF) output


manu-xm:
Problem

OpenWiki generates structured, multi-file repository documentation, but its output currently follows OpenWiki-specific conventions. This limits portability: external agents, knowledge catalogs, search tools, and developer platforms must build custom logic to understand page types, metadata, navigation, and relationships.

As OpenWiki wikis grow, this also makes structured discovery and selective context retrieval harder across tools.

Why OKF would benefit OpenWiki

Adding optional Open Knowledge Format support would:

* Make OpenWiki documentation portable across agent and knowledge ecosystems.
* Reduce custom integration work for downstream consumers.
* Add consistent metadata for type, title, resource, tags, and update time.
* Improve discovery and retrieval of relevant repository subcontext.
* Allow OpenWiki output to connect with broader knowledge such as APIs, runbooks, architecture, and data documentation.
* Preserve OpenWiki’s existing Git-native, Markdown-based workflow without introducing a proprietary platform or runtime.
* Position OpenWiki as an early repository-focused producer in the emerging OKF ecosystem.

Proposal

Add optional OKF v0.1 output support, including:

* YAML frontmatter for generated concept pages.
* A stable repository-documentation type taxonomy.
* A root index.md for navigation and version declaration.
* Metadata preservation during incremental updates.
* Basic OKF validation.
* Backward-compatible opt-in configuration.

OpenWiki would continue to generate and maintain repository knowledge; OKF would standardize how that knowledge is packaged and consumed by other tools.

This is likely the right length for a GitHub issue: enough substance to justify the feature without overwhelming maintainers.

acgxv:
Came from OpenKB, and they have already supported OKF.
Love to see this in OpenWiki.

colifran:
Love this. Thank you for the suggestion. We're looking into this.

ayrtonvwf:
@colifran I have been building okf support into a fork of mine in hope that this helps to advance the initiative, would you take a look?

https://github.com/ayrtonvwf/openwiki

I did it step by step with issues and PRs because this is not a small change, including specs (that can be striped off as the original repo does not seem to use SDD)

I'd be happy to contribute by shaping my fork into something mergeable back into openwiki

pikujs:
OKF Compliance would be really useful.
Do let us know if any help needed to get this merged

gaborho:
@colifran can you please share some details, if there is any plan to include OKF output format support in openwiki?

[...]

PR #373, opened 2026-07-17
fix: align okf output with google v0.1

ruanbarroso:
## Summary

Follow-up to #345 and #84 that aligns OpenWiki with the canonical [Google OKF v0.1 specification](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md).

- accept the optional `timestamp` field and producer-defined extension fields
- require only a non-empty `type` for concepts
- treat `index.md` and `log.md` as reserved documents, not concepts
- emit `okf_version: "0.1"` only in the root index and no frontmatter in nested indexes
- preserve extension fields in agent and migration guidance
- remove contradictory prompt requirements for `title`, `description`, and `tags`
- document the supported OKF behavior

## Tests

- `pnpm run format:check`
- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm run test` — 338 tests passed across 32 files
- `pnpm run build`
- strict validation of `openwiki/` with an independent OKF v0.1 validator — 7 concepts, 0 errors, 0 warnings
- generated end-to-end fixture — 2 concepts, 0 errors, 0 warnings


src/agent/frontmatter-validator.ts
```diff
@@ -5,8 +5,14 @@ import { parse } from "yaml";
 import { MUTATION_PATH_METADATA_KEY } from "./docs-only-backend.js";
 import type { OpenWikiOutputMode } from "./types.js";
 
-const OKF_STRING_FIELDS = ["type", "title", "description", "resource"];
-const OKF_FIELDS = new Set([...OKF_STRING_FIELDS, "tags"]);
+const OKF_STRING_FIELDS = [
```
colifran, line 8: are we losing tags here?

src/agent/index-middleware.ts
```diff
@@ -6,7 +6,13 @@ import { addFrontmatterWarning } from "./frontmatter-validator.js";
 import type { OpenWikiOutputMode } from "./types.js";
 
 const INDEX_FILE = "index.md";
-const EXCLUDED_FILES = new Set([INDEX_FILE, "_plan.md", "INSTRUCTIONS.md"]);
+const LOG_FILE = "log.md";
+const EXCLUDED_FILES = new Set([
+  INDEX_FILE,
+  LOG_FILE,
```
colifran, line 12: nice - good catch.

src/agent/index-middleware.ts
```diff
@@ -135,17 +137,18 @@ async function synchronizeDirectory(
...
+  isRoot: boolean,
 ): string {
   const sections = [
     renderLinks("Files", files, true),
     renderLinks("Directories", directories, false),
   ]
     .filter(Boolean)
     .join("\n\n");
-  return `---\ntype: Documentation Index\ntitle: ${JSON.stringify(title)}\ndescription: ${JSON.stringify(`Files and subdirectories in ${title}.`)}\n---\n\n${sections}\n`;
+  const version = isRoot ? '---\nokf_version: "0.1"\n---\n\n' : "";
```
colifran, line 150: also a great catch.

colifran (commented):
Looks great and thank you! I have one question about tags before we merge.

src/agent/frontmatter-validator.ts
ruanbarroso: No, `tags` is still validated — it just isn't part of the `OKF_STRING_FIELDS` list because it's a list, not a string. It has its own dedicated check further down (unchanged by this PR):

```ts
if (
  Object.hasOwn(fields, "tags") &&
  (!Array.isArray(fields.tags) ||
    fields.tags.some((tag) => typeof tag !== "string" || !tag.trim()))
) {
  issues.push(issue("invalid_tags", "Field `tags` must be a YAML list of non-empty strings."));
}
```

What went away is the old `OKF_FIELDS` allowlist that produced the `unsupported_field` error. That removal is intentional: OKF v0.1 permits producer-defined extension fields, so we can no longer reject unknown keys.

One conscious trade-off worth flagging: without the allowlist, a typo in a standard field (e.g. `descriptoin:`) now passes silently as an "extension" instead of erroring. Erroring would violate the spec, so I left it as-is — happy to open a follow-up issue for a heuristic "did you mean" warning if you think it's worth it.

colifran (approved):
Thank you for this @ruanbarroso!

PR #376, opened 2026-07-17
fix: tolerate optional OKF index metadata

Eijnewgnaw:
## Summary

- tolerate non-string or blank optional `title` and `description` values while generating directory indexes
- fall back to the filename for an unusable title and omit an unusable description
- add regression coverage for scalar, list, and mapping metadata values

## Why

OKF v0.1 requires a parseable frontmatter block and non-empty `type`; optional display metadata should not prevent a consumer from indexing the rest of a bundle. Index generation now uses only optional values that can be rendered safely.

## Validation

- `pnpm exec prettier --check src/agent/index-middleware.ts test/index-middleware.test.ts`
- `pnpm exec vitest run test/index-middleware.test.ts`
- `pnpm run typecheck`

Related to #84.


src/agent/index-middleware.ts
```diff
@@ -191,21 +191,20 @@ function parseFrontmatter(
...
-  if (
-    description !== undefined &&
-    (typeof description !== "string" || !description.trim())
-  ) {
-    throw new Error(`${filePath} YAML description must be a non-empty string.`);
-  }
-  if (title !== undefined && typeof title !== "string") {
-    throw new Error(`${filePath} YAML title must be a string.`);
-  }
+  const usableDescription = usableString(description);
```
colifran, line 194: nice - i like this. makes sense to not fail here.

colifran (approved):

PR #382, opened 2026-07-17
Add claude-cli provider write guard

deanjstone:
## Summary
- First piece of the `claude-cli` provider design: since a bare `claude -p` subprocess can't be handed a DeepAgents `LocalShellBackend` subclass (the mechanism `docs-only-backend.ts` uses today), this adds the equivalent write restriction as an out-of-band Claude Code `PreToolUse` hook.
- `src/agent/claude-cli/write-guard.ts` — pure, unit-tested path decision logic (`evaluateWritePath`), covering directory containment, prefix-collision (`openwiki-fake` vs `openwiki`), traversal, and repo-boundary checks.
- `src/agent/claude-cli/write-guard-hook.ts` — thin CLI entrypoint invoked by Claude Code itself: parses the PreToolUse stdin payload, only inspects `Write`/`Edit` calls, and exits 2 (block, stderr fed back to the model) for any file path outside `CLAUDE_CLI_ALLOWED_DIR` (default `openwiki`) under `CLAUDE_CLI_REPO_ROOT`. Fails closed on missing config or malformed input.

Not wired into the agent runtime yet — the `claude-cli-backend.ts` module that spawns `claude -p` with this hook configured via a generated `.claude/settings.json` is a follow-up PR.

## Test plan
- [x] `pnpm test` — 340/340 passing (8 new tests in `test/write-guard.test.ts`)
- [x] `pnpm run typecheck` / `pnpm run lint:check` — clean
- [x] `pnpm run build`, then manually piped real PreToolUse-shaped JSON at the compiled hook for an allow case (write inside `openwiki/`), a deny case (write to `AGENTS.md`), and a non-matching tool case (`Bash`) — exit codes 0/2/0 as expected

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,4 +1,25 @@
-## OpenWiki
+# argus-wiki
+
+Light-touch fork of [langchain-ai/openwiki](https://github.com/langchain-ai/openwiki), retargeted at the CONTEXT/CORTEX Obsidian vaults instead of a fresh `~/.openwiki` tree.
+
+## Remotes
+
+- `origin` → `deanjstone/argus-wiki` (this fork)
+- `upstream` → `langchain-ai/openwiki` (pull periodically: `git fetch upstream && git merge upstream/main`)
+
+## Local divergence from upstream
+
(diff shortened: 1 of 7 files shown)


deanjstone:
Opened against the wrong repo by mistake (meant to target my fork, deanjstone/argus-wiki, not upstream). Closing — apologies for the noise.

PR #392, opened 2026-07-18
fix: Harden skill writing

bracesproul:
WIP - i still need to review this code

### DIFF
--- a/src/agent/skills.ts
+++ b/src/agent/skills.ts
@@ -1,4 +1,6 @@
-import { cp, readdir, rm } from "node:fs/promises";
+import { cp, mkdir, readdir, rename, rm, stat } from "node:fs/promises";
+import { randomUUID } from "node:crypto";
+import { setTimeout as delay } from "node:timers/promises";
 import { fileURLToPath } from "node:url";
 import path from "node:path";
 import { ensureOpenWikiHome, openWikiSkillsDir } from "../openwiki-home.js";
@@ -6,6 +8,10 @@ import { ensureOpenWikiHome, openWikiSkillsDir } from "../openwiki-home.js";
 const bundledSkillsDir = fileURLToPath(
   new URL("../../skills", import.meta.url),
 );
+const SKILL_SYNC_LOCK_NAME = ".openwiki-skill-sync.lock";
+const SKILL_SYNC_LOCK_RETRY_MS = 25;
+const SKILL_SYNC_LOCK_TIMEOUT_MS = 10_000;
+const SKILL_SYNC_LOCK_STALE_MS = 5 * 60_000;
 
 /** Copies bundled skills into the OpenWiki home while preserving other skills. */
 export async function syncBundledSkills(): Promise<void> {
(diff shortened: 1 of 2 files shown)


mherzog4:
This fixes #489, which is currently biting people on 0.2.3 (`latest`) — worth linking the two, since the issue has an independent root-cause analysis that lands on exactly this code and nobody there seems aware this PR exists.

I tested the branch logic against the reproduction from that issue. **It works, including the cross-process case**, and I found one edge worth fixing before merge.

## Confirmed fixed

Two overlapping `replaceSkillDirectories()` passes into a shared target, using the real bundled skill dirs:

| | unpatched | this PR |
|---|---|---|
| 2 overlapping passes, 30 iterations | **30/30 threw** | **0/30** |
| 4 concurrent *processes*, 8 rounds | — | **0/8 rounds failed** (32/32 processes ok) |

The cross-process result is the important one and is why the lock is the right call over the module-level promise memoisation suggested in #489 — I had applied memoisation locally and it does stop the crash, but it cannot help `src/schedules.ts` running `openwiki ingest all --scheduled --print` from launchd against the same `~/.openwiki`.

Environment: Node v23.10.0, macOS 26 (Apple Silicon), logic extracted verbatim from this diff.

## One issue: an abandoned lock hard-fails runs for 5 minutes

`withSkillSyncLock` removes the lock in a `finally`, which does not run if the process is interrupted. `SKILL_SYNC_LOCK_STALE_MS` is 5 minutes but `SKILL_SYNC_LOCK_TIMEOUT_MS` is 10 seconds, so a lock left by a dead holder is not yet stale enough to reclaim and every run in that window blocks the full timeout and then throws.

Reachable with plain Ctrl-C, not just `SIGKILL`:

```
$ node slow.mjs &          # acquires the lock, then sleeps
$ kill -INT $!             # ordinary Ctrl-C
after SIGINT, lock still present? YES

$ node worker.mjs /tmp/...
FAIL Timed out waiting for the bundled skill sync lock at /tmp/.../.openwiki-skill-sync.lock.
next run: 10s
```

So every `openwiki` invocation for the next 5 minutes stalls 10 seconds and dies with a lock-timeout error. That trades a coin-flip crash for a deterministic 5-minute outage, and it is easy to hit precisely because interrupting a long agent run is normal usage.

The `recovers an abandoned skill sync lock` test does not catch this — it backdates the lock past the stale threshold with `utimes`, which is the *already-recovered* state rather than the window that hurts.

Options, in rough order of how much I like them:

1. **Record the holder pid in the lock** (e.g. a `pid` file inside the lock dir) and treat the lock as stale when that pid is gone, regardless of age. Deterministic rather than time-based, and collapses the bad window to zero.
2. **Refresh the lock mtime while held** and drop `STALE` to a few seconds — a live holder keeps it fresh, so an abandoned lock is reclaimable almost immediately.
3. **Release on signals** (`SIGINT`/`SIGTERM` handlers). Helps the common case but still leaves `SIGKILL` and hard crashes, so it wants one of the above underneath.

Happy to push a commit for whichever you prefer, or leave it to you since the PR is marked WIP.

## Minor

`await mkdir(targetDir, { recursive: true })` before the lock omits `mode: 0o700`, where `src/openwiki-home.ts` creates `~/.openwiki` and `~/.openwiki/skills` as `0700`. In practice `ensureOpenWikiHome()` has usually created it already so this rarely decides the mode, but if `replaceSkillDirectories` is ever called first it would create the skills dir with the default umask instead. Passing the mode here would make it order-independent.

colifran:
Resolved with #502

Issue #368, opened 2026-07-16
openwiki --update in GitHub Actions reports false "empty checkout" and produces hallucinated output instead of real doc updates


jhonurrego-tekton:
### Summary

Running `openwiki --update --print` (OpenAI provider) inside a GitHub Actions workflow produced two bad runs back to back on the same commit: one that silently no-op'd while claiming the repository was empty, and an immediate re-run that opened a PR with fabricated content and a log narrative that doesn't match the actual repository state.

### Environment

- `openwiki@0.1.2` (`npm install --global openwiki@0.1.2`)
- GitHub-hosted `ubuntu-latest` runner, Node 22 requested via `actions/setup-node@v4` (runner auto-forced these actions onto Node 24 per the Sept 2025 Node 20 deprecation)
- Provider: `OPENWIKI_PROVIDER=openai`, `OPENWIKI_MODEL_ID=gpt-5.4-mini`, `OPENAI_API_KEY` set via GitHub Actions secret
- Workflow follows the structure of `examples/openwiki-update.yml`, `openwiki --update --print` on `workflow_dispatch` + a daily `schedule`
- `openwiki/` already existed in the repo from a prior successful `--update` run (with real generated docs and a valid `.last-update.json`)

### What happened

**Run 1 (scheduled trigger)** — job reported success, but the "Run OpenWiki" step's full output was:

> No repository-local `openwiki/` documentation exists in this checkout, so there's nothing to surgically update.
>
> I also couldn't verify the claimed workflow change from the source tree because the repository path provided appears empty in this runtime. If you want, I can still help by:
> - initializing the wiki from the repository once the files are available, or
> - updating the docs after you point me at the correct checkout path.

This is false — `openwiki/` was present and committed on the checked-out branch (added by a previous `--update` run days earlier). Because our post-run check only verified that the `openwiki/` directory exists (not that anything changed), this failure was invisible in CI status — the job showed green.

**Run 2 (manual re-run immediately after, same commit, same config)** — this time it opened a pull request, but the diff only:
- bumped the timestamp in `openwiki/.last-update.json`
- added a stray `openwiki/_plan.md` (~10 lines, scratch/planning content, not documentation)

The step's printed summary for this run was also inaccurate about the repo it was supposedly reading:

> No OpenWiki docs were updated.
>
> What I found:
> - The repository has no checked-in `openwiki/` documentation tree to refresh.
> - The only relevant source change is `.github/workflows/openwiki-update.yml`.
[...]

PR #393, opened 2026-07-19
fix: point repository runtime note at openwiki docs

HwangJohn:
## Summary

Updates the repository-mode runtime note appended to agent run messages so it points the agent at the repo-local `/openwiki` docs instead of the personal wiki path `~/.openwiki/wiki`.

## Why

PR #383 fixed the shared system prompt blocks, but `src/agent/index.ts` still appended a per-run runtime note that said the canonical generated wiki was `~/.openwiki/wiki` and not repository-local `openwiki/`. In repository runs that contradicts the actual virtual filesystem layout and can push the agent toward the wrong path.

Related to #368.

## Testing

- `corepack pnpm exec vitest run test/agent-runtime-root.test.ts test/prompt.test.ts`
- `corepack pnpm run format`
- `corepack pnpm run lint`

I also ran `corepack pnpm test`; it still fails on this Windows checkout in `test/env-behavior.test.ts` for the same 3 HOME/permission isolation assertions that also fail on a clean `origin/main@ac5007c` worktree.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -393,12 +393,14 @@ function formatRuntimeRootLabel(outputMode: OpenWikiOutputMode): string {
   return outputMode === "local-wiki" ? "Local wiki root" : "Repository root";
 }
 
-function formatRuntimeRootInstruction(outputMode: OpenWikiOutputMode): string {
+export function formatRuntimeRootInstruction(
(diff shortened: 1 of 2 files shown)


colifran (approved):
Nice one. Thanks @HwangJohn!

PR #394, opened 2026-07-19
fix: clean up temporary openwiki plan files

HwangJohn:
## Summary

Adds host-side cleanup for the temporary `_plan.md` file that OpenWiki asks the agent to create during init/update runs.

## Why

The prompt already tells the agent to delete `_plan.md` before finishing, but issue #368 shows that a run can still leave `openwiki/_plan.md` behind and include it in the generated PR. Since the plan is scratch state rather than final documentation, the runtime should clean it up deterministically.

Related to #368.

## What changed

- Removes the mode-specific temporary plan file after init/update streams finish, before metadata persistence.
- Attempts the same cleanup before late-failure metadata persistence without masking the original run error.
- Excludes the root `_plan.md` from OpenWiki content snapshots so a leftover temporary plan alone does not look like a documentation change.

## Testing

- `corepack pnpm exec vitest run test/run-metadata.test.ts test/update-noop.test.ts --no-file-parallelism`
- `corepack pnpm run format`
- `corepack pnpm run lint`

I also ran `corepack pnpm test`; it still fails on this Windows checkout in `test/env-behavior.test.ts` for the same 3 HOME/permission isolation assertions that also fail on a clean `origin/main@ac5007c` worktree.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -89,6 +89,7 @@ import {
   getUpdateNoopStatus,
   createRunContext,
   persistRunMetadataIfChanged,
+  removeTemporaryPlanFile,
   shouldCheckUpdateNoop,
 } from "./utils.js";
 import { classifyError, recordRunSafe } from "../telemetry/index.js";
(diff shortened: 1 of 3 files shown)


colifran (approved):
Thanks @HwangJohn! Aligned on making this deterministic. Possibly a follow-up to adjust the prompt is needed.

PR #398, opened 2026-07-19
fix: harden oauth discovery endpoints

jkennedyvz:
## Summary
- add shared OAuth endpoint validation for discovery, registration, authorization, and token URLs
- require HTTPS, block localhost/private/link-local targets, reject credentials, and disable redirects for OAuth fetches
- constrain Notion MCP OAuth discovery endpoints to `notion.com` hosts
- pin the OpenWiki update workflow actions (and generated/example workflow templates) to immutable SHAs

## Corridor findings addressed
- a9526bc7-571d-4c8a-a7f8-d4ce62aa0fa4 — workflow action pinning
- 46b05f6b-df52-474b-b84e-65f4dc3855c4 — HTTPS enforcement on MCP OAuth endpoints
- 2fc3df4d-cb8b-497c-9405-1ecf4fe0c559 — SSRF / refresh-token exfiltration via discovered URLs

## Verification
- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm test`


src/auth/oauth-discovery.ts
```diff
@@ -0,0 +1,176 @@
+import { isIP } from "node:net";
+
+export type OAuthMetadata = {
```
colifran, line 3: nit: not block but would love for these to be interfaces. i know convention in the codebase is mixed but we're standardizing on interfaces which is more correct for these types of shapes. we'll do a refactor later though!

colifran (approved):
Looks good. Thanks @jkennedyvz!

PR #401, opened 2026-07-20
fix: harden raw connector file handling

HwangJohn:
## Summary
- Normalize raw connector file listings to POSIX separators before newest-run sorting and latest-run filtering.
- Reject raw item reads when any path component is a symlink before opening the file.
- Preserve `O_NOFOLLOW` where the platform exposes it, while keeping normal raw file reads and truncation behavior unchanged.

## Testing
- `corepack pnpm run format`
- `corepack pnpm run lint`
- `corepack pnpm exec prettier --check src/connectors/tools.ts test/raw-connector-tools.test.ts`
- `corepack pnpm exec vitest run test/raw-connector-tools.test.ts`
- `corepack pnpm test` *(fails on this Windows checkout only in the existing `test/env-behavior.test.ts` baseline cases: temp HOME path resolution, 0600 mode bits, and absent saved key handling)*

No linked issue; this came from a focused connector hardening sweep.


colifran (approved):
Thanks @HwangJohn!

PR #402, opened 2026-07-20
fix: tolerate malformed slack and x array config

HwangJohn:
## Summary
- Treat Slack and X connector array config fields as untrusted input from JSON config.
- Normalize malformed non-array `streams` values back to connector defaults instead of throwing.
- Normalize Slack `conversationTypes` to defaults when malformed, and normalize Slack `assistantSearchQueries` plus X `listIds` to clean string arrays.
- Add regression coverage alongside the existing connector config override tests.

## Testing
- `corepack pnpm run format`
- `corepack pnpm run lint`
- `corepack pnpm exec prettier --check src/connectors/sources/slack.ts src/connectors/sources/x.ts test/connector-config-overrides.test.ts`
- `corepack pnpm exec vitest run test/connector-config-overrides.test.ts`
- `corepack pnpm test` *(fails on this Windows checkout only in the existing `test/env-behavior.test.ts` baseline cases: temp HOME path resolution, 0600 mode bits, and absent saved key handling)*

No linked issue; this follows the same malformed-array hardening pattern as the Gmail connector.


src/connectors/sources/x.ts
```diff
@@ -282,17 +283,37 @@ function getDefaultTweetParams(): Record<string, string> {
...
+    Array.isArray(requested) && requested.length > 0
+      ? requested
+      : DEFAULT_STREAMS;
 
   return streams.filter(isXStream);
 }
 
-function isXStream(value: string): value is XStream {
-  return (DEFAULT_STREAMS as readonly string[]).includes(value);
+function normalizeStringArray(value: unknown): string[] {
```
colifran: we could probably pull this into a common utils file somewhere?

colifran (approved):
Pushed a commit to consolidate normalizeStringArray. Thanks for your help with these!

PR #406, opened 2026-07-20
fix: reject ".." traversal in docs-only write guard

### LINKED ISSUE #405: [Bug]: docs-only write guard permits ".." traversal out of /openwiki/
### Description

`isOpenWikiDocsPath` in `src/agent/docs-only-backend.ts` is the sole enforcement of the
"init/update runs may only write under `/openwiki/`" rule. It confines writes with a raw
string-prefix check (`virtualPath.startsWith("openwiki/")`) on a path that is **never
path-normalized**, so `..` segments are not collapsed before the check. A crafted path such as
`/openwiki/../AGENTS.md` passes the guard, and once `LocalShellBackend.write` resolves the path it
lands on a repo file **outside** `openwiki/`.

The system prompt notes that connector/MCP page bodies are untrusted and can influence a write path,
so this boundary is reachable from untrusted input.

### Steps to Reproduce

1. Create a docs-only backend: `new OpenWikiLocalShellBackend({ docsOnly: true, rootDir, virtualMode: true })`.
2. Call `backend.write("/openwiki/../AGENTS.md", "…")`.
3. Observe the write is permitted and lands at `<rootDir>/AGENTS.md`, outside `openwiki/`.

Unit level: `isOpenWikiDocsPath("/openwiki/../AGENTS.md")` returns `true`.

### Expected Behavior

Paths that resolve outside `/openwiki/` are refused — `isOpenWikiDocsPath(...) === false` and the
guard returns its `Refused path:` error.

### Actual Behavior

`virtualPath = "openwiki/../AGENTS.md"`; `.startsWith("openwiki/")` is `true`, so the write is
permitted. `super.write` then collapses `..` to `<rootDir>/AGENTS.md`, a file outside `openwiki/`.

### Environment

- OS: macOS / Linux
- Node.js version: 22+
- OpenWiki version: `main` @ e2a7ca5 (0.2.0)

### Additional Context

The sibling `isWikiMarkdownPath` in `src/agent/frontmatter-validator.ts` already calls
`path.posix.normalize(...)` before the same class of check; `isOpenWikiDocsPath` omits it. The fix
is to normalize with `path.posix.normalize` before the prefix comparison. I have a fix + regression
test ready and will open a PR referencing this issue.

thejesh23:
## What & why

`isOpenWikiDocsPath` (`src/agent/docs-only-backend.ts`) is the sole enforcement of the "init/update
runs may only write under `/openwiki/`" rule. It compared a raw, **un-normalized** path with a
string prefix, so `..` segments were never collapsed before the check. A crafted path like
`/openwiki/../AGENTS.md` passed the guard (`"openwiki/../AGENTS.md".startsWith("openwiki/")` is
`true`), and once `LocalShellBackend.write` resolved the path it landed on a repo file **outside**
`openwiki/`. Since the system prompt treats connector/MCP page bodies as untrusted and able to
influence a write path, this confinement is reachable from untrusted input.

The fix normalizes the path with `path.posix.normalize(...)` before the prefix comparison — the same
approach the sibling `isWikiMarkdownPath` guard in `src/agent/frontmatter-validator.ts` already uses.
A `..` that resolves back inside `openwiki/` is still allowed; one that escapes is now refused.

## How it was tested

- Extended `test/docs-only-backend.test.ts`:
  - unit: `isOpenWikiDocsPath("/openwiki/../AGENTS.md")` and other escaping variants now return
    `false`; a `..` resolving back inside `openwiki/` stays `true`.
  - backend: `backend.write("/openwiki/../AGENTS.md", …)` is refused and no file is written outside
    `openwiki/`.
- `pnpm run format:check`, `pnpm run lint:check`, `pnpm run typecheck`, `pnpm test` all pass
  (415 tests).

Fixes #405


colifran (approved):
Looks good - thank you!

PR #408, opened 2026-07-20
fix: write ~/.openwiki/.env atomically so a failed write keeps existing tokens

### LINKED ISSUE #407: [Bug]: non-atomic ~/.openwiki/.env write wipes all saved tokens on a failed write
### Description

`saveOpenWikiEnv` in `src/env.ts` persists newly obtained/refreshed OAuth tokens and API keys to
`~/.openwiki/.env` with a plain `writeFile(openWikiEnvPath, …)`. `writeFile` opens the existing
credential file with `O_TRUNC` — it truncates the file to zero **before** writing the new contents.
The write is not staged to a temp file and `rename`d into place, so any failure mid-write (ENOSPC,
crash, power loss) leaves `~/.openwiki/.env` truncated/empty and **every** saved token/key is lost.

### Steps to Reproduce

1. Have valid credentials saved, e.g. `OPENWIKI_GMAIL_REFRESH_TOKEN`,
   `OPENWIKI_SLACK_USER_REFRESH_TOKEN`, `OPENAI_API_KEY` in `~/.openwiki/.env`.
2. Trigger a save that fails mid-write — e.g. a token refresh that calls `saveOpenWikiEnv` while the
   disk is full (or the process is killed during the write).
3. Inspect `~/.openwiki/.env`.

### Expected Behavior

A failed write leaves the previous `~/.openwiki/.env` intact — the user keeps all previously saved
tokens and keys.

### Actual Behavior

`writeFile` truncates the file to `0` and then fails (e.g. ENOSPC), so `~/.openwiki/.env` is left
empty. The user must re-run OAuth for every provider and re-enter every API key.

### Environment

- OS: macOS / Linux
- Node.js version: 22+
- OpenWiki version: `main` @ e2a7ca5 (0.2.0)

### Additional Context

Fix: write to a temp file in the same directory (`<path>.<pid>.tmp`), `chmod 0600`, then
`rename(...)` it over the target — an atomic swap that keeps the original file until the new contents
are fully on disk. I have a fix plus a regression test (the test fails on the current code and passes
with the atomic write) ready and will open a PR referencing this issue.

thejesh23:
## What & why

`saveOpenWikiEnv` (`src/env.ts`) persisted OAuth tokens and API keys to `~/.openwiki/.env` with a
plain `writeFile(openWikiEnvPath, …)`. `writeFile` opens the existing file with `O_TRUNC`, so it
truncates the file to zero **before** writing the new contents. Any failure mid-write (ENOSPC, crash,
power loss) therefore left `~/.openwiki/.env` empty and wiped **every** saved token and key —
forcing the user to re-run OAuth for every provider and re-enter every API key.

The fix stages the new contents in a same-directory temp file (`<path>.<pid>.tmp`), `chmod 0600`s it,
then `rename`s it over the target. `rename` is atomic and never truncates the original, so a failed
write leaves the existing credentials fully intact.

## How it was tested

- Added a regression test to `test/env-behavior.test.ts` that seeds `~/.openwiki/.env` with
  credentials, then forces the content write to fail the way a full disk would (truncate the target,
  then throw `ENOSPC`) and asserts the original file is unchanged. The test **fails on the previous
  code** (the real file is truncated) and **passes with the atomic write**.
- `pnpm run format:check`, `pnpm run lint:check`, `pnpm run typecheck`, `pnpm test` all pass
  (415 tests).

Fixes #407


colifran (approved):
Nice. Love this. Thank you!

PR #415, opened 2026-07-20
fix: reject cron day+weekday combinations launchd cannot represent

### LINKED ISSUE #410: launchd schedule ANDs day-of-month and weekday, contradicting cron OR semantics
### Summary

A cron expression that pins **both** a specific day-of-month and a specific weekday is translated into a launchd `StartCalendarInterval` that sets `Day`, `Month`, and `Weekday` simultaneously. launchd **ANDs** those constraints, while cron **ORs** day-of-month and day-of-week — so the installed schedule fires far less often than intended (often effectively never).

### Details

```ts
// src/schedules.ts:597-617
const parsedDay = getSingleCronNumber(day, { max: 31, min: 1 });
if (parsedDay !== null) {
  interval.Day = parsedDay;
} else if (day !== "*") {
  return null;
}

const parsedMonth = getSingleCronNumber(month, { max: 12, min: 1 });
if (parsedMonth !== null) {
  interval.Month = parsedMonth;
} else if (month !== "*") {
  return null;
}

const parsedWeekday = getSingleCronNumber(weekday, { max: 7, min: 0 });
if (parsedWeekday !== null) {
  interval.Weekday = parsedWeekday === 7 ? 0 : parsedWeekday;
} else if (weekday !== "*") {
  return null;
}
```

For a cron like `0 2 3 6 1` (day-of-month 3, June, Monday), cron fires on **June 3rd OR any Monday in June**. The emitted plist requires **June 3rd AND Monday** — i.e. only when June 3rd happens to be a Monday, which is almost never. The result is a near-dead timer with no error surfaced.

### Impact

Latent rather than a live regression: the setup wizard's default schedules use `*` for day/weekday, so most users won't hit it. But any user who hand-enters a cron combining a specific DoM and DoW gets a silently broken schedule.

### Suggested fix

When both a specific day-of-month and a specific weekday are present, treat the expression as too complex for launchd's AND semantics and return `null` so it flows through the existing "unsupported schedule" warning path (`src/schedules.ts:152-159`) rather than emitting a contradictory plist.

### Location

`src/schedules.ts:597-617`

bikeusaland:
## Summary

Fixes a scheduling bug where a cron expression pinning both a day-of-month and a weekday installs a launchd timer that almost never fires (#410).

launchd's `StartCalendarInterval` **ANDs** every field it is given, whereas cron **ORs** day-of-month and day-of-week when both are restricted. `parseLaunchdCalendarInterval` set `Day`, `Month`, and `Weekday` together, so a cron like `0 2 3 6 1` — which cron reads as "June 3rd **OR** any Monday in June" — became a plist meaning "June 3rd **AND** a Monday", i.e. only when June 3rd happens to be a Monday. The result was a near-dead timer with no error surfaced to the user.

## Fix

When both day-of-month and weekday are restricted, `parseLaunchdCalendarInterval` now returns `null`, so the caller falls back to the existing "Schedule saved, but this cron expression is too complex for direct launchd installation." warning path instead of writing a contradictory plist.

Single-field schedules are unaffected: day-only (`0 2 3 * *`), weekday-only (`0 2 * * 1`, including the Sunday `7 -> 0` normalization), and monthly (`0 2 3 6 *`) all still translate as before.

## Tests

New `test/launchd-calendar-interval.test.ts` (6 tests) covers the daily / day-only / weekday-only / monthly mappings, the Sunday normalization, the rejected day+weekday combination, and unrepresentable ranges/steps. The translator is exported for this coverage.

Full suite green: typecheck, `eslint .`, `prettier --check`, and 420 tests pass.

Closes #410


colifran (approved):
Good catch. Thanks @bikeusaland!

PR #416, opened 2026-07-20
fix: make OpenRouter debug-fetch patch concurrency-safe

### LINKED ISSUE #411: Global fetch monkey-patch for OpenRouter diagnostics is not concurrency-safe
### Summary

`installOpenRouterDebugFetch` reassigns `globalThis.fetch` for the duration of a run and restores it blindly in a `finally`. If two `runOpenWikiAgent` calls overlap in the same process, their install/restore steps corrupt each other — the wrapper can leak permanently or be lost. It also only captures error bodies for OpenRouter, leaving Anthropic/OpenAI/Bedrock 4xx/5xx bodies invisible to `classifyError`.

### Details

```ts
// src/agent/index.ts:1399-1452
function installOpenRouterDebugFetch(
  options: OpenWikiRunOptions,
): OpenRouterFetchCapture {
  const originalFetch = globalThis.fetch;
  // ...
  globalThis.fetch = (async (input, init) => { /* ... */ }) satisfies typeof fetch;

  return {
    // ...
    restore: () => {
      globalThis.fetch = originalFetch;
    },
  };
}
```

Failure mode with overlapping runs (subagents, tests running in parallel, a long-lived host process):

1. Run A installs, capturing the real `fetch` as `originalFetch`.
2. Run B installs, capturing **A's patched fetch** as its "original."
3. Run A finishes and restores the real `fetch`.
4. Run B finishes and restores **A's patched fetch** — now leaked into the global forever.

Depending on interleaving, the patch can also be lost mid-run.

### Suggested fix

Attach diagnostics per-model via the provider client's `configuration.fetch` hook — the same mechanism already used for Codex and Vertex (`src/agent/index.ts:710,853`) — instead of a process-global swap. This removes the concurrency hazard and generalizes error-body capture to every provider, not just OpenRouter.

If a global swap must be retained short-term, guard it with a reentrancy counter and a stack-based restore so nested installs compose correctly.

### Location

`src/agent/index.ts:1399-1452`

bikeusaland:
## Summary

Fixes a concurrency hazard where the OpenRouter debug-fetch patch could leak permanently or be lost when two OpenWiki runs overlap in the same process (#411).

`installOpenRouterDebugFetch` swapped `globalThis.fetch` for the duration of a run and restored it blindly in a `finally`. With two overlapping `runOpenWikiAgent` calls — subagents, parallel tests, or a long-lived host process — the install/restore steps corrupted each other:

1. Run A installs, capturing the real `fetch` as its original.
2. Run B installs, capturing **A's patched fetch** as its "original".
3. Run A restores the real `fetch`.
4. Run B restores **A's patched fetch** — now leaked into the global forever.

Both runs also shared a single failure slot, so one run's captured error could overwrite another's.

## Why not a per-model `configuration.fetch` hook

The issue suggested attaching diagnostics per-model like the Codex/Vertex clients do. That isn't available here: `ChatOpenRouter` (from `@langchain/openrouter`) extends `BaseChatModel` and calls `globalThis.fetch` directly — its constructor exposes no `fetch`/`configuration` field. A global wrapper is genuinely the only interception point, which is why the original code patched the global.

## Fix

Make the global wrapper **reference-counted**:

- Install the wrapper once, capturing the genuine original `fetch`.
- Register a **per-run failure sink** so concurrent runs never share or clobber each other's captured failure. `getLastFailure`/`clearLastFailure` operate on the calling run's own sink, and debug events route to each run's own `onEvent`.
- Restore the global only when the **last** run detaches, reverting to the true original — never to a wrapper installed by a concurrent run.
- Redundant `restore()` calls on the same handle are idempotent (can't prematurely restore while another run is active).

The `OpenRouterFetchCapture` return contract (`getLastFailure`, `clearLastFailure`, `restore`) is unchanged, so callers are untouched.

## Tests

New `test/openrouter-debug-fetch.test.ts` (4 tests): single-run restore returns the exact original; overlapping runs each capture their own 429 while restore stays reference-counted; non-OpenRouter requests pass through untouched; redundant restore is a no-op that doesn't disturb an active run.

Full suite green: typecheck, `eslint .`, `prettier --check`, and all tests pass.

Closes #411


bikeusaland:
@colifran This PR is also ready for review when you have a chance — it fixes #411, which is still open. The concurrency hazard in `installOpenRouterDebugFetch` is still present on main, the PR has no conflicts, and CI is green. Thanks!

colifran:
@bikeusaland just letting you know i see both of your PRs and i'll work on reviewing and hopefully merging by the end of the week!

bikeusaland:
**CI failure root cause:** the new `test/openrouter-debug-fetch.test.ts` imports from a module that doesn't exist:

```ts
import { OPENROUTER_BASE_URL } from "../src/constants.ts"; // line 3
```

There is no `src/constants.ts` in the repo — the constant is exported from `src/config/constants.ts`. The whole suite fails at import time on **Test (24)**:

```
Error: Cannot find module '../src/constants.ts' imported from test/openrouter-debug-fetch.test.ts
{ code: 'ERR_MODULE_NOT_FOUND' }
```

so none of the 4 new tests ran (all other 221 test files / 3020 tests passed). **Test (22)** is only CANCELLED via matrix fail-fast — nothing separately wrong on Node 22. Typecheck/Build passed because they only cover `src/`, not the test files.

Fix is a one-liner:

```ts
import { OPENROUTER_BASE_URL } from "../src/config/constants.ts";
```

colifran (approved):
Thanks, @bikeusaland! Looks good

PR #417, opened 2026-07-20
feat: resilient HTTP for connectors (timeout + retry/backoff on 429/5xx)

bikeusaland:
## Summary

Adds shared resilient HTTP to connector ingestion, addressing the resilience findings from the code review (connector rate-limit / retry gap #5 and the HTTP MCP no-timeout gap #6).

Every direct-API connector (Gmail, Slack, X, Hacker News) and the HTTP MCP client called `fetch` directly with **no timeout and no retry**. Consequences:

- A single **429** or a transient **5xx** threw straight out and aborted the whole ingestion run.
- A **non-responsive server hung ingestion forever** — the stdio MCP client had a 60s per-request timeout, but the HTTP one had none.

## What changed

New `src/connectors/http.ts` exports `fetchWithResilience(input, init, options)`:

- **Per-request timeout** via `AbortSignal.timeout` (default 30s; the HTTP MCP client passes 60s for parity with the stdio client).
- **Bounded exponential backoff with jitter** on 429 and 5xx, honoring a numeric or HTTP-date **`Retry-After`** header when present.
- **Retry on network errors** (connection reset, DNS failure, timeout), rethrowing the last error once attempts are exhausted.

Deliberately **not** retried: 401/403 and other 4xx are returned unchanged. That keeps Gmail's existing `401 → refresh token → retry` path working (it must see the 401), and avoids retry storms that could lock accounts.

A caller-supplied `AbortSignal` is combined with the timeout via `AbortSignal.any`. Sleep and the jitter RNG are injectable, so tests run instantly and deterministically without a `Math.random` dependency.

Wired through Gmail, Slack, X, Hacker News, and the HTTP MCP client (`HttpJsonRpcClient.post`). The Tavily-backed web-search connector uses its own SDK and is unaffected.

## Tests

New `test/fetch-with-resilience.test.ts` (11 tests): retry-then-succeed on 429, `Retry-After` honored (numeric and HTTP-date), 5xx retried to the limit, 401 and 4xx **not** retried, network errors retried then rethrown, and an `AbortSignal` passed to `fetch`. Existing connector tests still pass unchanged.

Full suite green: typecheck, `eslint .`, `prettier --check`, and 425 tests pass.


src/connectors/http.ts
```diff
@@ -0,0 +1,142 @@
...
+ * transient, and callers such as Gmail need to see a 401 to trigger a token
+ * refresh. Retrying them would waste attempts and could lock accounts.
+ */
+
+const DEFAULT_TIMEOUT_MS = 30_000;
+const DEFAULT_MAX_RETRIES = 3;
+const DEFAULT_BASE_DELAY_MS = 500;
+const MAX_BACKOFF_DELAY_MS = 20_000;
+
+export type FetchWithResilienceOptions = {
```
colifran: nit: can we make this an interface?

bikeusaland: Done — converted `FetchWithResilienceOptions` to an interface in cd46216.

bikeusaland:
Requested change is done — converted `FetchWithResilienceOptions` from a type alias to an interface in cd46216. eslint/prettier clean and all 11 resilience tests pass.

bikeusaland:
Merge conflicts resolved — merged latest `main` into the branch (commit 6baedc2). The 3 conflicts (gmail, hackernews, x) were all import-block collisions where `main` added `normalizeStringArray` alongside this branch's `fetchWithResilience`; kept both imports and confirmed each is used. Full suite green after `pnpm install`: typecheck, eslint, prettier, and 670 tests pass. PR is now mergeable.

src/connectors/http.ts
```diff
@@ -0,0 +1,142 @@
...
+    const timeoutSignal = AbortSignal.timeout(timeoutMs);
+    const signal = init.signal
+      ? AbortSignal.any([init.signal, timeoutSignal])
+      : timeoutSignal;
+
+    try {
+      const response = await fetch(input, { ...init, signal });
+
+      if (attempt < maxRetries && isRetryableStatus(response.status)) {
+        const retryAfterMs = parseRetryAfterMs(
```
colifran, line 120: shouldn't this be passing a`now` value?

src/connectors/http.ts
```diff
@@ -0,0 +1,142 @@
...
+export async function fetchWithResilience(
+  input: Parameters<typeof fetch>[0],
+  init: Parameters<typeof fetch>[1] = {},
+  options: FetchWithResilienceOptions = {},
+): Promise<Response> {
+  const timeoutMs = options.timeoutMs ?? DEFAULT_TIMEOUT_MS;
+  const maxRetries = options.maxRetries ?? DEFAULT_MAX_RETRIES;
+  const baseDelayMs = options.baseDelayMs ?? DEFAULT_BASE_DELAY_MS;
+  const sleep = options.sleep ?? realSleep;
+  const random = options.random ?? (() => 0.5);
```
colifran: We could have true jitter if random is `undefined`:

```suggestion
  const random = options.random ?? Math.random;
```

colifran (commented):
A few comments and then good to merge!

src/connectors/http.ts
bikeusaland: Good catch — this was a real bug. Without a `now`, `parseRetryAfterMs` hit the `now === undefined` guard and returned `null` for HTTP-date headers, so the advertised HTTP-date `Retry-After` support was effectively dead code (it silently fell back to plain backoff). Fixed in 02b6baf by passing `Date.now()`, and added a test that exercises an HTTP-date header end-to-end with the clock pinned via fake timers.

src/connectors/http.ts
bikeusaland: Agreed — the fixed 0.5 defeated the point of jitter (every client waits exactly half the ceiling, so they don't de-correlate). Applied your suggestion in 02b6baf: `random` now defaults to `Math.random`. The original 'unavailable in some sandboxes' note didn't apply here — `Math.random` is already used for jitter elsewhere in `src` (e.g. `langsmith/api.ts`). Added a test covering the default jitter path.

bikeusaland:
Addressed both review comments in 02b6baf:

- **HTTP-date `Retry-After`** — now passes `Date.now()` into `parseRetryAfterMs`. This was a genuine bug: without `now`, HTTP-date headers returned `null` and silently fell back to plain backoff, so the advertised HTTP-date support was dead code.
- **Jitter** — `random` now defaults to `Math.random` for real full jitter instead of a fixed 0.5.

Added 2 tests (HTTP-date honored end-to-end with a pinned clock; default `Math.random` jitter path). Full suite green: typecheck, eslint, prettier, and 672 tests pass. Both threads resolved.

colifran (approved):
Nice - great work!

PR #400, opened 2026-07-20
feat: support Bedrock default AWS credentials

pengliangtian06-commits:
## Summary
- let Bedrock use the AWS SDK default credential chain when static keys are absent
- preserve static access-key authentication when both key values are configured
- accept `AWS_REGION` and `AWS_DEFAULT_REGION` in addition to `BEDROCK_AWS_REGION`
- add a Simplified Chinese README and a language switcher in the main README
- document IAM role/profile setup and add regression coverage

## Validation
- `pnpm format:check`
- `pnpm lint:check`
- `pnpm test` (417 tests)
- `pnpm build`

Closes #397

Issue #397, opened 2026-07-19
[Feature]: Allow using current AWS IAM Role rather then BEDROCK_AWS_ACCESS_KEY_ID when using Bedrock provider


the-mentor:
### Summary

Allow using the current AWS IAM Role rather the using the BEDROCK_AWS_ACCESS_KEY_ID environment variable. 

### Motivation

This will simplify the usage of openwiki implementation when running in CI with already existing AWS Role.

### Proposed Solution

Implement support for AWS IAM Role for bedrock

colifran:
I believe this has been resolved via #418.

PR #384, opened 2026-07-17
feat: support default AWS credentials chain instead of static creds

jzhn:
### What
Make the `bedrock` provider authenticate via the **AWS SDK default credential provider chain** (shared profile, IRSA/web-identity, ECS/EC2 instance role, or ambient `AWS_*` env vars) instead of requiring a pasted static access key + secret. Static keys remain supported for backwards compatibility.

### Why
Previously Bedrock only worked with long-lived `BEDROCK_AWS_ACCESS_KEY_ID` / `BEDROCK_AWS_SECRET_ACCESS_KEY`, and onboarding blocked until they were entered. That's a non-starter for IRSA/ECS/instance-role deployments and anyone using a shared AWS profile.

### How
Bedrock is now a **keyless provider**, modeled on the existing `gemini-enterprise` (Google ADC) pattern:
- `PROVIDER_CONFIGS.bedrock` no longer declares `apiKeyEnvKey`/`secretKeyEnvKey`/`regionEnvKey`, so no provider-config machinery treats AWS creds as required.
- `createModel` passes `credentials` only when both `BEDROCK_AWS_ACCESS_KEY_ID` and `BEDROCK_AWS_SECRET_ACCESS_KEY` are set; otherwise it omits them and lets the SDK's default chain resolve. Region resolves from `BEDROCK_AWS_REGION` → `AWS_REGION` → `AWS_DEFAULT_REGION` (the SDK only checks `AWS_DEFAULT_REGION` on its own, so we bridge `AWS_REGION` which IRSA/ECS commonly inject).
- Onboarding no longer prompts for AWS credentials when Bedrock is selected.

### Usage

Default credential chain (no static keys needed):
```bash
OPENWIKI_PROVIDER=bedrock
OPENWIKI_MODEL_ID=anthropic.claude-sonnet-5
# region from BEDROCK_AWS_REGION / AWS_REGION / AWS_DEFAULT_REGION
```

Static-key override (unchanged behavior):
```bash
OPENWIKI_PROVIDER=bedrock
BEDROCK_AWS_ACCESS_KEY_ID=...
BEDROCK_AWS_SECRET_ACCESS_KEY=...
BEDROCK_AWS_REGION=us-east-1
OPENWIKI_MODEL_ID=anthropic.claude-sonnet-5
```

### Tests
- Updated unit test
- manual testing with `AWS_PROFILE` with credentials_process - credentials_process is invoked by openwiki to obtain Bedrock credentials

PR #418, opened 2026-07-21
feat: support the AWS SDK credential chain for Bedrock

geonwoo-jeong:
> [!NOTE]
> #384 and #400 pursue the same Bedrock credential-chain goal. This PR is an independent implementation based on current `main`, with the behavioral contract and regression coverage made explicit for comparison. I am happy to adapt or consolidate whichever implementation the maintainers prefer.

## Summary

- let the Bedrock provider use the [AWS SDK default credential provider chain](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html) when a legacy `BEDROCK_*` key pair is absent
- preserve complete legacy Bedrock key pairs, including `BEDROCK_AWS_SESSION_TOKEN`
- resolve the region from `BEDROCK_AWS_REGION`, `AWS_REGION`, then `AWS_DEFAULT_REGION`
- keep onboarding, startup validation, diagnostics, and redaction consistent with SDK-managed credentials

## Why

Addresses #397.

OpenWiki currently requires explicit Bedrock access keys during setup and injects them when constructing `ChatBedrockConverse`. That prevents deployments from using temporary credentials already supplied through GitHub Actions OIDC/web identity, IAM roles, AWS profiles/SSO, ECS, or EC2.

This change delegates credential discovery and refresh to `@langchain/aws` and the AWS SDK instead of implementing a separate credential resolver.

## Credential behavior

| Configuration | Behavior |
| --- | --- |
| `AWS_BEARER_TOKEN_BEDROCK` | Preserve the existing LangChain precedence |
| Complete legacy Bedrock key pair | Continue using it, including an optional session token |
| No legacy Bedrock key pair | Use the AWS SDK default chain, including standard `AWS_*` credentials |
| Partial legacy or standard AWS access/secret pair | Fail early with the missing environment variable instead of silently falling back |
| Region configuration | Prefer `BEDROCK_AWS_REGION`, then `AWS_REGION`, then `AWS_DEFAULT_REGION` |

The model-selection and cross-region inference behavior introduced in #327 is unchanged.

## Scope and review guide

This is one Bedrock-authentication change. It does not modify dependencies, generated workflow templates, model selection, or generated `openwiki/` documentation. The README change only notes that the credential chain is supported. Of the 953 added lines, 551 are regression tests.

- `src/constants.ts`: represent SDK-managed authentication explicitly and centralize credential-pair and region resolution
- `src/agent/index.ts`: omit the explicit `credentials` option so the SDK owns resolution and refresh, while retaining the existing model, region, and retry configuration
- onboarding, CLI, and diagnostics: skip static-key collection for SDK credentials, reject incomplete pairs, provide AWS-specific remediation, and redact AWS credential/token values
- tests: cover SDK delegation, temporary environment credentials, legacy compatibility, startup/onboarding behavior, region precedence, auth-error classification, and redaction

A generated GitHub Actions workflow is intentionally out of scope: this PR makes OpenWiki consume credentials already supplied to the process, such as those exported by `aws-actions/configure-aws-credentials`.

## Validation

- [x] `pnpm run format:check`
- [x] `pnpm run lint:check`
- [x] `pnpm run typecheck`
- [x] `pnpm run build`
- [x] `pnpm test` — 40 files, 478 tests
- [x] `pnpm run coverage`
- [x] `pnpm audit --prod`
- [x] `git diff --check`

Manual smoke test: ran OpenWiki from this branch with Bedrock and confirmed a repository run completes successfully through the AWS SDK credential chain. The automated tests additionally exercise model construction and temporary environment credentials without network access; credential exchange and refresh remain AWS SDK responsibilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,8 +302,8 @@ Some local servers ignore the API key value, but OpenWiki still requires
 ### AWS Bedrock
 
 The `bedrock` provider calls foundation models hosted on AWS Bedrock using IAM
-credentials rather than a single vendor API key. It authenticates with an AWS
-access key ID, a secret access key, and a region:
+credentials rather than a single vendor API key. Existing installations can
+continue to provide an AWS access key ID, secret access key, and region:
 
 ```bash
 OPENWIKI_PROVIDER=bedrock
@@ -312,6 +312,11 @@ BEDROCK_AWS_SECRET_ACCESS_KEY=your-secret-access-key
 BEDROCK_AWS_REGION=us-east-1
 OPENWIKI_MODEL_ID=anthropic.claude-sonnet-5
 ```
+
+When the explicit Bedrock credentials are not set, OpenWiki uses the AWS SDK
+default credential provider chain, including OIDC/web identity, IAM roles,
+AWS profiles, and ECS/EC2 credentials. The region is resolved from
+`BEDROCK_AWS_REGION`, `AWS_REGION`, or `AWS_DEFAULT_REGION`.
 
 Which model IDs are available depends on your AWS account and region (which
 foundation models you've enabled in the Bedrock console), so there is no
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -55,7 +55,10 @@ import type {
 import {
   ANTHROPIC_BASE_URL_ENV_KEY,
   BASETEN_BASE_URL_ENV_KEY,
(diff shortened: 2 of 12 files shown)


colifran (approved):
Awesome work! Thanks @geonwoo-jeong!

PR #424, opened 2026-07-21
fix: restore scheduled OpenWiki Update fork guard and protect it from regeneration

### LINKED ISSUE #423: [Bug]: openwiki code --update regenerates .github/workflows/openwiki-update.yml unconditionally, silently reverting the fork-scheduling guard
## Description

The fork-scheduling guard on `.github/workflows/openwiki-update.yml` (originally added in #303, restored in #370 after disappearing once already in #179) has been silently removed a second time — this time by #377, an automated `docs: update OpenWiki` run from the scheduled bot itself.

`openwiki code --update` appears to regenerate this workflow file from its internal template on every run rather than preserving the existing, repository-specific content. #285 already tracks this as the root cause and looks like the right fix — this report is mainly to flag that, until it merges, the guard keeps getting reverted on roughly a weekly cadence as the bot cycles, so the underlying `#302` failure mode is live again on `main` right now.

## Steps to Reproduce

1. `.github/workflows/openwiki-update.yml` contains the fork guard from #370 (`if: github.event_name == 'workflow_dispatch' || github.repository == 'langchain-ai/openwiki' || vars.OPENWIKI_ENABLE_SCHEDULED_UPDATE == 'true'`).
2. The scheduled `OpenWiki Update` workflow runs `openwiki code --update --print` (as it does daily).
3. The regenerated `.github/workflows/openwiki-update.yml` from that run's PR (#377) no longer contains the guard, the explanatory comment, or the pinned `actions/checkout@v7`/`actions/setup-node@v7` versions — they were reverted to the generic template's defaults.

## Expected Behavior

Repository-specific customizations to `.github/workflows/openwiki-update.yml` — the fork guard, its comment, and pinned action versions — should survive a scheduled `code --update` run rather than being overwritten by the generic template.

## Actual Behavior

As of this report, `main`'s installed workflow has no fork guard again. Any contributor who forks the repo and enables Actions (which most do, just to see their own CI) will hit the same daily failure #302 originally described.

I reproduced the current failure mode on my own fork via manual dispatch: https://github.com/jyje/openwiki/actions/runs/29509927346 — checkout/setup/install succeed, then `Run OpenWiki` fails for lack of `OPENROUTER_API_KEY`, which is exactly what the guard exists to prevent by default.

## Environment

Not applicable to a specific machine — this reproduces from the scheduled/generated CI workflow itself on any fork or on `main`.

## Additional Context

Full chain for reference: #302 (original report) → #303 (fix) → #179 (unrelated PR, first accidental regression) → #369 (regression report) → #370 (second fix, also added a `workflow_dispatch` carve-out) → #377 (scheduled bot run, second regression) → #285 (tracked root-cause fix, still open as of this report).

Not trying to pile on — #285 looks like exactly the right fix, this is just to make the current live impact and the repeat pattern visible with fresh evidence while it's in review. Happy to help verify once it lands! 🙂

jyje:
Closes #423

## Summary

The scheduled `OpenWiki Update` workflow's fork guard (#303 → #370) was silently dropped a second time by #377 — an automated `docs: update OpenWiki` run from the bot itself. `openwiki code --update` regenerates `.github/workflows/openwiki-update.yml` from its internal template on every run, which drops any repository-specific customization. #285 tracks the proper CLI-level fix but is still open.

This restores the #370 guard (including its `workflow_dispatch` carve-out) and adds a step that discards whatever the agent just wrote to this file before the PR-creation step ever sees it — an interim, workflow-level mitigation until #285 lands.

## Testing

**Local reproduction against the published CLI**
In a scratch repo seeded with the guarded workflow file, `npx openwiki@0.2.1 code --update --print` reliably stripped the guard (16 lines removed) even though the run failed immediately on model validation — confirming the regeneration happens unconditionally, before any credential check. `git checkout -- .github/workflows/openwiki-update.yml` restored it losslessly.

**Live run on my fork**
https://github.com/jyje/openwiki/actions/runs/29840884136 — `Run OpenWiki` fails as expected (no provider secret configured on my fork), and `Restore protected workflow file` still runs (via `if: always()`) and succeeds, confirming the guard survives even on the failure path that #377 hit.


jyje:
Follow-up: also verified the success path directly (the exact path #377 hit), since the earlier test only covered the failure path.

My fork's `OPENROUTER_API_KEY` didn't have enough credit for `z-ai/glm-5.2`'s token budget, so I temporarily pointed `OPENWIKI_MODEL_ID` at a free OpenRouter model with tool-calling support (`nvidia/nemotron-3-super-120b-a12b:free`, 1M context) just for this test, then reverted it right after.

Result: https://github.com/jyje/openwiki/actions/runs/29842223327 — `Run OpenWiki` completes successfully this time, and `Restore protected workflow file` still runs and succeeds *after a successful run*, not just after a failed one. Since `git checkout -- <path>` deterministically makes the working-tree file identical to the committed version, a successful `Restore` step here guarantees the file is unchanged by the time `Create OpenWiki update pull request` would evaluate its diff.

(That last step itself failed independently, on "GitHub Actions is not permitted to create or approve pull requests" — a repo Actions-permission setting on my fork, unrelated to this fix.)

colifran (approved):
Thanks!

PR #435, opened 2026-07-22
fix: keep personal mode from following repository agent files

BenDManning:
## What and why

Personal/local-wiki mode can inspect configured repositories as evidence. Those repositories may contain `AGENTS.md` and `CLAUDE.md` files intended for repository code agents. The existing prompt said personal mode should not manage those files, but did not explicitly say they are not personal-brain instructions.

This change tells local-wiki mode not to read or follow repository agent files unless the user explicitly asks about their contents. Repository/code mode and its agent-file generation are unchanged.

## Testing

- `pnpm test` — 444 tests passed
- `pnpm typecheck`
- `pnpm build`
- `pnpm lint:check`
- `pnpm format:check`
- Exact staged-candidate test and build verification
- Independent exact-snapshot review: clean


colifran (approved):
Thanks @BenDManning!

PR #441, opened 2026-07-23
fix: allow vertex claude client under the mermaid jsdom shim

### LINKED ISSUE #440: Mermaid support (jsdom globals) trips the Anthropic SDK browser guard on the Vertex Claude surface
## Summary

Enabling the optional Mermaid validation path (installing `mermaid` + `jsdom`) breaks documentation runs on the Gemini Enterprise **Vertex "anthropic" surface** (Claude on Vertex). The Anthropic SDK refuses to construct its client with a "browser-like environment" error and the whole run aborts with exit code 1 before any docs are generated.

As currently wired, Mermaid support and Claude-on-Vertex are mutually exclusive.

## Environment

- Provider: `OPENWIKI_PROVIDER=gemini-enterprise` (Claude on Vertex via the Anthropic Vertex SDK, ADC auth)
- Model: a `claude-*` model ID (resolves to the Vertex `anthropic` surface)
- `mermaid` + `jsdom` resolvable at runtime (the workflow scaffolded by `openwiki code` installs them for high-fidelity diagram validation)
- Command: `openwiki code --update`

## Error

```
It looks like you're running in a browser-like environment.

This is disabled by default, as it risks exposing your secret API credentials to attackers.
If you understand the risks and have appropriate mitigations in place,
you can set the `dangerouslyAllowBrowser` option to `true`, e.g.,

new Anthropic({ apiKey, dangerouslyAllowBrowser: true });
```

## Root cause

1. `src/mermaid/dom-shim.ts` (`ensureDomGlobals`) installs jsdom's `window` and `document` as process-wide globals so Mermaid/DOMPurify can parse headless. This is by design and must happen before `mermaid` is first imported.
2. The Anthropic SDK's browser guard, `isRunningInBrowser()`, returns true when `window`, `window.document`, and `navigator` are all defined. On Node ≥ 21 `navigator` is always a global, and the shim supplies the other two — so the guard genuinely trips in the CLI/CI process.
3. The SDK constructor throws unless `dangerouslyAllowBrowser: true` is passed.
4. `ChatAnthropic` *does* pass `dangerouslyAllowBrowser: true` into its `createClient` hook for exactly this reason — but OpenWiki's Vertex `createClient` override in `createGeminiEnterpriseModel` (`src/agent/index.ts`) is a zero-arg `() =>` that ignores the argument, so the flag never reaches `AnthropicVertex`.

The non-Vertex `anthropic` provider (which uses LangChain's default `createClient`, `(options) => new Anthropic(options)`) and all `ChatOpenAI`-family providers (which set `dangerouslyAllowBrowser: true` by default) are unaffected.

## Fix

Set `dangerouslyAllowBrowser: true` on the `AnthropicVertex` construction inside the Vertex `createClient` hook. OpenWiki always runs as a Node CLI/CI process, not a real browser, so the guard is a false positive here; the flag's only runtime effect is an `anthropic-dangerous-direct-browser-access` header and it exposes no credential.

Care must be taken to forward *only* that flag and not the `ANTHROPIC_*` auth options LangChain also hands the hook — those are deliberately neutralized on this surface so they can't clobber the Google OAuth token.

## Workaround

Until fixed: do **not** install `mermaid`/`jsdom` alongside the CLI when using the Gemini Enterprise Claude/Vertex surface. This disables diagram validation but keeps `code --update` working.

## Repro

1. Configure `OPENWIKI_PROVIDER=gemini-enterprise` with a `claude-*` model and Vertex ADC.
2. Ensure `mermaid` + `jsdom` are resolvable at runtime.
3. Run `openwiki code --update` — it aborts with the "browser-like environment" error before generating docs.

ppsplus-bradh:
## Summary

Enabling the optional Mermaid validation path (`mermaid` + `jsdom`) breaks documentation runs on the Gemini Enterprise **Vertex "anthropic" surface** (Claude on Vertex): the Anthropic SDK refuses to construct with a "browser-like environment" error and the run aborts before any docs are generated. This makes Mermaid support and Claude-on-Vertex mutually exclusive.

Fixes #440.

## Root cause

- `src/mermaid/dom-shim.ts` installs jsdom's `window`/`document` as process-wide globals so DOMPurify can parse headless (by design, and before `mermaid` is first imported).
- The Anthropic SDK's `isRunningInBrowser()` returns true when `window`, `window.document`, and `navigator` are all defined. On Node ≥ 21 `navigator` is always present, so the shim trips the guard in the CLI/CI process, and the SDK constructor throws unless `dangerouslyAllowBrowser: true`.
- `ChatAnthropic` already passes `dangerouslyAllowBrowser: true` into its `createClient` hook for exactly this reason, but the Vertex `createClient` override in `createGeminiEnterpriseModel` is a zero-arg `() =>` that ignores the argument, dropping the flag before it reaches `AnthropicVertex`.

The non-Vertex `anthropic` provider (LangChain's default `createClient` forwards the flag) and all `ChatOpenAI`-family providers (flag set by default) were already immune.

## Fix

Set `dangerouslyAllowBrowser: true` on the `AnthropicVertex` construction inside the Vertex `createClient` hook. OpenWiki always runs as a Node CLI/CI process, so the guard is a false positive here; the flag's only runtime effect is an `anthropic-dangerous-direct-browser-access` header and it exposes no credential.

Only that flag is forwarded — deliberately **not** the `ANTHROPIC_*` auth options LangChain also hands the hook, which are neutralized on this surface (`withAnthropicAuthEnvNeutralized`) so they can't clobber the Google OAuth token.

## Testing

- New end-to-end regression test (`test/gemini-enterprise-claude.e2e.test.ts`) constructs the client with the **real** Anthropic Vertex SDK and the **real** jsdom shim active; it throws the "browser-like environment" error if the flag is removed, and passes with it.
- Existing unit test extended to assert the flag is forwarded and that the `ANTHROPIC_*` auth options are not.
- Full suite green (474 tests), typecheck + lint + build clean.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -827,6 +827,17 @@ function createGeminiEnterpriseModel(
       // ANTHROPIC_API_KEY requirement. The env is neutralized around the
       // constructor so a stray ANTHROPIC_API_KEY/ANTHROPIC_AUTH_TOKEN cannot
       // clobber the Google OAuth token (see withAnthropicAuthEnvNeutralized).
+      //
+      // dangerouslyAllowBrowser: the Anthropic SDK (base of AnthropicVertex)
+      // refuses to construct when it detects `window`/`document`/`navigator` —
+      // its browser-credential-exposure guard. OpenWiki is always a Node CLI/CI
+      // process, but the optional Mermaid validation path installs jsdom DOM
+      // globals process-wide (see src/mermaid/dom-shim.ts), which trips that
+      // guard and aborts the whole run before any docs are generated. ChatAnthropic
+      // passes `dangerouslyAllowBrowser: true` into `createClient`, but this hook
+      // ignores its argument, so the flag is set explicitly here. It is forwarded
+      // to AnthropicVertex only — never the ANTHROPIC_* auth options LangChain
+      // also passes, which would defeat withAnthropicAuthEnvNeutralized.
       return new ChatAnthropic(stripPublisherPath(modelId), {
         createClient: () =>
           withAnthropicAuthEnvNeutralized(
@@ -834,6 +845,7 @@ function createGeminiEnterpriseModel(
               new AnthropicVertex({
                 projectId,
                 region: location,
+                dangerouslyAllowBrowser: true,
               }),
           ),
         ...retryOptions,
--- a/test/gemini-enterprise-claude.e2e.test.ts
+++ b/test/gemini-enterprise-claude.e2e.test.ts
@@ -1,25 +1,9 @@
-import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, test } from "vitest";
 import { ensureDomGlobals } from "../src/mermaid/dom-shim.ts";
 import { createModel } from "../src/agent/index.ts";
-
-// AnthropicVertex's constructor eagerly kicks off ADC resolution
-// (`new GoogleAuth().getClient()`), whose promise rejects with "Could not load
-// the default credentials" in an unauthenticated CI environment. That rejection
-// lands *after* this test resolves, so Vitest reports it as an unhandled
-// rejection and fails the run even though the test itself passes. Stub
-// google-auth-library so the eager auth lookup resolves to a dummy client. This
-// stubs *only* the credential lookup; the real AnthropicVertex is still
-// constructed, and its browser-environment guard still runs in the base-class
-// `super()` call — before any auth — so the fix remains exercised end-to-end.
-vi.mock("google-auth-library", () => ({
-  GoogleAuth: class {
-    getClient() {
-      return Promise.resolve({
-        getRequestHeaders: () => Promise.resolve({}),
-      });
-    }
-  },
-}));
 
 // End-to-end regression for issue #3, using the REAL Anthropic Vertex SDK (no
 // mock) and the REAL Mermaid DOM shim. The optional Mermaid validation path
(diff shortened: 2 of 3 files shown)


colifran:
Good catch

ppsplus-bradh:
Pushed a fix for the red `Test (24)` / `Test (22)` jobs, and merged in the latest `main`.

**What was failing:** the new e2e regression test actually *passed* (`✓ ... 465ms`), but it left behind an **unhandled promise rejection** that Vitest counts as a run failure:

```
Error: Could not load the default credentials.
This error originated in "test/gemini-enterprise-claude.e2e.test.ts"
 ELIFECYCLE  Test failed. See above for more details.
```

**Root cause:** `AnthropicVertex`'s constructor eagerly starts ADC resolution (`new GoogleAuth().getClient()`). In unauthenticated CI that promise rejects, and it settles *after* the test has already resolved — so nothing is there to catch it and Vitest fails the whole run. (`Test (22)` only shows `CANCELLED` because fail-fast killed the matrix sibling.)

**Fix:** stub `google-auth-library` in that test so the eager credential lookup resolves to a dummy client. Only the credential lookup is mocked — the **real** `AnthropicVertex` is still constructed, and its browser-environment guard still runs in the base-class `super()` call *before* any auth. Verified the guard is intact: removing `dangerouslyAllowBrowser` still makes this test fail with the real `It looks like you're running in a browser-like environment` error.

Locally after merging `main`: full suite green (507 tests, no unhandled rejection), typecheck + lint + format clean.

colifran (dismissed):
Good catch! Thanks @ppsplus-bradh!

colifran (approved):

ppsplus-bradh:
@colifran - Thanks for covering me there.  Was in the process of fixing my "fix".

colifran:
@ppsplus-bradh no worries! we are planning to release 0.2.3 EOD or early tomorrow so was anxious to get this one in. Thank you for your help!

PR #446, opened 2026-07-23
fix: stop reopening setup after every run when the LangSmith key was skipped

SadriG91:
## Summary

The first-run setup prompt for the LangSmith API key says *"Optional: … Press Enter with an empty value to skip"*, but skipping is never remembered: after **every** completed run, the setup wizard reopens at the LangSmith step.

### Root cause

Skipping the prompt persists `LANGSMITH_API_KEY=""` to `~/.openwiki/.env` (and to `process.env`). However, both setup gates treat any falsy value as "unanswered":

```ts
// needsCredentialSetup / needsCredentialPrompt
... || !process.env.LANGSMITH_API_KEY;
```

An empty string is falsy, so `needsCredentialSetup()` keeps returning `true`, and `shouldRunInteractiveCredentialSetup` reopens the wizard every time the run state returns to idle. The only ways to escape the loop are pasting a real LangSmith key or manually writing a placeholder into `~/.openwiki/.env` — both defeating "optional".

### Fix

Treat a **defined** key (including the empty value the skip path persists) as answered; only a fully unanswered (`undefined`) key triggers the LangSmith step:

```ts
... || process.env.LANGSMITH_API_KEY === undefined;
```

Applied to both occurrences (`needsCredentialSetup` and the wizard's `needsCredentialPrompt`).

### Reproduce (before this fix)

1. Fresh config (no `LANGSMITH_API_KEY` in `~/.openwiki/.env`), any provider configured.
2. Run `openwiki`, complete first-run setup, press Enter to skip the LangSmith step.
3. Ask any question; when the run completes, the setup wizard reopens at the LangSmith step. Repeats after every run.

## Testing

- `pnpm run build`, `pnpm exec vitest run` (505/505), `pnpm run lint:check`, `pnpm run format:check` all pass.
- Manually verified with a `~/.openwiki/.env` containing `LANGSMITH_API_KEY=""` (the state the skip path produces): the wizard no longer reopens after each run, and a config without the key at all still walks through the LangSmith step once.

### DIFF
--- a/src/credentials.tsx
+++ b/src/credentials.tsx
@@ -392,10 +392,7 @@ export function needsCredentialSetup(
     needsRegionStep(provider) ||
     (modelIdOverride === null &&
       process.env[OPENWIKI_MODEL_ID_ENV_KEY] === undefined) ||
-    // The LangSmith key is optional: skipping the prompt persists an empty
-    // value, so only a fully unanswered (undefined) key should reopen setup.
-    // Treating "" as unanswered made the wizard nag after every run.
-    process.env.LANGSMITH_API_KEY === undefined;
+    needsLangSmithStep();
 
   if (needsCredentials) {
     return true;
@@ -570,6 +567,23 @@ function needsRegionStep(provider: OpenWikiProvider): boolean {
   }
 
   return !isRegionConfigured(provider);
+}
+
+/**
+ * Whether the optional LangSmith tracing step still needs to be shown.
+ *
+ * The step is optional, so "answered" must include skipping it. Skipping does
+ * not persist `LANGSMITH_API_KEY` — `saveOpenWikiEnv` strips empty values, so
+ * the key is simply absent afterwards. What the step always records instead is
+ * `LANGCHAIN_TRACING_V2` (`"false"` on skip, `"true"` when a key is entered),
+ * which survives because it is non-empty. So the step is unanswered only when
+ * neither a key is present (e.g. from a shell export) nor a tracing decision
+ * has been recorded.
+ */
+export function needsLangSmithStep(
+  env: NodeJS.ProcessEnv = process.env,
+): boolean {
+  return !env.LANGSMITH_API_KEY && env.LANGCHAIN_TRACING_V2 === undefined;
 }
 
 function isRegionConfigured(provider: OpenWikiProvider): boolean {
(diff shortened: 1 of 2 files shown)


SadriG91:
@bracesproul or @colifran can you take a look it's a small change

colifran:
Thanks @SadriG91! Taking a look - this will be helpful. Thanks for the PR!

colifran:
I think the verification is hitting a state the skip path doesn't actually produce. When you skip, `saveOpenWikiEnv` strips empty values before writing, so the file ends up with `LANGSMITH_API_KEY` absent rather than ="". I was able to reproduce it even with this PR. After a real skip my .env had no key at all and `=== undefined` is still true in that state so setup keeps reopening.

The value that reliably survives is `LANGCHAIN_TRACING_V2` ("false" on skip, "true" if a key's entered), so gating on that instead does the trick I think. Could you take a look?

SadriG91:
Great catch, thank you — you're right, and I've pushed a fix.

You're correct that `saveOpenWikiEnv` strips empty values (`nextEnv[key] === "" → delete`), so a real skip leaves `LANGSMITH_API_KEY` **absent**, not `=""`. My original "verification" was misleading me: my `~/.openwiki/.env` had a stale `LANGSMITH_API_KEY=""` from before that stripping existed, so `=== undefined` happened to pass there — but it doesn't hold for the state the skip path actually produces today. Apologies for the noise.

Gating on `LANGCHAIN_TRACING_V2` is the right call, since the step always records it (`"false"` on skip, `"true"` on entry) and it survives the empty-stripping. I refined it slightly to also avoid regressing the case where a key is supplied via a shell export (which the original `!process.env.LANGSMITH_API_KEY` honored): the step is now considered unanswered only when **neither** a key is present **nor** a tracing decision has been recorded. Extracted it into a small `needsLangSmithStep()` helper so both gates (`needsCredentialSetup` and the wizard's `needsCredentialPrompt`) share one tested definition instead of the duplicated inline condition that drifted out of sync.

```ts
export function needsLangSmithStep(env: NodeJS.ProcessEnv = process.env): boolean {
  return !env.LANGSMITH_API_KEY && env.LANGCHAIN_TRACING_V2 === undefined;
}
```

Added unit tests covering the four states: fresh (needs it), skipped (`TRACING_V2="false"`, doesn't), key entered (`TRACING_V2="true"`, doesn't), and a shell-provided key with no flag (doesn't). Re-verified the real reproduction — skip, then run — and setup no longer reopens. `build` / `vitest` / `lint:check` / `format:check` all green. Ready for another look.

colifran (approved):
Thank @SadriG91!

Issue #427, opened 2026-07-21
Connector tools throw instead of returning errors as tool results, so the model can't self-correct (e.g. wrong MCP tool name)


misterclarity:
## Summary

The built-in connector tools in `createOpenWikiConnectorTools()` (`dist/connectors/tools.js`) **throw** on error instead of returning the error text as the tool result. A thrown tool error aborts the agent run (see companion issue on the unhandled-rejection crash), and even when that crash is avoided, the model never gets to *see* the error and self-correct.

The most common trigger: the model calls `openwiki_call_mcp_tool` with a slightly-wrong tool name (e.g. `notion-get-page-content` instead of `notion-fetch`). The connector correctly throws a helpful message —

> MCP tool notion-get-page-content was not returned by tools/list for notion. Run openwiki_list_mcp_tools first and use an exact discovered name.

— which is written *for the model*, but the model never receives it as a tool result because it was thrown rather than returned.

## Environment

- OpenWiki 0.2.1
- Any provider; most frequent with smaller/local models that occasionally guess tool names.

## Suggested fix

Wrap each connector tool's `func` so thrown errors are returned as a `Tool error: <message>` result. Minimal, localized change at the end of `createOpenWikiConnectorTools()`:

```js
    ].map(wrapToolErrors);
}

function wrapToolErrors(tool) {
    const originalFunc = tool.func;
    tool.func = async (...args) => {
        try {
            return await originalFunc(...args);
        } catch (error) {
            return `Tool error: ${error instanceof Error ? error.message : String(error)}`;
        }
    };
    return tool;
}
```

Verified locally: with this wrapper, a bad MCP tool name (`notion-get-page-content`) comes back to the model as a `Tool error: ...` result carrying the "run openwiki_list_mcp_tools first / use an exact discovered name" guidance, instead of throwing. The wrapper is unchanged between 0.1.2 and 0.2.1 (`createOpenWikiConnectorTools` is identical), and applies cleanly on 0.2.1. This complements the langchain unhandled-rejection fix — returning-not-throwing is the more robust layer since it keeps the guidance in front of the model.

PR #449, opened 2026-07-23
fix: connector tools fail gracefully + operator allow/deny tool filter

### LINKED ISSUE #444: [Bug]: Connector ingest tools are exposed to code-mode agents (should be gated by run mode)
### Description

`code`-mode runs (codebase documentation, `outputMode: "repository"`) are handed the full set of external **connector ingest** tools — `openwiki_ingest_connector` (enum includes `google`, `slack`, `x`, `notion`, `hackernews`, `web-search`, `git-repo`), `openwiki_ingest_all_connectors`, and the MCP connector tools. A codebase-documentation agent reads a git repo and writes `openwiki/*.md`; it has no business ingesting Gmail/Slack/X and should never be offered those tools.

The toolset is built unconditionally: `src/agent/index.ts:250` calls `tools: createOpenWikiConnectorTools()` with no dependence on the run mode/command, and `createOpenWikiConnectorTools()` (`src/connectors/tools.ts:23`) takes no mode argument. So `personal` (local-wiki) and `code` (repository) runs get an identical toolset.

The practical failure: under a large/`--recursive` codebase run the subagent machinery fans out, and a subagent can call `openwiki_ingest_connector({connectorId:"google"})`. With no connector credentials configured (a normal state for a pure codebase-doc run, e.g. CI), that path throws `Gmail refresh token is required for OAuth refresh.` (`src/auth/tokens.ts:66-68`, reached because `isGmailEnabled` defaults to enabled at `src/connectors/sources/gmail.ts:331`). Combined with the uncaught-tool-error behavior tracked in #426/#427, a single unwanted connector call terminates the entire run — after it may have executed for a long time and produced zero output.

This issue is specifically about **mode-gating the toolset** (the tools shouldn't be present in `code` mode at all). It is complementary to, not a duplicate of:
- #427 — connector tools throw instead of returning errors as tool results (defense in depth: even a legitimately-offered connector tool should return, not throw).
- #426 — any tool error becomes an unhandled rejection that crashes the process (the systemic backstop).

Gating (this issue) removes the tools from the mode where they make no sense; #427/#426 make whatever remains survivable. All three are worth fixing.

### Steps to Reproduce

1. Build from source.
2. Run a `code`-mode codebase-documentation run (`repository` output mode) with **no** connector credentials configured, ideally `--recursive` over a large tree so subagents fan out.
3. If any (sub)agent calls `openwiki_ingest_connector({connectorId:"google"})` (or `openwiki_ingest_all_connectors`), the connector path is entered even though this is a codebase-doc run.

### Expected Behavior

`code`-mode (codebase documentation, `outputMode: "repository"`) runs should not be offered external connector-ingest tools at all. Connector ingestion belongs to the `personal`/connector workflow (`local-wiki`).

### Actual Behavior

`code`-mode agents receive the full connector-ingest toolset (`openwiki_ingest_connector`, `openwiki_ingest_all_connectors`, MCP connector tools). A subagent can invoke Gmail/Slack/X ingestion during what should be a pure git-repo → markdown run; with no credentials this throws and (via #426) can crash the whole run.

### Environment

- OS: Linux (ubuntu-latest CI); reproducible cross-platform
- Node.js version: 22.x
- OpenWiki version: 0.2.x (main); `createOpenWikiConnectorTools` call site unchanged across 0.1.x–0.2.x

### Additional Context

Relevant source (main):
- `src/agent/index.ts:250` — `tools: createOpenWikiConnectorTools()` (unconditional; `outputMode` is already in scope at `runOpenWikiAgentCore` line ~213 and is the natural gating axis — `"repository"` = code mode).
- `src/connectors/tools.ts:23` — `createOpenWikiConnectorTools()` takes no mode arg; `openwiki_ingest_connector` enum at lines 96–104.
- `src/connectors/tools.ts:241-248` — `ingestConnector` calls `registry[connectorId].ingest(options)` with no try/catch. Note `ingestAllConnectors` (lines 270–281) is **also** an unguarded loop — neither single- nor all-connector ingest isolates a per-connector failure (relevant to #427).
- `src/connectors/sources/gmail.ts:122` → `getOAuthAccessToken("gmail")`; `:331` `isGmailEnabled` defaults to enabled.
- `src/auth/tokens.ts:66-68` — throws when refresh token is absent.

Suggested fix: pass the run mode/`outputMode` into `createOpenWikiConnectorTools()` (or filter at the `createDeepAgent` call site) so `repository`/code runs get no connector-ingest tools. A per-mode allowlist would also cover the MCP connector tools.

ppsplus-bradh:
Closes #444.

## What & why

A `code --update --recursive` documentation run could crash **uncaught** when a subagent invoked a connector-ingest tool without credentials configured — e.g. `openwiki_ingest_connector({connectorId:"google"})` throwing `Gmail refresh token is required for OAuth refresh.` The throw escaped the tool call and propagated through the langgraph stream, terminating the whole run after a long, expensive execution that produced zero output.

## Approach — a note on direction

The issue title suggested *gating connector tools out of code mode*. While implementing that, it became clear a blanket mode-gate **over-removes `git-repo`**, which this repo's own docs (`openwiki/integrations/connectors.md`) describe as legitimately relevant to code mode ("documenting a different target repo than the one being ingested from"). Deciding code mode should have *no* connectors would contradict documented behavior.

So this PR takes a more robust and less presumptuous route: **make tool calling resilient, and let the operator decide** which tools are exposed. This fixes the crash for *all* modes (not just code) and adds no behavior change by default.

## Changes

1. **Resilience — connector tools fail gracefully.** Every connector tool's `func` is wrapped so a thrown error becomes a structured, model-visible result (`Tool error: …` / `{status:"error"}`) instead of an uncaught crash. `ingestConnector` / `ingestAllConnectors` keep per-connector isolation (one failure doesn't abort the rest). **Abort and LangGraph graph-interrupt signals are re-thrown, never swallowed**, so cancellation and human-in-the-loop control flow keep working (`isUncatchableError` mirrors langgraph's `is_bubble_up` marker + `GraphInterrupt`/`NodeInterrupt` names + `AbortError`).

2. **Operator control — allow/deny tool filter.** `createOpenWikiConnectorTools(filter?)` accepts `{allow?, deny?}` by tool name (deny wins; empty allow = all; **default = all seven tools, no behavior change**). Wired via `--allow-tools` / `--deny-tools` CLI flags (space and `=` forms) and `OPENWIKI_ALLOW_TOOLS` / `OPENWIKI_DENY_TOOLS` env vars (CLI wins over env) across **both** the `run` path and the scheduled `ingest` (cron) path — so operators can restrict tools in unattended CI runs, e.g. `OPENWIKI_DENY_TOOLS=openwiki_ingest_connector,openwiki_ingest_all_connectors`.

3. **Prompt gating by actual toolset.** The "Connector ingestion discipline" block is only emitted when connector tools are actually present, so the agent is never instructed to call tools it doesn't have.

## Relationship to #426 / #427

Complementary. #426 (unhandled-rejection crash) and #427 (connector tools should return errors, not throw) address the framework/tool-result layer generally; this PR adds the connector-specific graceful-failure wrapper (with abort/interrupt safety) plus operator-facing tool control and prompt consistency. The graceful-failure wrapper here overlaps #427's intent for connector tools specifically.

## Testing

- `pnpm run typecheck`, `pnpm run lint:check`, `pnpm run format:check` — all clean.
- `pnpm run test` — 375 tests pass (34 files).
- New/extended tests cover: allow/deny filtering (deny-wins, allow-whitelist, default-all), graceful failure returning structured errors, **abort/interrupt correctly rejecting (not downgraded)**, `ingestAllConnectors` partial success, CLI/`=`-form flag parsing on both `run` and `ingest`, and prompt block presence/absence by toolset.


src/connectors/tools.ts
```diff
@@ -198,7 +211,102 @@ export function createOpenWikiConnectorTools(): StructuredToolInterface[] {
...
+ * `NodeInterrupt` error names. AbortError is Node/DOM's standard cancellation
+ * signal name.
+ */
+function isUncatchableError(error: unknown): boolean {
+  if (!isRecord(error)) {
+    return false;
+  }
+
+  // AbortController / cancellation.
+  if (error.name === "AbortError") {
```
colifran, line 299: nice - great catch here.

src/connectors/tools.ts
```diff
@@ -198,7 +211,102 @@ export function createOpenWikiConnectorTools(): StructuredToolInterface[] {
...
+    return true;
+  }
+
+  // LangGraph GraphBubbleUp marker (GraphInterrupt, NodeInterrupt, GraphDrained,
+  // ParentCommand all set this getter to true).
+  if (error.is_bubble_up === true) {
```
colifran, line 305: nice catch here as well. thanks for handling these

colifran:
In general a big focus for me is trying to prevent unnecessary crashes and making things degrade gracefully so thank you for getting started on this.

PR #450, opened 2026-07-23
fix: open oauth urls on windows via rundll32 (cmd /c start truncates at '&')

### LINKED ISSUE #425: Windows: OAuth browser URL truncated at '&' by 'cmd /c start' (Notion shows 'Missing redirect_uri', CLI hangs)
## Summary

On Windows, every `openwiki auth <provider>` OAuth flow opens a browser URL that is **truncated at the first `&`**, so the authorization server receives only the first query parameter. For providers where `redirect_uri`/`state`/etc. come after `client_id`, the provider rejects the request — with Notion the browser shows **"Missing redirect_uri in OAuth request"** and the CLI hangs forever waiting for a callback that can never arrive.

## Environment

- OpenWiki 0.2.1 (also present in 0.1.2)
- Windows 11, Node 24
- Any OAuth provider; most visible with `openwiki auth notion`

## Root cause

`dist/auth/oauth.js` → `openBrowser()`:

```js
if (platform === "win32") {
    await execFilePromise("cmd", ["/c", "start", "", url]);
    return true;
}
```

`cmd.exe /c start` treats `&` as a command separator. Even though `url` is passed as a single `execFile` argument, `cmd /c` re-parses the reconstructed command line, so a URL like

```
https://.../authorize?client_id=abc&redirect_uri=http://127.0.0.1:53682/callback&state=...
```

is split at the first `&`; everything from `redirect_uri=` onward is dropped (or interpreted as separate commands). The authorization endpoint then sees a request with no `redirect_uri`.

## Reproduction

1. On Windows, `openwiki auth notion`.
2. Select and authorize a workspace.
3. Browser lands on an error page ("Missing redirect_uri in OAuth request"); the CLI never returns and must be Ctrl+C'd.

## Suggested fix

Open the URL without going through `cmd`'s parser. `rundll32` passes the argument through verbatim:

```js
if (platform === "win32") {
    await execFilePromise("rundll32", ["url.dll,FileProtocolHandler", url]);
    return true;
}
```

(Alternatively `execFile("cmd", ["/c", "start", '""', '"' + url + '"'])` with the URL explicitly quoted, but `rundll32` is simpler and avoids `start`'s title-argument quirk.) Verified locally on Windows 11 — the full URL now reaches the browser and Notion/other OAuth flows complete.

ashudhanda:
## Summary

Closes #425

On Windows, `openBrowser()` used `cmd /c start "" <url>`. `cmd.exe` re-parses the reconstructed command line and treats `&` as a command separator, so every OAuth authorization URL was truncated at the first query parameter — the provider received only `client_id`, Notion showed **"Missing redirect_uri in OAuth request"**, and the CLI hung forever waiting for a callback.

## Change

Replace the `cmd /c start` invocation with:

```ts
await execFilePromise("rundll32", ["url.dll,FileProtocolHandler", url]);
```

as suggested in the issue. `rundll32` passes the URL through verbatim (no shell re-parsing), avoids `start`'s title-argument quirk, and requires no quoting gymnastics. macOS/Linux paths are untouched, and the existing `try/catch` fallback (printing the URL for manual open) still applies if `rundll32` is unavailable.

One-line behavioral change, scoped to the `win32` branch of `openBrowser()` in `src/auth/oauth.ts`.

### DIFF
--- a/src/auth/oauth.ts
+++ b/src/auth/oauth.ts
@@ -559,7 +559,11 @@ async function openBrowser(url: string): Promise<boolean> {
     }
 
     if (platform === "win32") {
-      await execFilePromise("cmd", ["/c", "start", "", url]);
+      // "cmd /c start" re-parses the reconstructed command line, so "&" in a
+      // URL acts as a command separator and truncates the authorization URL
+      // at the first query parameter. rundll32's FileProtocolHandler receives
+      // the URL as a verbatim argument and avoids start's title-arg quirk.
+      await execFilePromise("rundll32", ["url.dll,FileProtocolHandler", url]);
       return true;
     }
 


colifran (approved):
Thanks @ashudhanda!

PR #453, opened 2026-07-23
fix: preserve contested local-wiki claims

### LINKED ISSUE #380: Personal mode silently resolves contradictory sources by overwrite — preserve both claims as contested
### Problem

In local-wiki (personal) mode, when two credible sources make incompatible claims about the same fact, the synthesis guidance currently resolves the conflict by overwrite:

- The update discipline says to update entries in place ("Prefer replacing one stale sentence over adding new paragraphs"; "Update existing theme, open-question, and commitment entries instead of repeating the same detail on multiple source pages").
- The only contradiction-aware mechanism is `/open-questions.md`, where "contradictory evidence" is a listed trigger — but that records that *a question exists*, not *what the competing claims are*. The canonical page keeps exactly one version of the fact, and nothing says which claim survives, so in practice the most recently ingested source wins.

The existing "prefer current source evidence" rule is the right arbiter when docs conflict with code or git history — including personal mode's configured local repositories, where code is ground truth. But personal-mode knowledge sources are also people, posts, and documents that legitimately disagree with *each other*, and for those source-vs-source conflicts there is no arbiter: silent overwrite makes the wiki assert a disputed claim as `source-backed` and discards the losing claim's provenance entirely.

Example: a project contributor's post says a rewrite took 6 days; the company's launch post says eleven. Both are first-party, both have incentives, neither is authoritative over the other, and neither is stale. Today the wiki keeps whichever was ingested last.

### Proposal

Extend the local-wiki synthesis discipline (`localWikiSynthesisInstruction` in `src/agent/prompt.ts`):

1. A fifth confidence label, `contested`: incompatible claims from credible sources that current evidence does not settle.
2. A "Contested knowledge discipline" block: when no ground truth settles the conflict and neither source is clearly stale, record both claims — each with source and date — in a `## Contested` section on the canonical page, and label the disputed fact `contested` wherever it appears (including `/themes.md` Confidence cells). Never present either claim as `confirmed`/`source-backed` while the conflict is open.
3. Interaction with `/open-questions.md`: the Contested entry is the record of the conflict; an open question is added only when the conflict would impair future assistance, and links to the Contested entry instead of restating it. The existing "contradictory evidence" trigger is amended to point here, so the prompt gives one route, not two.
4. Resolution rule: only on new evidence or a shown-stale source, never recency alone — and the entry collapses to a one-line record (resolution date, deciding evidence, superseded claim + source) so the losing claim's provenance survives resolution too.

Scope: local-wiki mode only (`localWikiSynthesisInstruction` is empty in repository mode). Docs-vs-code conflicts keep resolving by current source evidence in both modes, unchanged.

Cost: the block adds roughly 300 tokens (~4%) to the local-wiki system prompt. Happy to compress further if that's a concern.

We run this convention ("Contested sections: both claims, sources, dates — never a silent overwrite") in a private agent-maintained knowledge vault of ~400 pages across 13 wikis; it has repeatedly saved us from confidently-wrong pages when sources disagreed (the 6-vs-11-days example is a real case from our vault).

I have a tightly-scoped PR ready (prompt guidance + tests, `format`/`lint`/`test` green) and will open it if this direction sounds right.

*Drafted with an AI agent; reviewed and submitted by me.*

ousamabenyounes:
## Summary

Fix #380

Local-wiki synthesis now has an explicit `contested` confidence label and a contested-knowledge discipline. When credible personal-mode sources disagree and current evidence does not settle the conflict, the prompt tells the agent to preserve both claims with source/date provenance in a canonical `## Contested` section instead of silently overwriting one side.

The guidance is scoped to local-wiki mode; repository mode keeps the existing prompt behavior.

## Test verification (RED → GREEN)

RED on upstream-base behavior with the new prompt regression test:

```text
FAIL  test/prompt.test.ts > createSystemPrompt filesystem path guidance > local-wiki mode > preserves unresolved source conflicts as contested knowledge
AssertionError: expected 'You are OpenWiki, an expert technical…' to contain 'contested:'
Test Files  1 failed (1)
Tests  1 failed | 15 skipped (16)
```

GREEN with the fix:

```text
pnpm exec vitest run test/prompt.test.ts --testNamePattern 'preserves unresolved source conflicts as contested knowledge'
Test Files  1 passed (1)
Tests  1 passed | 15 skipped (16)
```

## Local validation

```text
pnpm run format:check
All matched files use Prettier code style!

pnpm run lint:check
# passed

pnpm run typecheck
# passed

pnpm test
Test Files  44 passed (44)
Tests  512 passed (512)

pnpm run build
# passed

OPENWIKI_DEV=1 node dist/cli.js code --dry-run --init
# passed; printed the dry-run execution plan and did not invoke the agent
```

### DIFF
--- a/src/agent/prompt.ts
+++ b/src/agent/prompt.ts
@@ -410,8 +410,14 @@ function getOutputPromptConfig(
 - Apply confidence labels consistently:
   - confirmed: directly supported by authoritative evidence or repeated high-quality evidence.
   - source-backed: supported by one credible source but not yet independently confirmed.
+  - contested: incompatible claims from credible sources that current evidence does not settle.
   - watchlist: weak, low-signal, early, or potentially transient evidence worth checking again.
   - saved-context: useful context intentionally saved by the user or found in bookmarks, without implying it is true or important.
+- Contested knowledge discipline:
+  - When credible personal-mode sources disagree and no ground truth settles the conflict, preserve both claims in a ## Contested section on the canonical page. Include each claim's source and date when available.
+  - Label the disputed fact contested wherever it appears, including /themes.md Confidence cells. Never present either side as confirmed or source-backed while the conflict remains unsettled.
+  - Add an /open-questions.md entry only when the unresolved conflict would impair future assistance, and link that question to the canonical Contested entry instead of restating both claims.
+  - Never resolve a contested fact by recency alone. Resolve it only when new evidence settles the conflict or shows that a source is stale, then keep a short resolution note with the resolution date, deciding evidence, and superseded claim source.
 - Classify email-like evidence before writing it to the wiki. Use these labels: action_required, scheduled_commitment, decision_or_approval, direct_request, important_update, people_or_org_signal, project_context, security_or_account_notice, newsletter_or_digest, transaction_or_receipt, promotion_or_marketing, personal_logistics, noise.
 - For email-like evidence, also assign priority high, medium, low, or ignore, and durability ephemeral, durable, or recurring. Write only high/medium durable items, action items, scheduled commitments, approvals, personal logistics, and recurring patterns. Keep receipts, promotions, generic newsletters, routine security notices, and noise out of the wiki unless they are actionable, recurrent, or explicitly requested.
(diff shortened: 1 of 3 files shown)


colifran (approved):
Thank you for the PR! Looks good. I added the contested label into ingestion synthesis policy prompt.

PR #454, opened 2026-07-23
fix: support Bedrock default credential chain

### LINKED ISSUE #443: Bedrock provider should fall back to the AWS SDK default credential chain (SSO / assume-role / instance role)
## Summary

The Bedrock provider always constructs `ChatBedrockConverse` with an explicit `credentials: { accessKeyId, secretAccessKey }` object, and several guards hard-require `BEDROCK_AWS_ACCESS_KEY_ID`. This makes Bedrock unusable for anyone who authenticates with **temporary / non-static** AWS credentials — AWS IAM Identity Center (SSO), `sts assume-role`, or EC2/EKS instance roles — which is the norm in most AWS orgs where long-lived static IAM keys are discouraged or disabled by policy. There is also no field for a session token, so even manually-exported temporary credentials fail.

## Repro

1. Authenticate via SSO / assume a role (credentials are an `ASIA…` access key **plus a session token**, no static keys; e.g. `aws configure sso` then `export AWS_PROFILE=myprofile`).
2. Set `OPENWIKI_PROVIDER=bedrock`, `BEDROCK_AWS_REGION=us-west-2`, `OPENWIKI_MODEL_ID=us.anthropic.claude-sonnet-5`.
3. Run `openwiki code -p "hi"`.

### Result (two distinct failure modes)

- With only `AWS_PROFILE` set: `BEDROCK_AWS_ACCESS_KEY_ID is required for non-interactive runs.` — a guard rejects the run before any AWS call is made.
- If you export the temporary `BEDROCK_AWS_ACCESS_KEY_ID` / `BEDROCK_AWS_SECRET_ACCESS_KEY` from the SSO session: `UnrecognizedClientException: The security token included in the request is invalid.` — the session token is dropped.

## Root cause

- `getMissingProviderEnvKey` (`src/constants.ts`) treats `BEDROCK_AWS_ACCESS_KEY_ID` as mandatory; this drives the non-interactive guards in `startup.ts`, `cli.tsx`, and `agent/index.ts`.
- The `provider === "bedrock"` branch in `agent/index.ts` passes an explicit `credentials` object with only `accessKeyId` / `secretAccessKey` (no `sessionToken`). Passing an explicit object also prevents the AWS SDK's `fromNodeProviderChain()` from ever running.

## Suggested fix

Treat static keys as optional for Bedrock and let the AWS SDK resolve credentials when they are absent:

1. In `getMissingProviderEnvKey`, return `null` for `bedrock` (static keys optional).
2. In the Bedrock client builder, omit `credentials` entirely when no static keys are set, so `@langchain/aws` falls back to the AWS SDK default credential provider chain (`AWS_PROFILE` / SSO / assume-role / instance role) — which also auto-refreshes on expiry. When static keys **are** set, additionally pass `sessionToken` (from `AWS_SESSION_TOKEN`, or a new `BEDROCK_AWS_SESSION_TOKEN`) so manually-exported temporary credentials also work.

```ts
// agent/index.ts, provider === "bedrock" branch
const accessKeyId = getProviderApiKey(provider);
const secretAccessKey = secretKeyEnvKey ? process.env[secretKeyEnvKey] : undefined;
const hasStaticKeys = Boolean(accessKeyId && secretAccessKey);
  ...(hasStaticKeys
    ? { credentials: { accessKeyId, secretAccessKey,
        sessionToken: process.env.AWS_SESSION_TOKEN || process.env.BEDROCK_AWS_SESSION_TOKEN || undefined } }
    : {}), // else: AWS SDK default credential provider chain
  model: modelId,
  ...retryOptions,
});
```

## Verified locally (openwiki v0.2.2)

With these two changes applied to the installed package, `AWS_PROFILE=<sso-profile> BEDROCK_AWS_REGION=us-west-2 OPENWIKI_PROVIDER=bedrock OPENWIKI_MODEL_ID=us.anthropic.claude-sonnet-5 openwiki code -p "…"` authenticates and runs with **zero exported credentials**. The static-key + session-token path also works.

## Environment

- openwiki v0.2.2
- Provider: AWS Bedrock (`us.anthropic.claude-sonnet-5`), region `us-west-2`
- macOS, Node 26

ousamabenyounes:
## Summary

Fix #443

Bedrock no longer requires static `BEDROCK_AWS_ACCESS_KEY_ID` / `BEDROCK_AWS_SECRET_ACCESS_KEY` values before a run can start. When no complete static key pair is configured, OpenWiki omits the explicit `credentials` option so `ChatBedrockConverse` can use the AWS SDK default credential chain, including AWS profile, SSO, assume-role, and instance-role credentials.

When a complete static key pair is configured, OpenWiki still passes it explicitly and now includes temporary credential session tokens from `BEDROCK_AWS_SESSION_TOKEN` or `AWS_SESSION_TOKEN`. Those session-token keys are also managed and redacted in diagnostics.

## Test verification (RED → GREEN)

RED on upstream-base behavior with the Bedrock credential regression tests:

```text
FAIL  test/bedrock-credentials.test.ts > createModel bedrock credentials > omits explicit credentials so the AWS SDK default chain can run
AssertionError: expected { credentials: { …(2) }, …(3) } to not have property "credentials"

FAIL  test/bedrock-credentials.test.ts > createModel bedrock credentials > passes session tokens with complete static temporary credentials
AssertionError: expected { accessKeyId: 'access', …(1) } to deeply equal { accessKeyId: 'access', …(2) }

FAIL  test/constants.test.ts > bedrock provider (AWS SDK credentials + region) > does not require static Bedrock keys when using the AWS SDK credential chain
AssertionError: expected 'BEDROCK_AWS_ACCESS_KEY_ID' to be null
Test Files  2 failed (2)
Tests  6 failed | 4 passed | 55 skipped (65)
```

GREEN with the fix:

```text
pnpm exec vitest run test/constants.test.ts test/bedrock-credentials.test.ts test/redaction.test.ts --testNamePattern 'Bedrock|bedrock|static|BEDROCK_AWS_SESSION_TOKEN'
Test Files  3 passed (3)
Tests  11 passed | 93 skipped (104)
```

## Local validation

```text
pnpm run format:check
All matched files use Prettier code style!

pnpm run lint:check
# passed

pnpm run typecheck
# passed

pnpm test
Test Files  45 passed (45)
Tests  517 passed (517)

pnpm run build
# passed

OPENWIKI_DEV=1 node dist/cli.js code --dry-run --init
# passed; printed the dry-run execution plan and did not invoke the agent
```


src/constants.ts
```diff
@@ -479,6 +496,10 @@ export function getProviderSecretKeyEnvKey(
 }
 
 export function providerRequiresSecretKey(provider: OpenWikiProvider): boolean {
+  if (provider === "bedrock") {
+    return false;
```
breathOfTech, line 500: Supporting the default AWS credential chain here is correct, but this creates
an interactive setup regression because the same helper also controls the
setup-wizard topology.

Bedrock still has `apiKeyEnvKey: BEDROCK_AWS_ACCESS_KEY_ID`, so its primary
wizard step remains `api-key`. Returning `false` here removes only the
`secret-key` step. The resulting `code --init` flow is:

1. AWS Bedrock provider
2. `BEDROCK_AWS_ACCESS_KEY_ID`
3. `BEDROCK_AWS_REGION`
4. model ID

There is no `BEDROCK_AWS_SECRET_ACCESS_KEY` prompt.

I reproduced this end to end on this PR:

- With no explicit AWS or `BEDROCK_AWS_*` credentials, an AWS profile/default
  chain invocation succeeded and returned `BEDROCK_PROFILE_AUTH_OK`, confirming
  the main fix works.
- I then ran `pnpm dev code --init` and entered a fake access key.
- The wizard skipped directly to region/model.
- The final run failed with:
  `BEDROCK_AWS_SECRET_ACCESS_KEY is required to run OpenWiki with AWS Bedrock.`

Could the default-chain path omit both static-key prompts, with a separate
explicit-credentials path that collects both access key and secret key and
rejects partial pairs?

test/credentials.test.ts
```diff
@@ -194,11 +194,10 @@ describe("orderedSetupSteps", () => {
     );
   });
 
-  test("bedrock adds secret-key and region before model", () => {
+  test("bedrock adds region before model without requiring a secret key", () => {
     const spine = orderedSetupSteps("bedrock", "code", false);
-    expect(spine).toContain("secret-key");
+    expect(spine).not.toContain("secret-key");
```
breathOfTech, line 199: The default-chain intent makes sense, but this removes only the `secret-key`
step while Bedrock’s `api-key` step remains. The wizard can therefore save an
access key without its matching secret, producing a partial pair that the
runtime rejects.

Could the default-chain path remove both static-key prompts, with a separate
explicit-credentials path that collects both values and rejects partial pairs?

breathOfTech (changes requested):
I validated the main change successfully with a real AWS profile/default
credential chain and confirmed that Bedrock invocation works without static
credentials.

I also reproduced an interactive setup regression, documented inline: the
wizard still collects an access key but no longer collects its matching
secret, allowing a partial pair to be saved and causing the subsequent run to
fail.

Requesting changes to keep the default-chain path and explicit-credential path
consistent, with regression coverage for partial credential pairs.

colifran:
Thanks for the PR. I believe this has been resolved via #418

PR #455, opened 2026-07-24
feat: implement native wiki visualizer for openwiki

colifran:
## Summary

Adds an openwiki visualize command that serves a generated wiki as an interactive node graph with a live-reloading markdown reader.

- New subcommand openwiki visualize [path] [--port <port>] [--no-open] — defaults to ./openwiki, port 4321 (increments on conflict), opens the browser unless --no-open.
- Loopback-only HTTP server (src/visualize/server.ts): binds 127.0.0.1 exclusively, serves fixed string routes (no filesystem path is ever derived from the URL), and pushes live reloads over SSE backed by fs.watch (debounced). Ships a strict CSP and loads the graph/markdown/diagram libs from jsdelivr with SRI-pinned hashes.
- Browser client as testable TS modules (client.ts, client-lib.ts, page.ts, graph.ts) compiled via a separate DOM-enabled tsconfig.client.json. Wiki markdown is rendered through DOMPurify.sanitize(marked.parse(...)) before innerHTML as defense-in-depth on top of the CSP; scalar fields are HTML-escaped.
- README: documents the command, loopback behavior, and the CDN network requirement.

## Tests

- Unit tests
- Manual E2E tests

## Demo

https://github.com/user-attachments/assets/1941edff-b70b-4807-87dd-96e8da88de55


ppsplus-bradh:
This is sweet.

jyje:
Awesome, we don't need Obsidian anymore!

Bennoo:
This is exactly what is missing when working locally with the generated openwiki!

When can we expect a merge for this one?

colifran:
@Bennoo hoping very soon!

bracesproul (dismissed):

Issue #458, opened 2026-07-24
[Bug]: Truncated model tool calls can end runs without an error


geonwoo-jeong:
### Description

When a model reaches its output-token limit while streaming a large tool call, OpenWiki can finish the run without surfacing why the requested write or edit did not happen.

This is easiest to reproduce on larger repository updates where a `write_file` or `edit_file` call contains a substantial Markdown payload. Smaller edits usually remain within the output budget.

### Steps to Reproduce

1. Configure OpenWiki with the Bedrock provider and a tool-capable model.
2. Run `openwiki code --update --print` on a repository update that requires generating a large new page or multi-line edit.
3. Let the model reach its output-token limit while emitting the filesystem tool-call arguments.

### Expected Behavior

OpenWiki should fail the run with an actionable error explaining that the model reached its output limit and the incomplete tool call was not executed.

### Actual Behavior

The LangChain v3 stream emits both:

- `content-block-finish` with `content.type = "invalid_tool_call"`
- `message-finish` with `reason = "length"`

OpenWiki currently ignores these non-text message events. The incomplete tool call is not executed, but the outer run can finish without an error, leaving the requested file change missing.

### Environment

- OS: Linux (GitHub Actions)
- Node.js version: 22.x
- OpenWiki version: current `main` (`c95b6d6`)
- Provider: AWS Bedrock
- LangChain AWS package: `@langchain/aws@1.4.2`

### Additional Context

A stock `@langchain/aws@1.4.2` stream reproduction confirms that Bedrock `stopReason = "max_tokens"` maps to `message-finish.reason = "length"`, and a partially accumulated JSON tool payload finalizes as `invalid_tool_call`. Tool arguments may contain generated file contents and should not be included in the surfaced error.

PR #459, opened 2026-07-24
feat: configure model output and bedrock stream limits

### LINKED ISSUE #474: [Feature]: Make model output token limit configurable
### Summary

Allow users to configure the maximum output-token budget used when OpenWiki creates model clients.

### Motivation

OpenWiki currently relies on each provider client's default output-token limit. Large generated pages or multi-line edits can require a larger output budget, while deployments may also want an explicit lower limit for cost or latency control.

The truncation-detection behavior is tracked separately in #458; this request is only for configuring the model-client limit.

### Proposed Solution

Add an optional `OPENWIKI_MAX_OUTPUT_TOKENS` setting that:

- accepts a positive integer;
- maps to `maxOutputTokens` for Google clients and `maxTokens` for the other supported LangChain chat clients;
- preserves every provider's existing default when unset.

Provider and model limits should continue to apply.

### Alternatives Considered

Relying on provider defaults does not let users adapt the budget to their workload. Provider-specific settings would add unnecessary configuration surface for the same cross-provider concept.

### Additional Context

Implementation: #459
Related truncation handling: #457 and #458

### LINKED ISSUE #485: [Feature]: Make Bedrock stream idle timeout configurable
### Summary

Allow Bedrock users to configure the first/inter-chunk idle timeout used by `ChatBedrockConverse`.

### Motivation

`@langchain/aws@1.4.2` added a 60,000 ms watchdog around Bedrock Converse streams. It applies while waiting for both the first chunk and each subsequent chunk, and currently raises:

```text
Bedrock Converse stream timed out after 60000 ms without receiving a chunk.
```

OpenWiki does not pass `streamIdleTimeout` when it creates `ChatBedrockConverse`, so Bedrock users cannot tune this provider setting. We observed the timeout on a repository generation after increasing the output-token budget. A larger token budget is not established as the cause of first-chunk latency; the two controls are independent, but users who configure longer model outputs also need a way to set an appropriate stream watchdog.

### Proposed Solution

Add an optional `OPENWIKI_STREAM_IDLE_TIMEOUT` setting that:

- accepts an integer number of milliseconds in Node's timer range;
- passes the value only to `ChatBedrockConverse` as `streamIdleTimeout`;
- accepts `0`, matching LangChain's documented way to disable the watchdog;
- preserves the `@langchain/aws` default when unset;
- appears in environment diagnostics and debug output as a non-secret value.

Documentation should recommend a sufficiently long finite timeout over `0` so a genuinely stalled stream cannot hang forever.

### Alternatives Considered

- Hard-code a longer OpenWiki timeout: this would replace one universal default with another and remove the upstream default's ability to evolve.
- Disable the watchdog unconditionally: this would restore the indefinite-hang behavior the LangChain guard was introduced to prevent.
- Treat this as part of truncation detection: the timeout can occur before OpenWiki receives the protocol events used by truncation detection, so it is a separate configuration concern.

### Additional Context

- Related output-token setting: #474
- Implementation PR to extend: #459
- Related truncation handling: #457 and #458
- LangChain implementation and rationale: https://github.com/langchain-ai/langchainjs/pull/11098

geonwoo-jeong:
Closes #474
Closes #485

Context: #458 and #457 cover truncated tool-call detection. This PR stays separate from that failure guard and only exposes model-client runtime limits.

## Summary

- Add an optional `OPENWIKI_MAX_OUTPUT_TOKENS` setting across supported model providers.
- Add an optional Bedrock-only `OPENWIKI_STREAM_IDLE_TIMEOUT` setting for the first/inter-chunk watchdog.
- Preserve provider defaults when either setting is unset.
- Keep the Bedrock-only setting inert for other providers, including validation and diagnostics.

## Why these controls belong together

OpenWiki currently exposes neither the model output budget nor Bedrock's stream idle watchdog. The two limits are independent: `maxTokens` controls the allowed response length, while `streamIdleTimeout` controls how long the Bedrock client waits between chunks.

We observed the 60-second LangChain watchdog after enabling a larger output budget, but this PR does not claim that a larger budget causes first-chunk latency. It exposes both controls so operators can configure the output budget and the watchdog appropriate to their workload without replacing either provider default globally.

### `OPENWIKI_MAX_OUTPUT_TOKENS`

- Must be a positive integer.
- Maps to `maxOutputTokens` for Google clients.
- Maps to `maxTokens` for Anthropic, OpenAI, OpenAI-compatible, OpenRouter, Bedrock, Gemini Enterprise non-Google surfaces, ChatGPT OAuth, and GitHub Copilot.
- Remains absent when unset.
- Provider and model limits still apply; unsupported values may be rejected.

### `OPENWIKI_STREAM_IDLE_TIMEOUT`

- Applies only when the selected provider is Bedrock.
- Must be an integer from `0` to `2147483647` milliseconds.
- Maps to `ChatBedrockConverse.streamIdleTimeout`.
- `0` uses LangChain's documented watchdog-disable behavior and produces a diagnostic warning that stalled streams may hang indefinitely.
- Remains absent when unset, preserving the `@langchain/aws` default.
- Stale or invalid values are ignored while another provider is active.

## Scope

This PR changes model construction, environment management/diagnostics, documentation, and focused tests for these two runtime limits. It does not patch dependencies, alter prompts or workflows, or add truncation detection.

## Validation

- Constructor-level coverage for Bedrock, Gemini, Gemini Enterprise, Anthropic, OpenAI, OpenAI-compatible, OpenRouter, ChatGPT OAuth, and GitHub Copilot.
- Copilot regression coverage preserves GPT-5 Responses API routing and non-GPT-5 chat-completions routing while applying the configured output limit.
- Provider-aware regression coverage confirms the Bedrock timeout is ignored by other providers.
- Validation and diagnostics coverage includes unset, valid, invalid, maximum timer, and watchdog-disabled values.
- `pnpm install --frozen-lockfile`
- `pnpm test` — 64 files, 792 tests
- `pnpm run build`
- `pnpm run typecheck`
- `pnpm run lint:check`
- `pnpm run format:check`
- `pnpm audit --prod` — no known vulnerabilities


aravindrp:
@geonwoo-jeong I was getting the same issue and it was blocking an initial openwiki document creation for a repo I  was testing. I just tested this fix locally and with the increased timeout I was able to complete the oepration. Thanks! Hope this gets merged and released soon.

colifran (approved):
nice! thanks @geonwoo-jeong!

PR #466, opened 2026-07-24
fix: cap retry-after delays for connector retries

### LINKED ISSUE #464: [Bug]: Connector retries honor unbounded Retry-After values and can stall ingestion for days
## Description

`fetchWithResilience()` defines `MAX_BACKOFF_DELAY_MS = 20_000`, but only applies that cap to exponential-jitter delays. A valid `Retry-After` response header bypasses the cap.

An upstream API can make Gmail, Slack, X, Hacker News, or HTTP-MCP ingestion sleep for hours or days on every retry. This contradicts the helper's stated bounded-backoff behavior and can leave scheduled ingestion jobs stuck far longer than intended.

## Steps to Reproduce

Run this from the repository root:

```powershell
pnpm exec tsx -e "import { fetchWithResilience } from './src/connectors/http.ts'; (async()=>{const waits=[]; globalThis.fetch=async()=>new Response('rate-limited',{status:429,headers:{'retry-after':'86400'}}); await fetchWithResilience('https://example.test',{},{maxRetries:1,sleep:async(ms)=>{waits.push(ms)}}); console.log(JSON.stringify({waits}));})()"
```

## Expected Behavior

Retries should honor a server's `Retry-After` value only within a documented local maximum. A single remote response must not suspend an OpenWiki ingestion process indefinitely.

## Actual Behavior

The reproduction prints:

```json
{"waits":[86400000]}
```

`86,400,000` ms is a 24-hour sleep. With the default retry count, repeated responses can keep the connector waiting for multiple days.

The fallback jitter path is capped:

```ts
const ceiling = Math.min(MAX_BACKOFF_DELAY_MS, baseDelayMs * 2 ** attempt);
```

However, the header path is not:

```ts
const delay =
  retryAfterMs ?? backoffDelayMs(attempt, baseDelayMs, random);
await sleep(delay);
```

Both numeric and HTTP-date `Retry-After` values can exceed `MAX_BACKOFF_DELAY_MS`.

## Environment

- OS: Windows 11
- Node.js version: 22.21.0
- OpenWiki version: 0.2.3 (source checkout)

## Additional Context

Clamp the computed delay regardless of source:

```ts
const delay = Math.min(
  MAX_BACKOFF_DELAY_MS,
  retryAfterMs ?? backoffDelayMs(attempt, baseDelayMs, random),
);
```

Add tests for oversized numeric and HTTP-date `Retry-After` values.

This is separate from #431 (model-provider rate limits) and #237 (provider retry configuration).

Oxygen56:
Fixes #464

## Summary

- Cap connector retry delays at the existing 20-second maximum regardless of whether the delay comes from exponential jitter or a `Retry-After` header.
- Add regression coverage for oversized numeric and HTTP-date `Retry-After` values.

## Root Cause

`fetchWithResilience` capped only the fallback exponential backoff. A parsed `Retry-After` value bypassed `MAX_BACKOFF_DELAY_MS`, so an upstream response could suspend ingestion for hours or days.

## Verification

- `pnpm exec vitest run test/fetch-with-resilience.test.ts --no-color` — 15 tests passed.
- `pnpm run format:check` — passed.
- `pnpm run lint:check` — passed.
- `pnpm run typecheck` — passed.
- `pnpm run build` — passed.
- `OPENWIKI_DEV=1 node dist/cli.js code --dry-run --init` — passed.
- `pnpm test` — 54 files and 674 tests passed.
- `git diff --check` — passed.

## Notes for Reviewers

- Existing in-range `Retry-After` behavior is unchanged.
- This does not alter retry counts, retryable status codes, timeout behavior, or network-error backoff.

### DIFF
--- a/src/connectors/http.ts
+++ b/src/connectors/http.ts
@@ -8,7 +8,7 @@
  * `fetchWithResilience` adds:
  *   - a per-request wall-clock timeout via `AbortSignal.timeout`,
  *   - bounded exponential backoff with jitter on 429 and 5xx responses,
- *     honoring a numeric or HTTP-date `Retry-After` header when present,
+ *     honoring a numeric or HTTP-date `Retry-After` header within the cap,
  *   - the same backoff on network errors (connection reset, DNS, timeout).
  *
  * Auth failures (401/403) and other 4xx are returned as-is: they are not
@@ -121,8 +121,10 @@ export async function fetchWithResilience(
           response.headers.get("retry-after"),
           Date.now(),
         );
-        const delay =
-          retryAfterMs ?? backoffDelayMs(attempt, baseDelayMs, random);
+        const delay = Math.min(
+          MAX_BACKOFF_DELAY_MS,
+          retryAfterMs ?? backoffDelayMs(attempt, baseDelayMs, random),
+        );
         // Discard the body so the connection can be reused before we retry.
         await response.body?.cancel();
         await sleep(delay);
--- a/test/fetch-with-resilience.test.ts
+++ b/test/fetch-with-resilience.test.ts
@@ -109,6 +109,33 @@ describe("fetchWithResilience", () => {
     expect(delays).toEqual([2000]);
   });
 
+  test("caps an oversized numeric Retry-After header", async () => {
+    stubFetchSequence([
+      () =>
+        new Response("slow down", {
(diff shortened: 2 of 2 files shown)


colifran (approved):
Thanks @Oxygen56!

PR #467, opened 2026-07-24
fix: isolate env behavior tests from user home on windows

### LINKED ISSUE #463: [Bug]: Windows env-behavior tests write fixture credentials to the real OpenWiki config
## Description

On Windows, `test/env-behavior.test.ts` changes `process.env.HOME` to isolate credential-file tests. However, `src/env.ts` derives `openWikiEnvPath` from `os.homedir()`, which resolves from `%USERPROFILE%` on Windows rather than `HOME`.

The test can therefore import `src/env.ts` with the real user home directory and call `saveOpenWikiEnv()` against `%USERPROFILE%\.openwiki\.env`. Cleanup removes only the temporary test directory, not the real OpenWiki configuration. This can overwrite or remove saved provider credentials and OAuth tokens on a developer machine.

## Steps to Reproduce

1. On Windows, clone the repository. Note the timestamp of `%USERPROFILE%\.openwiki\.env`, or whether it exists.
2. Run:

   ```powershell
   pnpm test
   ```

3. Observe the failure in `test/env-behavior.test.ts > test isolation > resolves the env file inside the test HOME`.
4. Check the timestamp or contents of `%USERPROFILE%\.openwiki\.env` without sharing secrets. The file may have been created or changed by test fixtures.

## Expected Behavior

Tests read and write only a per-test temporary directory. Running `pnpm test` must never create, modify, or delete a developer's real OpenWiki credentials.

## Actual Behavior

The test changes only `HOME`:

```ts
process.env.HOME = tempHome;
env = await import("../src/env.ts");
```

But `src/env.ts` calculates the paths at module load:

```ts
export const openWikiEnvDir = path.join(os.homedir(), ".openwiki");
export const openWikiEnvPath = path.join(openWikiEnvDir, ".env");
```

On Windows, `os.homedir()` does not use that `HOME` override, so fixture writes can target the real OpenWiki home.

## Environment

- OS: Windows 11
- Node.js version: 22.21.0
- OpenWiki version: 0.2.3 (source checkout)

## Additional Context

Mock `node:os` so `os.homedir()` returns `tempHome` before importing `src/env.ts`, then restore the mock after each test. This is safer than relying on platform-specific environment-variable semantics.

This is distinct from #366 (Windows ACL enforcement) and #223 (configurable OpenWiki home).

rajarshidattapy:
Description:

Fixes #463 

  ## Summary

  Fixes Windows test isolation for test/env-behavior.test.ts.

  The test now mocks os.homedir() before importing src/env.ts, preventing fixture writes from targeting %USERPROFILE%\.openwiki\.env.

  ## Changes

  - Replace the HOME environment override with a node:os mock.
  - Restore the mock after every test.
  - Update the isolation regression test and documentation comments.

  ## Validation

  - pnpm exec vitest run test/env-behavior.test.ts -t "test isolation"
  - pnpm typecheck
  - pnpm exec prettier --check test/env-behavior.test.ts


colifran (dismissed):
Nice! Thanks @rajarshidattapy!

colifran (approved):

rajarshidattapy:
@colifran thanks!

Issue #349, opened 2026-07-15
[Feature]: Support using OpenWiki without an API key by exporting it as Claude Code skills


ghost:
## Problem

Today, running OpenWiki on a repository requires configuring an OpenWiki model
provider and API key — OpenWiki's CLI *is* the LLM host and runs its own agent
loop. That's a barrier for a large group of users who already have a capable
coding agent (Claude Code) in their editor and simply want repository docs
generated and maintained, without setting up and paying for a separate
inference provider.

## Proposal

Add a first-class, spec-kit-style subcommand:

    openwiki integration claude [path]

It scaffolds two Claude Code skills — `openwiki-init` and `openwiki-update` —
into `.claude/skills/` in the target repository (defaulting to the current
directory). The user then runs `/openwiki-init` / `/openwiki-update` inside
Claude Code, and Claude Code itself generates and maintains the `openwiki/`
docs. No OpenWiki model provider or API key is required — Claude Code is the
engine.

## Scope

- **In scope:** code mode only (repository docs under `openwiki/`).
- **Out of scope:** personal mode and connectors (Gmail/Slack/Notion/X), which
  depend on OpenWiki's credentialed runtime.

## Approach

- The skill bodies are bundled, maintained templates derived from the
  repository-mode instructions in `src/agent/prompt.ts`, with the tool
  vocabulary remapped to Claude Code natives (Read/Write/Edit/Glob/Grep/Bash),
  connector/CLI/local-wiki text removed, and virtual-root `/openwiki/...` paths
  converted to real relative `openwiki/...` paths.
- This never touches the live agent's prompt path, keeping regression risk off
  the running agent.
- The subcommand sits alongside `auth`/`cron`/`ingest`/`ngrok` and is built on a
  small integration registry so other host agents can be added later.

## Acceptance criteria

- [ ] `openwiki integration claude` writes
      `.claude/skills/openwiki-init/SKILL.md` and `.../openwiki-update/SKILL.md`
      with valid Claude Code skill frontmatter.
- [ ] `openwiki integration claude <path>` targets that repository.
- [ ] The command requires no model provider or API key.
- [ ] Skill bodies use Claude-native tools and real relative `openwiki/` paths;
      no connector-only text or virtual-root paths.
- [ ] Unknown/missing agent prints a clean error and exits 1.
- [ ] Existing OpenWiki commands are unaffected.
- [ ] Unit tests cover parsing, scaffolding, idempotency, and template content.

## Notes / trade-off

[...]

PR #470, opened 2026-07-25
feat: add claude code skills integration

### LINKED ISSUE #156: Proposal: an official "Claude Code mode" — run OpenWiki keyless on the host agent's inference
## Summary

I built a Claude Code plugin that generates and maintains wikis in OpenWiki's **exact on-disk
format**, but runs on **Claude Code's own inference instead of a provider API key**. I'd like to
gauge interest in bringing this into OpenWiki as an official "Claude Code mode," or at least linking
it from the README.

Repo: https://github.com/jatinmayekar/openwiki-for-claude-code (MIT)

## Motivation

OpenWiki's engine calls a configured provider (Anthropic/OpenRouter/…) via an API key. But a large
population of users already pay for inference through Claude Code (Pro/Max/Team). For them, the
second key/bill is pure friction. Since OpenWiki's value is its **methodology + on-disk format** —
not the specific runtime — that format can be produced by the host coding agent directly, keyless.

## What it does

A plugin exposing four skills:

- `/openwiki:generate` — build a first-pass wiki (your init methodology: `openwiki/quickstart.md` +
  ≤~8 evidence-grounded section pages).
- `/openwiki:update` — git-diff-driven surgical refresh; no-op when nothing relevant changed.
- `/openwiki:ask` — answer repo questions from the maintained `openwiki/` wiki.
- `/openwiki:schedule` — set up recurring keyless updates.

The skills encode the rules from your `src/agent/prompt.ts` (targeted discovery, evidence-grounding,
one-canonical-home, section-quality caps, security rules). The current Claude Code agent does the
work with its own file/search/git tools — no LangChain, no provider key.

## Interoperability (the important part)

It writes the **same artifacts your CLI does**, so the two are interchangeable on one repo:

- `openwiki/quickstart.md` + section pages.
- `openwiki/.last-update.json` in your exact shape (`{ updatedAt, command, gitHead, model }`).
- The **byte-identical** OpenWiki reference section injected into top-level `AGENTS.md`/`CLAUDE.md`.

The plugin detects an existing `openwiki/` and continues it rather than clobbering, so an existing
OpenWiki user can `openwiki --init` in CI and then keep it fresh from Claude Code keyless, or vice
versa.

## Evidence it works (incl. a side-by-side with your CLI)

Tested end-to-end on `sindresorhus/query-string`. I ran **both** the plugin (Claude Code, keyless)
and your real CLI (OpenAI `gpt-5.4-mini`) on the same repo + the same source change (a new
`isEmpty()` export):

- Both produced a `quickstart.md` entrypoint + section pages and a valid `.last-update.json`.
- The `.last-update.json` **keys were identical** and the injected **AGENTS.md reference block was
  byte-identical** between the two engines.
- After the `isEmpty()` commit, both did a **surgical** update — adding only the new function to
  their API page, leaving unrelated pages untouched, and removing the temp `_plan.md`.
- Prose and page split differed (different models), as expected; the format contract matched.

## Question for maintainers

Would you be open to any of these? Happy to shape a PR to your CONTRIBUTING norms (one-PR-one-change,
Conventional Commits):

1. Vendor the plugin under OpenWiki (e.g. `integrations/claude-code/`) as an official "Claude Code
   mode."
2. A lighter touch: a `claude-code` provider path that shells out to a non-bare `claude -p` (uses
   the user's subscription; must avoid `--bare`, which forces `ANTHROPIC_API_KEY`).
3. Just link the standalone plugin from your README.

Happy to do the work whichever way you prefer — wanted to check direction before opening a PR.

jatinmayekar:
## What

Adds `openwiki integration claude [path]`, a command that scaffolds two Claude Code skills —
`openwiki-init` and `openwiki-update` — into `.claude/skills/` in the target repo. The user then
runs `/openwiki-init` inside Claude Code, and Claude Code itself generates and maintains the
`openwiki/` wiki using its own inference. No OpenWiki provider or API key is involved.

Scope is deliberately narrow: **code mode only**, **one command**, and the only files ever written
are under `.claude/skills/`. No new provider, no repository-file modification, no marketplace
metadata, no helper tooling.

## Why

A large group of users already have Claude Code and just want repo docs generated without setting up
and paying for a second inference provider. This closes #156 (and overlaps #349).

It's also, as far as I can tell, the only shape Anthropic's terms permit here. A Claude-subscription
*provider* inside OpenWiki would have OpenWiki route requests through the user's Pro/Max credentials,
which [Anthropic does not allow](https://code.claude.com/docs/en/legal-and-compliance) for
third-party developers. Scaffolding skills that run *inside* Claude Code, on the user's own
credentials, avoids that entirely — OpenWiki never sees a token. (This likely explains why #160,
#181 and #293 have stalled.) This is a smaller, single-concern alternative to the plugins approach in
#76 (thanks @kylin0421) — happy to close in favor of that if you'd rather consolidate.

## Design: rendered from `prompt.ts`, not vendored

The skill bodies are **generated from the same prompt the OpenWiki agent already uses**, so they
can't drift from the CLI. The shared methodology (documentation goals, OKF front-matter rules,
concept-graph modeling, section quality, coverage/backlog) is extracted from `createSystemPrompt`
into exported functions that both the live prompt and the skills renderer consume.

Two things are authored specifically for the keyless surface, because the CLI does them *outside*
the agent and a host coding agent can't invoke them:

- writing `openwiki/.last-update.json`,
- generating directory `index.md` deterministically, and
- getting OKF front matter right in place, since there's no post-run repair pass (an
  `okfFrontMatterRules(..., { repairPass })` flag toggles that one bullet).

The Claude Code surface also uses real relative paths (`openwiki/quickstart.md`) instead of
OpenWiki's virtual `/openwiki/...` paths, and its own tool vocabulary.

## Files

- `src/agent/prompt.ts` — extract 5 methodology blocks into exported functions (behavior-preserving;
  see test), add the `repairPass` flag and `MethodologyContext` type.
- `src/integrations/claude-code.ts` — the skills renderer (new).
- `src/commands.ts` — parse `integration claude [path]` + help.
- `src/cli.tsx` — one dispatch branch + handler.
- `test/claude-code-integration.test.ts` — command parsing + rendered-skill assertions (new).
- `test/prompt.test.ts` — regression test asserting `createSystemPrompt` still embeds each extracted
  block verbatim.

Most of the `prompt.ts` line count is methodology text relocated verbatim into the new functions,
not new logic.

## Test plan

- [x] `pnpm run format:check`, `pnpm run lint:check`, `pnpm run typecheck` — clean.
- [x] `pnpm test` — 685 passing (13 new). The existing suite is unchanged.
- [x] The extraction is byte-for-byte behavior-preserving: `createSystemPrompt` output is identical
  for every (surface × command) combination before and after, asserted by the new regression test
  and verified out-of-band by diffing all 12 combinations.
- [x] `pnpm dev integration claude <dir>` scaffolds both `SKILL.md` files; rendered bodies contain
  real `openwiki/...` paths, zero virtual `/openwiki/...` paths, and no connector/CLI-only
  vocabulary.

---

### DIFF
--- a/src/agent/prompt.ts
+++ b/src/agent/prompt.ts
@@ -13,6 +13,115 @@ function formatLastUpdate(lastUpdate: UpdateMetadata | null): string {
   return JSON.stringify(lastUpdate, null, 2);
 }
 
+/**
+ * Paths the shared methodology instructions reference. A structural subset of
+ * OutputPromptConfig, so any surface config satisfies it, and lightweight
+ * enough for an out-of-tree caller (e.g. the Claude Code skills renderer) to
+ * construct directly.
+ */
+export interface MethodologyContext {
+  docsLocation: string;
+  planPath: string;
+  quickstartPath: string;
+}
+
+/**
+ * Documentation goals shared across every OpenWiki output surface.
+ */
+export function documentationGoals(output: MethodologyContext): string {
+  return `Documentation goals:
+- Someone with zero knowledge of the wiki should be able to start at ${output.quickstartPath} and understand what the knowledge base covers, how it is organized, what it tracks, and where to go next.
+- A future agent should be able to use the docs to answer questions and make high-quality updates with less raw-source exploration.
+- Capture both technical details and business/product logic.
+- Explain why important code exists, not only what files contain.
+- Prefer clear Markdown with stable links between pages.
+- Organize the docs like human documentation, not a raw file inventory.
+- Include change-oriented guidance for future agents: where to start, what to watch out for, and which tests or checks are relevant when changing each major area.
+- Keep the docs concise enough to maintain. Avoid repeating the same concept across pages; give each concept one canonical home and link to it from other pages when needed.
+- Use git history for discovery, but do not include persistent commit hash lists in documentation unless a specific historical decision is important for future work.`;
+}
+
+/**
+ * OKF concept-graph relationship-modeling rules.
+ */
+export function okfRelationshipModeling(output: MethodologyContext): string {
+  return `OKF relationship modeling:
+- Treat every non-reserved Markdown document as a concept node. Standard Markdown links between concept documents are directed relationship edges; tags, resource fields, directory placement, source-code references, and index.md links do not replace concept-to-concept links.
+- Model meaningful runtime, dependency, ownership, data-flow, security, lifecycle, and user-flow relationships, not only navigation from ${output.quickstartPath}.
+- Put a concept link in the sentence that explains the relationship. Use the surrounding prose to state its meaning, such as \`dispatches to\`, \`depends on\`, \`shares infrastructure with\`, \`is configured through\`, \`is surfaced by\`, or \`is secured by\`.
+- Do not add links solely to increase graph density, and do not automatically add reciprocal links. Add an inverse link only when it helps explain the target concept and is supported by evidence.
+- ${output.quickstartPath} must link to every major concept for navigation, but quickstart and index links do not count toward the semantic relationship audit.
+- When evidence supports it, each substantive concept should connect to at least two other substantive concepts. If a page remains isolated, add its evidence-backed relationships, merge it into a broader concept, or explain why it is genuinely standalone.
+- Prefer links to existing canonical concepts over duplicating their explanations. Do not mint thin concepts merely to create more nodes or edges.`;
+}
+
+/**
+ * OKF front-matter requirements. `repairPass` controls the trailing note:
+ * OpenWiki's runtime repairs front matter deterministically after every run,
+ * but a keyless surface with no post-run pass must get it right in place.
+ */
+export function okfFrontMatterRules(
+  output: MethodologyContext,
+  options: { repairPass: boolean },
+): string {
+  const head = `Front matter requirements (OKF):
+- Every non-reserved Markdown concept file you create or update under ${output.docsLocation}, including the temporary ${output.planPath} file, MUST begin with OKF-compliant YAML front matter.
+- The front matter MUST follow the Google Knowledge Catalog OKF v0.1 schema.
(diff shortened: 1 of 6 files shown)


ethan-parkk:
This looks like a good integration boundary for host-agent-native OpenWiki workflows.

  Would the maintainers be open to a narrowly scoped follow-up for OpenCode using the same approach?

  The proposed command would be:

`  openwiki integration opencode [path]
`
  It would scaffold:

  - .opencode/skills/openwiki-init/SKILL.md
  - .opencode/skills/openwiki-update/SKILL.md

  The generated skills would reuse the shared methodology functions introduced here, while adapting the invocation guidance, tool vocabulary, and
  repository-relative paths for OpenCode.

  The scope would remain intentionally small:

  - code mode only,
  - init and update skills,
  - inference delegated to the model already configured in OpenCode,
  - no additional OpenWiki provider or API key,
  - no plugin hooks,
  - no marketplace metadata,
  - no helper tooling,
  - and no OMOS-specific behavior.

  Although OpenCode can discover Claude-compatible skill locations, a dedicated target would avoid Claude-specific invocation and tool guidance and would
  use OpenCode’s native .opencode/skills/ location.

  If this direction is welcome, I’d be happy to submit it as a separate follow-up PR after this one.

colifran:
I've been thinking a lot about this and i agree with this being the only way to make this type of integration work inside of anthropic's terms of service. I think probably having a collection of integrations would be interesting for code based wikis. Will start by reviewing this one!

colifran:
@ethan-parkk i think we should use this PR to land on the format of these integrations - once this lands i'm more than happy to take additional integrations!

colifran:
Also sorry for the delay getting to this

colifran:
Superseded by: https://github.com/langchain-ai/openwiki/pull/685

PR #476, opened 2026-07-26
fix: prune old checkpoints so chat sqlite storage doesn't grow unbounded

### LINKED ISSUE #360: [Bug]: Open wiki filled the mac storage while generating the documentation
### Description

I was trying out a OpenWiki and I gave it an OpenAI key. Models are deployed on our azure tenant. It kept on working for multiple hours but did not produce anything. After a few hours my mac storage was full and I noticed that there a sqlite db inside .openwiki/ which was ~ 890 GB!

I had to delete this and also uninstall openwiki to recover.

### Steps to Reproduce

Install
initialize
Generate documentation through chat

### Expected Behavior

Generated docs in fairly quick time as the repo I was trying this on was very simple and straight forward.

### Actual Behavior

Mac Storage was full

### Environment

- OS: Mac
- Node.js version: 11.17.0
- OpenWiki version: 0.1.2

### Additional Context

_No response_

Zelys-DFKH:
Fixes #360.

In chat mode, OpenWiki keeps one thread_id for the whole session and persists checkpoints in a sqlite file (`~/.openwiki/openwiki.sqlite`). The checkpointer (`SqliteSaver` from `@langchain/langgraph-checkpoint-sqlite`) only inserts new rows; each graph step writes the full state, and nothing prunes the old ones. Since OpenWiki only ever resumes from the latest checkpoint, that history just accumulates forever. In a long chat session, the file grows without bound. The reporter hit roughly 890GB.

I've added a `pruneCheckpointHistory` step that runs after each chat turn (success or error) and keeps only the latest checkpoint per thread/namespace, discarding write rows we don't need anymore. Both deletes run in a single transaction, so a crash mid-prune can't leave the tables in an inconsistent state. This only touches the persistent (chat) checkpoint target; the in-memory init/update checkpointer stays untouched.

How I tested it:
- `pnpm run typecheck`, `pnpm run lint:check`, and `pnpm run format:check` all pass.
- Added `test/checkpoint-pruning.test.ts`. It builds a checkpointer with synthetic checkpoint/write rows across two threads and two namespaces, including a parent-checkpoint chain. Tests confirm that pruning keeps only the latest checkpoint per thread/namespace, leaves the other thread alone, and retains the write rows needed to look up the retained checkpoint's parent.
- `pnpm test`: 670 passing. The 3 failures (all in `test/env-behavior.test.ts`) reproduce identically on main without this change; I stashed my changes and reran to confirm. CI runs on ubuntu-latest only, so these Windows-specific failures won't show up there.
- Added `@types/better-sqlite3` as a devDependency. `checkpointer.db` was untyped because `better-sqlite3` ships no bundled types and `@types/better-sqlite3` was only a transitive devDependency. Once code touches `checkpointer.db` directly, ESLint's type-aware rules flag every call as unsafe without the types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^10.0.1",
+    "@types/better-sqlite3": "^7.6.12",
     "@types/jsdom": "^28.0.3",
     "@types/node": "^22.10.2",
     "@types/react": "^18.3.17",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -84,6 +84,9 @@ importers:
       '@eslint/js':
         specifier: ^10.0.1
         version: 10.0.1(eslint@10.6.0)
+      '@types/better-sqlite3':
+        specifier: ^7.6.12
+        version: 7.6.13
       '@types/jsdom':
         specifier: ^28.0.3
         version: 28.0.3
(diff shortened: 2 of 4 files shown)


colifran (approved):
Thanks @Zelys-DFKH!

PR #477, opened 2026-07-26
feat: add support for multilingual output in openwiki agent

### LINKED ISSUE #469: [Feature]: Multi-language support for generated wiki output
## Summary
Add multi-language support so that OpenWiki can generate wiki documentation in languages other than English.

## Motivation
Currently, all system prompts in `src/agent/prompt.ts` are hardcoded in English, and the generated `openwiki/` documentation is always English-only. For teams and developers whose primary language is not English, this creates a significant barrier to adoption and usefulness.

## Proposed Solution
1. Add a `--language` / `-l` CLI flag (e.g. `openwiki --init --language zh-CN`)
2. Pass the locale through `RunContext` to `createSystemPrompt()` and `createUserPrompt()`
3. Localize the system prompt instructions and documentation goals based on the selected locale
4. Instruct the LLM to generate all wiki content in the target language

Abhiiishek44:
## Summary

Adds first-class multilingual output to OpenWiki. A new --language flag lets a repo generate and maintain its wiki in a target language (for example zh-CN), while keeping everything machine-facing canonical so aggregation, retrieval, and tooling stay stable.

When no language is provided, existing behavior is unchanged.

### Language plumbing
- New --language <tag> flag threaded through commands.ts into the run context and agent prompt.
- src/language.ts normalizes and validates BCP-47 tags; the resolved language is persisted as wiki state in .last-update.json.

### Translation middleware
- A beforeAgent pass brings existing pages into the target language before the agent runs, so an incremental update never leaves a page half in the old language and half in the new one.
- Three cost tiers, cheapest first: a no-op sweep walks the tree with zero model calls; a plain update retranslates only pages a prior run left marked openwiki_translation_pending; a real language switch retranslates every page.
- Graceful degradation: a single page's failure never aborts the run. The page keeps its previous content, gains a pending marker carrying the target language so the next update retries it, and is reported once through a secret-redacted warning.
- Translation output is treated purely as file text and written through the sandboxed docs-only backend; paths come from backend enumeration, never model output.

### OKF front-matter localization
- Reader-facing values (title, description, type) and body prose are localized.
- tags stay in English as canonical cross-cutting aggregation keys, so tag-based grouping does not fragment across languages. Front-matter keys and all machine values (URLs, file paths, timestamps, identifiers) are copied byte-for-byte.
- Deterministic index generation localizes structural labels (Files/Directories) via src/okf/index-labels.ts, falling back to English by primary subtag.

### Clean TUI output
- Per-page translations are tagged langsmith:nostream, so the raw translated Markdown never floods the token stream.
- A single "Translating wiki docs..." status line is shown instead, emitted once and only when at least one page is actually translated.

### Prompt guidance
- The agent is told that whole-wiki language reconciliation is code-owned: it must not re-translate existing pages or act on stale language metadata mid-run, mirroring the existing openwiki_generated and openwiki_translation_pending guidance.

## Example:

```
openwiki --init --language zh-CN
openwiki --update -l hi
openwiki --update          # inherits the wiki's persisted language
```

## Why

OpenWiki's value is wiki-as-agent-memory, and that only reaches non-English teams if the wiki reads in their language while staying canonical where machines consume it. Splitting "translate what's read, keep canonical what's matched" keeps retrieval, tag aggregation, and index generation stable across a language switch.

## Testing

Added coverage for:
- --language <locale> parsing, the -l alias, and missing language values
- propagation of the selected language through the run context
- language-specific prompt instructions and CLI help output
- translation-plan resolution (switch vs. plain update vs. marker sweep)
- translating every page on a switch, and sweeping only pending pages on a plain update
- partial-failure recovery: stamping and clearing the openwiki_translation_pending marker
- marker preservation across deterministic front-matter regeneration
- suppression of the token stream behind the status line
- unchanged behavior when no language is provided

## Authorship

Co-authored by: @Abhiiishek44 and @colifran

---

Closes #469


colifran:
Good start on the CLI wiring. I've pushed some changes on top since the prompt-only approach left a few gaps:

1. **Existing pages weren't translated**. Instructing the model only affects what it writes this run but on an update it rewrites just the changed pages, so a language switch left the wiki mixed. Added a pass that translates the pages the agent doesn't touch.
2. **Language wasn't persisted**. The flag had to be re-passed every run. Now an update inherits the wiki's current language and only re-translates on an explicit switch.
3. **Deterministic content stayed English**. The directory indexes and code-repaired front matter are generated in code, not by the model, so they didn't follow the prompt. Localized those (index headings and the fallback concept type).
4. Front-matter type needed coordination with the OKF index generator, not just a prompt instruction, or indexes drift.

Left the CLI flag and prompt scoping mostly as-is.

colifran (approved):

PR #481, opened 2026-07-26
fix: ignore stray oauth callback requests

HwangJohn:
﻿## Summary
- keep the OAuth callback server waiting when browsers request non-callback paths like `/favicon.ico`
- only reject the pending OAuth flow for malformed real `/callback` requests
- cover the redirect URI, stray request, missing callback data, provider error, and shutdown trailing request cases

## Why
Browsers and local tooling can send extra requests to the loopback OAuth server before the provider redirect completes. Those requests should not fail the login attempt before the real callback arrives.

No dedicated issue; found during an OAuth callback hardening sweep.

## Testing
- `corepack pnpm exec vitest run test/oauth-callback-server.test.ts` passed
- `corepack pnpm run format:check` passed
- `corepack pnpm run lint:check` passed
- `corepack pnpm run typecheck` passed
- `git diff --check` passed
- `corepack pnpm test` was attempted on Windows and still fails outside this change in existing `test/env-behavior.test.ts` HOME/permission assertions plus Mermaid/jsdom timeout tests


colifran (approved):
Looks good - thanks @HwangJohn!

PR #483, opened 2026-07-26
fix: reject invalid hackernews feed configs

HwangJohn:
﻿## Summary
- warn when configured or requested Hacker News feeds are invalid
- stop with `status: "error"` and no raw result when there are no valid feeds and no query work
- preserve the default feed behavior when no feeds are explicitly configured

## Why
Explicitly invalid Hacker News feed config could previously produce an empty raw file and report success. This makes the failed configuration visible and avoids recording a successful ingest when no Hacker News work ran.

No dedicated issue; found during connector config hardening.

## Testing
- `corepack pnpm exec vitest run test/hackernews.test.ts` passed
- `corepack pnpm run format:check` passed
- `corepack pnpm run lint:check` passed
- `corepack pnpm run typecheck` passed
- `git diff --check` passed
- `corepack pnpm test` was attempted on Windows and still fails outside this change in existing `test/env-behavior.test.ts` HOME/permission assertions plus Mermaid/jsdom timeout tests

### DIFF
--- a/src/connectors/sources/hackernews.ts
+++ b/src/connectors/sources/hackernews.ts
@@ -17,11 +17,11 @@ import type {
 
 type HackerNewsConfig = {
   enabled?: boolean;
-  feeds?: HackerNewsFeed[];
+  feeds?: unknown;
   maxItemsPerFeed?: number;
   maxResultsPerQuery?: number;
-  queries?: string[];
-  queryTags?: string[];
+  queries?: unknown;
+  queryTags?: unknown;
 };
 
 type HackerNewsFeed = "ask" | "best" | "job" | "new" | "show" | "top";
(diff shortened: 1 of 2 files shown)


src/connectors/sources/hackernews.ts
```diff
@@ -63,6 +63,20 @@ type AlgoliaResponse = {
...
+const VALID_FEEDS: readonly HackerNewsFeed[] = [
+  "ask",
+  "best",
+  "job",
+  "new",
+  "show",
+  "top",
+];
+
+type FeedSelection = {
```
colifran, line 75: nit: in the future these shapes should really be an interface over a type.

colifran (approved):
looks good - thanks @HwangJohn!

PR #487, opened 2026-07-27
feat: add native Z.AI provider

### LINKED ISSUE #486: [Feature]: Add native Z.AI provider
## Summary

Add Z.AI as a first-class, native `zai` inference provider to OpenWiki.

## Problem

The current upstream `main` has no Z.AI provider in its provider registry, onboarding, diagnostics, or model-creation path. `openai-compatible` is a separate generic gateway provider; it is not a Z.AI provider and must not be repurposed or aliased as one.

The former wrapper/preload approach also made a Z.AI configuration appear as OpenRouter and could route requests to OpenRouter instead of Z.AI. It kept provider-specific transport and operational behavior outside the upstream CLI.

## Proposed solution

- Add a new independent provider selected with `OPENWIKI_PROVIDER=zai`.
- Use `ZAI_API_KEY`, optional `ZAI_BASE_URL`, and `glm-5.2` as the Z.AI defaults.
- Add Z.AI-specific setup, diagnostics, model creation, request normalization, and 429 retry handling.
- Preserve the existing `openai-compatible`, OpenRouter, and all other provider contracts unchanged; Z.AI must not fall back to their credentials.
- Keep provider-specific transport inside OpenWiki, without a wrapper, preload hook, `node_modules` patch, or global `fetch` hook.
- Move wrapper lifecycle safeguards for per-repository locking and bundled-skill synchronization into OpenWiki.

## Alternatives considered

- Continue using a wrapper/preload hook: rejected because it changes global fetch behavior and obscures the active provider.
- Reuse or rename `openai-compatible`: rejected because it would change an existing generic-provider contract and cannot express Z.AI-specific credentials, defaults, diagnostics, or request handling.

## Additional context

The implementation includes unit coverage for provider resolution, transport normalization/retries, execution locks, and skill synchronization.

m16khb:
Closes #486

## Context

The current upstream `main` has no native Z.AI provider. This PR adds `zai` as a new, independent first-class provider.

It does **not** convert, rename, alias, or otherwise change the existing `openai-compatible` provider. `openai-compatible` remains the generic gateway integration; Z.AI has its own credentials, defaults, diagnostics, transport behavior, and model creation path.

## Changes

- Add `OPENWIKI_PROVIDER=zai`, `ZAI_API_KEY`, optional `ZAI_BASE_URL`, and the `glm-5.2` default model.
- Add Z.AI onboarding and secret-safe diagnostics that show only Z.AI-relevant settings.
- Use a provider-local Z.AI transport for file-block normalization and 429 retries.
- Keep Z.AI independent from OpenRouter and `openai-compatible`: no credential fallback or request routing through either provider.
- Move wrapper lifecycle safeguards for scoped execution locks and bundled-skill synchronization into OpenWiki.
- Document native Z.AI configuration in the README.

## Validation

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm lint:check`
- `pnpm build`
- Full Vitest coverage was rerun in sharded batches. Under local macOS CPU contention, five existing Gemini/Mermaid tests exceeded their 5s default timeout; each passed when rerun in isolation or with a 30s runner timeout.
- Credential-redacted Z.AI JSON/SSE smoke requests and isolated CLI `code --init` smoke run.

No credentials, generated dependencies, wrapper/preload logic, `node_modules` changes, or global fetch hooks are included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,26 @@ notes.
 
 ## Customizing
 
-OpenWiki supports OpenAI (with an API key or a ChatGPT login), OpenRouter, Gemini (AI Studio), Gemini Enterprise (Vertex AI), Nebius Token Factory, Fireworks, Baseten, NVIDIA NIM, an OpenAI-compatible provider, AWS Bedrock, and Anthropic out of the box. The onboarding default is OpenAI with `gpt-5.6-terra`, and each inference provider also includes pre-defined model options plus support for custom model IDs.
+OpenWiki supports OpenAI (with an API key or a ChatGPT login), OpenRouter, Z.AI, Gemini (AI Studio), Gemini Enterprise (Vertex AI), Nebius Token Factory, Fireworks, Baseten, NVIDIA NIM, an OpenAI-compatible provider, AWS Bedrock, and Anthropic out of the box. The onboarding default is OpenAI with `gpt-5.6-terra`, and each inference provider also includes pre-defined model options plus support for custom model IDs.
+
+### Z.AI
+
+The native `zai` provider calls Z.AI directly; it does not route through
+OpenRouter or the generic `openai-compatible` provider. Set `ZAI_API_KEY` and,
+optionally, a custom coding endpoint. The default model is `glm-5.2` and the
+default endpoint is `https://api.z.ai/api/coding/paas/v4`:
+
+```bash
+OPENWIKI_PROVIDER=zai
+ZAI_API_KEY=your-zai-api-key
+# Optional; overrides the default Z.AI coding endpoint.
+ZAI_BASE_URL=https://api.z.ai/api/coding/paas/v4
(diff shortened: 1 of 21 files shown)


colifran:
Thanks for putting this together, and for keeping it cleanly separated from the existing openai-compatible provider.

After reviewing, I'm going to close this in favor of the existing path. Z.AI exposes a standard OpenAI-compatible chat-completions endpoint, so GLM already works today with no code changes:

```
OPENWIKI_PROVIDER=openai-compatible
OPENAI_COMPATIBLE_API_KEY=<your z.ai key>
OPENAI_COMPATIBLE_BASE_URL=https://api.z.ai/api/paas/v4
OPENWIKI_MODEL_ID=glm-5.2
```

This is the same approach we take for other hosted OpenAI-compatible gateways rather than adding a dedicated provider per vendor.

Issue #490, opened 2026-07-27
[Feature]: Validate selected models against provider account entitlements


jyje:
### Summary

OpenWiki should optionally validate whether a selected model is available to the currently authenticated provider account.

This is separate from discovering or maintaining a global model catalog. The existing curated `modelOptions` lists can remain responsible for model names and recommendations, while an account-scoped entitlement check answers the narrower question: “Can this account actually use this model?”

### Motivation

PR #192 exposed a case where globally valid models were not consistently available across accounts.

During review, another contributor tested the GitHub Copilot provider and found `claude-opus-5` and `claude-fable-5` unavailable on their account. The same models were available on mine through the same API endpoint.

That difference can come from plan tiers, organization or GHE tenant settings, staged rollouts, feature flags, or other provider-side entitlements.

A static catalog cannot represent those account-specific differences. Without an earlier check, users can select a valid-looking model and only discover that it is unavailable when the first inference request fails.

### Proposed Solution

Introduce an optional, provider-specific entitlement check for providers that expose an account-scoped model availability endpoint.

For GitHub Copilot, the first implementation could query:

```text
GET {baseURL}/models
```

using the current authenticated session. The response already includes fields such as `capabilities.type` and `model_picker_enabled`, which can be used to determine whether a selected chat model is available to that account.

The existing curated model list should remain the source of labels and recommendations. Entitlement validation should only affect account-specific availability:

- Models confirmed unavailable can be marked as unavailable in the picker.
- A saved or manually entered model can be checked before starting a run.
- The resulting error should explain that the model exists but is not enabled for the current account, rather than surfacing a less actionable provider error during inference.
- Providers without an entitlement endpoint should keep the current behavior.
- If validation cannot run because the user is offline, credentials are unavailable, or the endpoint fails, OpenWiki should fall back gracefully instead of blocking setup.

The implementation should be provider-extensible, but it does not need to solve cross-provider model discovery or maintain a shared catalog.

### Alternatives Considered

- **Full live discovery.** Fetch and merge every provider’s live model catalog into the picker. This broader approach was discussed in #475, but it combines global catalog freshness with account-specific availability. Shared model discovery appears better suited to a common JavaScript model-profile layer rather than being independently implemented in OpenWiki.

- **Static model lists only.** Preserve the current implementation and accept that users may discover entitlement mismatches only after selecting a model and attempting inference.

[...]

PR #491, opened 2026-07-27
feat: validate openai model availability

jyje:
## Summary

- Validate the selected model against the direct OpenAI Models API before inference starts.
- Stop only when the selected model is authoritatively absent; preserve the existing inference path when the lookup fails or `OPENAI_BASE_URL` points to a custom endpoint.
- Add unit coverage for available, unavailable, lookup-failure, custom-endpoint, and unsupported-provider paths.

## Validation

- `pnpm test` — 58 files, 694 tests passed
- `pnpm typecheck`
- `pnpm lint:check`
- Live GitHub Actions verification: [OpenAI availability contract](https://github.com/jyje/openwiki/actions/runs/30277633642)
  - positive: `gpt-5.6-luna` is available
  - negative: a nonexistent model is unavailable
  - fallback: missing credentials and custom endpoints return `unknown` without blocking inference

The live workflow runs only on a fork-only verification branch. Its repository secret is not part of this PR, its workflow, or its logs.

## Related issue

This is the first provider-specific implementation for #490. Follow-up providers can reuse the same `available` / `unavailable` / `unknown` contract without making a catalogue lookup failure block inference.

### DIFF
--- a/src/model-availability.ts
+++ b/src/model-availability.ts
@@ -5,12 +5,12 @@ export type ModelAvailability =
   | { status: "unavailable"; reason: string }
   | { status: "unknown"; reason?: string };
 
-type ModelAvailabilityCheck = {
+interface ModelAvailabilityCheck {
   apiKey?: string;
   baseUrl?: string;
   modelId: string;
   provider: OpenWikiProvider;
-};
+}
 
 type OpenAIModelListResponse = {
   data?: Array<{ id?: unknown }>;
--- a/src/model-availability.ts
+++ b/src/model-availability.ts
@@ -1,0 +1,85 @@
+import type { OpenWikiProvider } from "./config/constants.js";
+
+export type ModelAvailability =
+  | { status: "available" }
+  | { status: "unavailable"; reason: string }
+  | { status: "unknown"; reason?: string };
+
+type ModelAvailabilityCheck = {
+  apiKey?: string;
+  baseUrl?: string;
+  modelId: string;
+  provider: OpenWikiProvider;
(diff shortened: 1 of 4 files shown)


src/model-availability.ts
```diff
@@ -0,0 +1,85 @@
+import type { OpenWikiProvider } from "./constants.js";
+
+export type ModelAvailability =
+  | { status: "available" }
+  | { status: "unavailable"; reason: string }
+  | { status: "unknown"; reason?: string };
+
+type ModelAvailabilityCheck = {
```
colifran: nit: can we make this an interface?

jyje: Updated in 5d55e28 - `ModelAvailabilityCheck` is now an interface. Thanks!

jyje:
I also captured two manual checks for the updated model selection behavior:

- True-positive example for `gpt-5.6-luna`

  OpenWiki accepts gpt-5.6-luna as an available OpenAI model

- True-negative example for `bad model`

  OpenWiki rejects bad model as an invalid OpenAI model ID

This is only covering the official OpenAI provider path. Custom OpenAI-compatible endpoints can be considered separately if needed.

jyje:
Hello @colifran, quick bump. Interface nit's fixed and verification screenshots are up.
Whenever you have a chance :D

colifran:
hey @jyje sorry for the delay. will prioritize reviewing this!

colifran (approved):
Looks good - thanks @jyje!

PR #502, opened 2026-07-28
fix(cli): make bundled skill sync idempotent when directories already exist

### LINKED ISSUE #499: EEXIST error on --init when skills directories already exist
## Bug

Running `openwiki --init` fails with:

```
Error EEXIST: file already exists, mkdir '/Users/.../.openwiki/skills/mermaid-diagrams'
```

## Steps to reproduce

1. Run `openwiki --init` in a repo
2. The command partially completes, creating `~/.openwiki/skills/mermaid-diagrams` and `~/.openwiki/skills/write-connector`
3. Run `openwiki --init` again (or if the first run itself triggers the sync twice)
4. Error thrown — `cp` in `replaceSkillDirectories` (skills.js) calls `mkdir` on a target that already exists

## Root cause

In `dist/agent/skills.js`, `replaceSkillDirectories` does `rm(target)` then `cp(src, target, { recursive: true })`. The `cp` call internally uses `mkdir` without gracefully handling the case where the destination directory already exists, throwing `EEXIST`.

## Expected behavior

`--init` should be idempotent — re-running it should not error if the skills directories are already present.

## Fix suggestion

Pass `{ force: true }` (or equivalent) to the `cp` call in `replaceSkillDirectories`, or wrap the mkdir inside `cp` to use `{ recursive: true }` so it doesn't throw on existing directories.

## Environment

- openwiki version: 0.2.3
- Node.js version: v26.5.0
- OS: macOS Darwin 25.5.0

Mohith26:
Fixes #499

## What

`openwiki --init` could crash with `EEXIST: file already exists, mkdir '~/.openwiki/skills/<skill>'`. Bundled-skill sync did `rm(target)` then `cp(source, target)` per skill, and `fs.cp` stats the destination then calls a non-recursive `mkdir`, so overlapping syncs (init triggers the sync more than once) raced into EEXIST. Reproduced on the first concurrent pair, 5/5 trials.

Skill installation is now idempotent and atomic: each skill is staged via `cp` into a unique `mkdtemp` scratch dir, the old copy is displaced with `rename` (in-place `rm` races into ENOTEMPTY mid-delete), and the fresh copy is swapped in with `rename`, accepting an already-installed directory from a concurrent sync. Uses the `mkdir(..., { recursive: true })` exists-ok pattern already present in `openwiki-home.ts`.

## Tests

Regression test pre-creates existing skill dirs, re-runs sync sequentially, then runs 25 concurrent-pair syncs and asserts bundled content wins with no staging leftovers. Fails on main with the exact reported EEXIST error; passes with the fix. Full suite: 769 tests green. prettier, eslint, tsc clean. Also stress-tested 500 iterations of 3-way concurrent sync (1,500 calls, 0 failures).

Credit to @marcuss for the report in #499.


amarakramali (approved):
Reviewed commit `5f26c7310f546a98f71c78febf92bf154c3d7581`, focusing on the overlapping-sync path and native Windows directory-rename behavior.

The same-filesystem `mkdtemp` staging plus `rename` swap is sound for this use case: the target parent is created first, each caller gets an isolated scratch directory, and the target-exists check correctly lets a losing concurrent installer accept the complete copy installed by the winner. The regression test also covers stale-directory replacement and cleanup of staging artifacts.

I validated this on native Windows/NTFS with Node.js `v24.18.0`:

- `test/skills.test.ts`: passed in 10 independent runs; each run includes 25 overlapping sync pairs (250 pairs / 500 concurrent calls total), plus the sequential rerun case.
- `pnpm run typecheck`: passed.
- `pnpm run lint:check`: passed.
- `git diff origin/main...HEAD --check`: passed.
- Full Vitest run: 766/769 passed in parallel; the remaining three were unrelated Mermaid tests hitting their 5 s timeout. Re-running those tests in isolation passed all three, including the initially slow `mermaid-wiki` case.

I found no blocking issue with the fix.

colifran (approved):
This has been a very annoying bug. Thanks for this!

Issue #489, opened 2026-07-27
openwiki code --init fails with EEXIST on second run (mkdir ~/.openwiki/skills/mermaid-diagrams)


amdeldacc:
## Bug Description

Running `openwiki code --init` a second time (after having already run it once before) crashes with:

```
Error EEXIST: file already exists, mkdir '~/.openwiki/skills/mermaid-diagrams'
```

The `--init` command is not idempotent — it fails rather than skipping already-initialized directories.

## Steps to Reproduce

1. Run `openwiki code --init` (first time — works fine)
2. Exit the session
3. Run `openwiki code --init` again

## Expected Behavior

Running `--init` a second time should either:
- Be a no-op if already initialized, or
- Skip existing directories and continue

## Environment

- openwiki v0.2.3 (npm global install)
- Linux (WSL2)

## Full Output

```
$ openwiki code --init
  ___                  __        ___ _    _
 / _ \ _ __   ___ _ __ \ \      / (_) | _(_)
| | | | '_ \ / _ \ '_ \ \ \ /\ / /| | |/ / |
| |_| | |_) |  __/ | | | \ V  V / | |   <| |
 \___/| .__/ \___|_| |_|  \_/\_/  |_|_|\_\_\ |_|

╭────────────────────────────────────────────────────────────────────────────╮
│ >_ OpenWiki v0.2.3 agent docs for codebases                               │
│ provider: NVIDIA NIM                                                       │
│ model: nvidia/nemotron-3-ultra-550b-a55b                                   │
│ directory: ~/                                                   │
╰────────────────────────────────────────────────────────────────────────────╯

* LangSmith tracing enabled - Run failed
Tip: ask for a docs change, or use /exit when you are done.

* Error EEXIST: file already exists, mkdir '~/.openwiki/skills/mermaid-diagrams'
```

The issue is likely in the `--init` code path that creates `~/.openwiki/skills/mermaid-diagrams` without first checking if it already exists.

asgarov:
Hit this too, but with a repro that rules out the "second run" theory — it reproduces on a **completely fresh install with no `~/.openwiki` at all**, which suggests the seeding runs twice *within a single `--init`* rather than being non-idempotent across runs.

## Repro

```sh
mv ~/.openwiki ~/.openwiki.bak   # ensure no prior state whatsoever
openwiki code --init
```

```
* Error EEXIST: file already exists, mkdir '~/.openwiki/skills/write-connector'
```

[...]

PR #503, opened 2026-07-28
fix(cli): handle unhandled promise rejections with a clean error and non-zero exit

### LINKED ISSUE #494: --init aborts with a raw unhandled rejection on provider 429 (no top-level rejection handler)
## Summary

When the model provider returns `429`, `openwiki code --init` dies as an **unhandled promise rejection** — raw Node stack trace, process abort, no diagnostic. The exact same error on `openwiki code --update` is handled cleanly. The graceful handler already exists; the `--init` path just doesn't route through it.

## Actual behavior

`openwiki code --init`:

```
file:///usr/local/lib/node_modules/openwiki/node_modules/langchain/dist/agents/transformers/tool-call.js:152
                        pending.rejectOutput(new Error(message));
                                             ^

Error: 429 The usage limit has been reached

Troubleshooting URL: https://docs.langchain.com/oss/javascript/langchain/errors/MODEL_RATE_LIMIT/

    at Object.process (file:///usr/local/lib/node_modules/openwiki/node_modules/langchain/dist/agents/transformers/tool-call.js:152:29)
    at StreamMux.push (file:///usr/local/lib/node_modules/openwiki/node_modules/@langchain/langgraph/dist/stream/mux.js:173:66)
    at pump (file:///usr/local/lib/node_modules/openwiki/node_modules/@langchain/langgraph/dist/stream/mux.js:321:36)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)

Node.js v22.23.1
```

`openwiki code --update`, same account, same exhausted quota, minutes later:

```
429 The usage limit has been reached

Troubleshooting URL: https://docs.langchain.com/oss/javascript/langchain/errors/MODEL_RATE_LIMIT/
```

## Expected behavior

`--init` should surface the same clean message as `--update` and exit non-zero, rather than aborting on an unhandled rejection.

## Steps to reproduce

1. Configure any provider whose quota can be exhausted (reproduced with OpenAI ChatGPT OAuth on a Business plan).
2. Run `openwiki code --init` against a large repository — large enough that generation outlives the remaining quota.
3. When the quota is hit mid-run, the process aborts with the stack trace above.
4. Run `openwiki code --update` while still rate-limited to see the graceful path.

## Impact

Beyond the poor error message, the aborted `--init` leaves the wiki in a state that misrepresents itself:

- `openwiki/quickstart.md` was written, and all seven section pages it links to were not — so every link on the entry page is dead.
- Section directories (`architecture/`, `domain/`, `integrations/`, `operations/`, `testing/`, `workflows/`) exist but are empty.
- **`openwiki/.last-update.json` was still stamped as a successful run:**

```json
{
  "updatedAt": "2026-07-27T15:41:07.480Z",
  "command": "init",
  "gitHead": "<sha>",
  "model": "gpt-5.6-sol"
}
```

That last point is the one with teeth. The scheduled GitHub Actions workflow that `--init` scaffolds runs `openwiki code --update` daily; a partially-generated wiki carrying a success stamp can be treated as current. Writing `.last-update.json` only after all pages are generated — or recording partial state explicitly — would make the failure recoverable.

Also worth noting: `--modelId gpt-5.4-mini` returns the same 429, so on ChatGPT OAuth the limit is account-wide rather than per-model. Dropping to a cheaper model is not a workaround.

## Environment

- openwiki 0.2.3 (npm global)
- Node.js v22.23.1
- macOS 26 (Apple Silicon)
- Provider: OpenAI (ChatGPT login), model `gpt-5.6-sol`
- Target: ~11k-file Rails monorepo

Mohith26:
Fixes #494

## What

The CLI installs no `process.on("unhandledRejection")` handler, so a provider 429 rejecting in a promise nothing awaits (e.g. the orphaned second agent run from the #489 double-start) aborts with a raw Node stack trace instead of the clean error awaited failures get.

This adds a top-level handler installed at CLI entry before any async work: the rejection reason is routed through the existing `getErrorMessage()` path (secret redaction, provider special cases) to stderr, and the process exits 1. Command-agnostic per the root-cause follow-up in the issue: it covers `--init`, `--update`, and dependency-floated promises alike. The double-start itself stays tracked in #489.

## Tests

5 tests: the 429 shape from the issue produces the clean two-line message with no stack frames and exit 1; secrets are redacted; non-Error reasons handled; handler registration and dispatch of an emitted escaped rejection. Full suite: 773 passed. prettier, eslint, tsc clean.

Credit to @asgarov for the report and root-cause follow-up in #494.

### DIFF
--- a/src/cli.tsx
+++ b/src/cli.tsx
@@ -41,6 +41,7 @@ import {
   isSecretLikeKey,
   sanitizeDiagnosticText,
 } from "./diagnostics.js";
+import { installUnhandledRejectionHandler } from "./unhandled-rejection.js";
 import { stripHtmlTags } from "./utils.js";
 import {
   type OpenWikiRunEvent,
@@ -91,6 +92,12 @@ import {
   FIRST_RUN_NOTICE_OPT_OUT,
   FIRST_RUN_NOTICE_VERIFY,
 } from "./telemetry/index.js";
+
+// Installed before any async work starts: a rejection that escapes every
+// awaited chain (e.g. a provider 429 inside a run nothing is awaiting) must
+// print a clean message and exit non-zero instead of aborting with a raw
+// Node stack trace.
+installUnhandledRejectionHandler();
 
 type RunState =
   | { status: "idle" }
--- a/src/unhandled-rejection.ts
+++ b/src/unhandled-rejection.ts
@@ -1,0 +1,30 @@
+import { getErrorMessage } from "./diagnostics.js";
+
+/**
+ * Last-resort handler for promise rejections that escape every awaited chain —
+ * for example a provider 429 raised inside an agent run that nothing is
+ * awaiting, or a promise floated inside a dependency. Without a listener Node
+ * aborts the process with a raw stack trace; this routes the failure through
+ * the same user-facing path as awaited errors (`getErrorMessage`, which
+ * redacts secrets and keeps the message actionable) and exits non-zero.
+ *
+ * Exported separately from the installer so tests can exercise the exact
+ * function the process invokes.
+ */
+export function handleUnhandledRejection(reason: unknown): void {
+  process.stderr.write(`${getErrorMessage(reason)}\n`);
+
+  // An escaped rejection leaves the process in an unknown state, so keep
+  // Node's crash semantics (its default is to abort) — just fail cleanly.
+  // Setting `process.exitCode` is not enough here: unlike the awaited error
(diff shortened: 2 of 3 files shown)


colifran:
Thanks for addressing this. The CLI and error-handling architecture were later reorganized, and the current entrypoint now installs a crash guard before every run, with hardened error classification and run accounting from #604. Since every path changed by this PR has been removed or replaced, I’m closing this implementation as obsolete.

PR #505, opened 2026-07-28
docs: 添加 OpenWiki 文档中文翻译（20 个文件）

topbat:


colifran:
Thanks for the PR but this isn't something we want to support in the repo. It sets a precedent of having per language support. OpenWiki generates multilingual docs via the `--language` flag on init or update.

PR #516, opened 2026-07-29
fix(agent): make the diagnostics fetch patch reentrant (#411)

### LINKED ISSUE #411: Global fetch monkey-patch for OpenRouter diagnostics is not concurrency-safe

Mr-Neutr0n:
Fixes #411.

`installOpenRouterDebugFetch` captured `globalThis.fetch` per call and restored it blindly in a `finally`, so two overlapping runs in one process corrupt each other exactly as the issue describes:

1. run A installs, capturing the real `fetch`
2. run B installs, capturing **A's wrapper** as its "original"
3. A restores the real `fetch`
4. B restores A's wrapper — leaked into the global permanently

One wrapper is now installed on the first entry and removed on the last exit, so nesting composes in any order. Each run keeps its own failure slot.

Three details beyond the plain refcount:

- **`restore()` is idempotent.** A double call would otherwise drop the depth below the number of live installs and un-patch a run that is still going.
- **The final restore only un-patches if `globalThis.fetch` is still our wrapper.** If a third party patched over us, clobbering it would be this same bug wearing different clothes.
- **Out-of-order restore is handled**, not just LIFO.

## Scope — this is the alternative, not the preferred fix

The issue prefers migrating to the provider client's `configuration.fetch` hook and says *"if a global swap must be retained short-term, guard it with a reentrancy counter."* I did the guard, deliberately.

The migration touches ~10 constructions across five client classes — `ChatOpenAI`, `ChatOpenRouter`, `ChatAnthropic`, `ChatGoogle`, `ChatBedrockConverse` — with three different fetch-injection shapes, and only some accept `configuration.fetch` at all. That is its own PR with its own review. This one closes the concurrency hazard the issue reports, today.

## Known limitation, commented in the code

A process-global `fetch` cannot be attributed to the run that issued it, so every concurrently-open capture records a failure. Runs read their slot from the `catch` that just failed, so in practice the reader is the owner — and a second run seeing a stray failure is strictly better than the previous behaviour of leaking the wrapper. Precise attribution is exactly what the `configuration.fetch` migration buys.

## Tests

`test/openrouter-debug-fetch-reentrancy.test.ts` — 7 cases. **Three fail on current `main`**, including the exact interleaving above:

```
× overlapping runs do not leak the wrapper (the issue's interleaving)
× a double restore cannot un-patch a run that is still live
× a foreign patch installed over ours is not clobbered on restore
```

Full suite **776 passed**, typecheck and `lint:check` clean.

---

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1593,28 +1593,38 @@ const OPENROUTER_DEBUG_BODY_LIMIT = 4_000;
 let debugFetchDepth = 0;
 let debugFetchReal: typeof globalThis.fetch | null = null;
 let debugFetchWrapper: typeof globalThis.fetch | null = null;
-const activeDebugCaptures = new Set<{
+type DebugCaptureSlot = {
   lastFailure: OpenRouterFetchFailure | null;
-}>();
+  options: OpenWikiRunOptions;
+};
 
-function recordDebugFailure(failure: OpenRouterFetchFailure): void {
+const activeDebugCaptures = new Set<DebugCaptureSlot>();
+
+function recordDebugFailure(
+  failure: OpenRouterFetchFailure,
+  note?: string,
+): void {
   // A process-global fetch cannot be attributed to the run that issued it, so
   // every concurrently-open capture records it. Runs read their slot from the
   // catch block that just failed, so in practice the reader is the owner; a
   // second concurrent run seeing a stray failure is strictly better than the
   // previous behaviour of leaking the wrapper. Per-provider
   // `configuration.fetch` injection would attribute precisely — see the issue.
+  //
+  // Debug output goes to each slot's OWN options: one shared wrapper must not
+  // log a second run's failures at the first run's debug level.
   for (const capture of activeDebugCaptures) {
     capture.lastFailure = failure;
+    if (note) {
+      emitDebug(capture.options, note);
+    }
   }
 }
 
 export function installOpenRouterDebugFetch(
   options: OpenWikiRunOptions,
 ): OpenRouterFetchCapture {
-  const slot: { lastFailure: OpenRouterFetchFailure | null } = {
-    lastFailure: null,
-  };
+  const slot: DebugCaptureSlot = { lastFailure: null, options };
   activeDebugCaptures.add(slot);
 
   if (debugFetchDepth === 0) {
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Self-review catch, pushed as `cc4835c`.

Collapsing to one shared wrapper introduced a bug my first version didn't have: the wrapper closed over the **first** installer's `options`, so with overlapping runs a second run's OpenRouter HTTP failures were logged at the first run's debug level — silently dropped when run A had `debug: false` and run B had it on.

The original code got this right by accident, because each run had its own wrapper. The refcount is what created the hazard, so it's on me.

Options now live on each capture slot and `recordDebugFailure` emits per slot. New test: A installs with debug off, B with debug on, and B must receive the `openrouter.http` line while A stays quiet.

8 reentrancy cases now, suite at 777 passed, lint and typecheck clean.

colifran:
Closing in favor of #416

PR #414, opened 2026-07-20
fix: correct git-repo incremental diff and isolate connector ingestion failures

bikeusaland:
## Summary

Fixes two correctness bugs in the connector ingestion path.

### 1. git-repo connector diffed the working tree, not changes since the last run (#409)

`GitRepoManifest` declared `previousHead` and connector state already persisted each repo's head in `latestIds`, but `createRepoManifest` never read the prior head back. `changedFiles` was always `git diff --name-status HEAD` (uncommitted working-tree changes only), so the manifest never described commits landed since the last ingestion — contradicting the recent-changes window the agent is told about.

Now the recorded head is passed in and, when it is still a reachable commit and differs from the current head, the manifest is built from `git diff <previousHead> HEAD` and `git log <previousHead>..HEAD`. It falls back to the working-tree diff on the first run, when nothing new was committed, or when the recorded head is unreachable (force-push / rewritten history) — a `rev-parse --verify` guard ensures a diff against a missing commit can never throw.

### 2. One connector/stream failure aborted the whole ingestion run (#412)

`ingestAllConnectors` ran connectors serially with no error handling, so the first connector that threw (e.g. an un-refreshable token) discarded results from every connector that already succeeded. It now runs them concurrently with `Promise.allSettled`, turning each rejection into a well-formed `error` result.

The X connector had the same shape internally: any single stream error (e.g. a 429 on one endpoint) threw out of `ingest`, orphaning already-written dumps and skipping the state write so `latestIds` never advanced — forcing a full re-fetch next run. Each stream (and each list within `list_posts`) is now wrapped in try/catch, records a warning, and always writes state at the end, mirroring the resilient pattern already used by the hackernews and git-repo connectors. A run that fetched nothing but hit an error now reports `error` instead of a benign `skipped`.

## Tests

Three new test files (7 new tests):

- `test/git-repo-connector.test.ts` — drives real temp repos across two runs: first run has no `previousHead`, second run diffs since the recorded head, and an unreachable recorded head falls back safely.
- `test/x-connector-stream-isolation.test.ts` — one failing stream keeps the succeeding dump and records a warning; all-failing yields `error`.
- `test/ingest-all-connectors.test.ts` — a throwing connector does not discard a succeeding connector's result.

Full suite green: typecheck, `eslint .`, `prettier --check`, and 420 tests pass.

Closes #409
Closes #412

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #517, opened 2026-07-29
fix(git-repo): report changes since the last run, not the dirty tree (#409)

### LINKED ISSUE #409: git-repo connector reports working-tree diff instead of changes since last run
### Summary

The `git-repo` connector reports working-tree dirty files as "changed files" instead of what changed since the last ingestion run, so the ingestion window semantics the agent is told about don't match reality.

### Details

`GitRepoManifest` declares an optional `previousHead`, and connector state already persists the current head:

```ts
// src/connectors/sources/git-repo.ts:111-116
nextState.latestIds = {
  ...(nextState.latestIds ?? {}),
  ...Object.fromEntries(
    manifests.map((manifest) => [manifest.id, manifest.head]),
  ),
};
```

But `createRepoManifest(id, repoPath)` never reads the prior head back out of state, so `previousHead` is always `undefined` and `changedFiles` is always computed against the working tree:

```ts
// src/connectors/sources/git-repo.ts:147-151
const changedFiles = (
  await runGit(repoPath, ["diff", "--name-status", "HEAD"])
)
  .split(/\r?\n/u)
  .filter(Boolean);
```

`git diff --name-status HEAD` only reports **uncommitted** changes in the working tree — not commits landed since the last run. The ingestion orchestrator (`src/ingestion.ts`) presents these to the agent as the recent-changes window, so the manifest is misleading.

### Suggested fix

Pass the prior `state.latestIds[id]` into `createRepoManifest`, set `previousHead`, and when present compute:

- `git diff --name-status <previousHead> HEAD` for changed files
- `git log <previousHead>..HEAD` for commits since the last run

Fall back to the current behavior only on the first run (no prior head recorded) or when the recorded head is unreachable (e.g. history rewritten/force-pushed).

### Location

`src/connectors/sources/git-repo.ts:130-162`

Mr-Neutr0n:
Fixes #409.

`createRepoManifest` never read the prior head back out of connector state, so `previousHead` was always `undefined` and `changedFiles` was always:

```ts
git diff --name-status HEAD
```

which reports only **uncommitted** working-tree changes. The connector already persists each repo's head in `state.latestIds`, and `src/ingestion.ts` presents `changedFiles` to the agent as the recent-changes window — so the agent was shown the developer's dirty tree and told it was the commits landed since last ingestion.

Threads `state.latestIds[repo.id]` into `createRepoManifest` and diffs `previousHead..HEAD`.

## The three cases a naive version gets wrong

Each has a test:

| case | behaviour |
|---|---|
| no recorded head | first run has no baseline — fall back to the working tree, the only sensible answer |
| head unchanged | nothing landed since the last run, so the window is **empty**. A dirty tree is not a change since last run, and it is already reported separately via `status` |
| head unresolvable | rebase / force push / gc removed the recorded commit. Degrade to the working tree rather than aborting the repo, and **omit `previousHead`** so the manifest does not claim a window it could not compute |

That last one matters: the connector catches per-repo and pushes a warning, so an unresolvable head would otherwise turn one stale state entry into a permanently skipped repo.

## Tests

`test/git-repo-diff-window.test.ts` — 5 cases against a **real temporary git repo**, not a stubbed `runGit`. The bug was a wrong git invocation, and a stub would have happily agreed with the wrong one. Covers commits-since-head with an unrelated dirty file present, unchanged head, first run, unresolvable head, and a rename.

Full suite **774 passed**, typecheck and `lint:check` clean.

---

### DIFF
--- a/src/connectors/sources/git-repo.ts
+++ b/src/connectors/sources/git-repo.ts
@@ -89,7 +89,11 @@ async function ingest(
 
     const repoPath = path.resolve(repo.path);
     try {
-      const manifest = await createRepoManifest(repo.id, repoPath);
+      const manifest = await createRepoManifest(
+        repo.id,
+        repoPath,
+        state.latestIds?.[repo.id],
+      );
       manifests.push(manifest);
     } catch (error) {
       warnings.push(`${repo.id}: ${getErrorMessage(error)}`);
@@ -131,6 +135,7 @@ async function ingest(
 async function createRepoManifest(
   id: string,
   repoPath: string,
+  previousHead?: string,
 ): Promise<GitRepoManifest> {
   const branch = await runGit(repoPath, ["rev-parse", "--abbrev-ref", "HEAD"]);
   const head = await runGit(repoPath, ["rev-parse", "HEAD"]);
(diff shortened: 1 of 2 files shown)


colifran:
Closing in favor of #414

PR #518, opened 2026-07-29
fix(connectors): isolate per-connector and per-stream ingest failures (#412)

### LINKED ISSUE #412: ingestAllConnectors aborts all connectors on first failure; X connector loses partial state
### Summary

`ingestAllConnectors` runs connectors sequentially with no error handling, so the first connector that throws (e.g. an un-refreshable token) aborts the entire "all" run and discards results from every connector that already succeeded.

### Details

```ts
// src/connectors/tools.ts:270-281
async function ingestAllConnectors() {
  const registry = createConnectorRegistry();
  const results = [];

  for (const connector of Object.values(registry)) {
    results.push(await connector.ingest());
  }

  return {
    results,
  };
}
```

A single `throw` from any `connector.ingest()` propagates out of the loop, so `results` is never returned even for the connectors that completed successfully.

### Related: same shape inside the X connector

`src/connectors/sources/x.ts:114-160` iterates streams/lists with no per-iteration try/catch (unlike `hackernews.ts:128-151` and `git-repo.ts:90-95`, which catch per item and push warnings). Any single stream error (e.g. a 429 on `mentions`) throws out of `ingest`, so:

- `rawFiles` already written to disk are orphaned, and
- `writeConnectorState` (`x.ts:175`) never runs, so `latestIds` isn't advanced — the next run re-fetches everything.

### Suggested fix

- `ingestAllConnectors`: use `Promise.allSettled` (or per-connector try/catch), collect per-connector success/error into the results array, and skip unconfigured connectors early (mirror `getConfiguredConnectorIds` in `registry.ts:49`).
- X connector: wrap each stream/list iteration in try/catch, push failures into `warnings`, and always write state at the end — mirroring the resilient pattern already used by the Hacker News and git-repo connectors.

### Location

`src/connectors/tools.ts:270-281` (primary); `src/connectors/sources/x.ts:114-160` (related)

Mr-Neutr0n:
Fixes #412. Both halves.

## `ingestAllConnectors`

```ts
for (const connector of Object.values(registry)) {
  results.push(await connector.ingest());
}
```

A single throw — an un-refreshable token, a 429 — propagated out of the loop, so `results` was never returned and every connector that had already completed was discarded with it.

`ConnectorIngestResult` already models failure via `status: "error"`, so a failed connector now reports itself and the run continues. The registry is an optional parameter so the isolation contract is testable without real connectors.

## The X connector, one level down

`src/connectors/sources/x.ts` had the same shape inside `ingest`: the streams/lists loop had no per-iteration try/catch, unlike `hackernews.ts` and `git-repo.ts` which catch per item and push a warning.

Any single stream error threw out of `ingest()`, which meant:

- the `rawFiles` already written were orphaned on disk, and
- `writeConnectorState` never ran, so `latestIds` never advanced — the next run re-fetched everything it had just fetched

Each stream and each list is now caught individually and reported as a warning, matching the existing per-item pattern, so partial progress is kept and the cursor advances for the streams that did work.

## Tests

`test/ingest-error-isolation.test.ts` — 5 cases: a failure in the middle (successful connectors keep their `rawFiles`), a failure in the **first** position, the cause being carried rather than just a status, every connector failing, and a non-`Error` rejection.

Full suite **774 passed**, typecheck and `lint:check` clean.

---

### DIFF
--- a/src/connectors/sources/x.ts
+++ b/src/connectors/sources/x.ts
@@ -115,52 +115,65 @@ async function ingest(
   const latestIds = { ...(state.latestIds ?? {}) };
   const startTime = getWindowStartTime(options.windowHours);
 
+  // Per-stream isolation (#412). A single stream failure - a 429 on mentions, a
+  // deleted list - used to throw out of ingest(), which orphaned the rawFiles
+  // already written and skipped writeConnectorState entirely, so latestIds never
+  // advanced and the next run re-fetched everything. Mirrors the per-item
+  // try/catch in hackernews.ts and git-repo.ts.
   for (const stream of streams) {
     if (stream === "list_posts") {
       for (const listId of listIds) {
         const key = `list_posts:${listId}`;
-        const pages = await fetchPaginatedX(
-          accessToken,
-          `/lists/${encodeURIComponent(listId)}/tweets`,
-          {
-            since_id: latestIds[key],
-            start_time: startTime,
-          },
-          config.maxPagesPerStream,
-        );
-        latestIds[key] = getNewestId(pages) ?? latestIds[key] ?? "";
-        rawFiles.push(
-          await writeRawJson("x", runId, `list-${listId}.json`, {
-            fetchedAt: new Date().toISOString(),
-            listId,
-            pages,
-            stream,
-            windowHours: normalizeWindowHours(options.windowHours),
-          }),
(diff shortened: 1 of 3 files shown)


colifran:
Closing in favor of #414

PR #519, opened 2026-07-29
fix(connectors): return tool errors as results so the model can self-correct (#427)

### LINKED ISSUE #427: Connector tools throw instead of returning errors as tool results, so the model can't self-correct (e.g. wrong MCP tool name)

Mr-Neutr0n:
Fixes #427.

The connector tools threw on error instead of returning the message as the tool result. A thrown tool error aborts the agent run, and even when the run survives it, the model never sees the text — which matters because these messages are written **for the model**. `callMcpTool` answers a wrong tool name with:

> MCP tool notion-get-page-content was not returned by tools/list for notion. Run openwiki_list_mcp_tools first and use an exact discovered name.

That is precisely the hint needed to retry correctly, and the model never received it.

## Verified before/after on that exact path

```
before:  >>> REJECTED (model never sees the hint)
after:   >>> RETURNED: Tool error: notion MCP connector is not enabled.
```

Errors are surfaced, not swallowed — the result is prefixed `Tool error:` so a failure is never mistaken for data.

## ⚠️ Contract change that needs your eye

Three tests in `test/raw-connector-tools.test.ts` asserted that symlink rejections **throw**:

```ts
).rejects.toThrow(/symbolic links/u);
```

They now assert the same refusal on the result channel instead. **The security property is unchanged** — the symlink is still refused before any read, and the reason still reaches the caller — but this is a real change to what those tests pin, and they are security tests, so I did not want to quietly rewrite them and move on.

If you would rather security refusals keep hard-failing while ordinary tool errors are returned, say so and I will exempt the raw-file tools from the wrapper. That is a two-line change.

## Scope

LangChain rejects schema violations *before* `func` runs, so those remain rejections — a different layer. This covers errors thrown inside the tool body, which is what the issue is about.

## Tests

`test/connector-tool-errors.test.ts` — 6 cases: thrown `Error`, rejected promise, non-`Error` rejection, success passthrough untouched, the issue's MCP tool-name case end to end, and a successful connector call still returning parseable JSON.

Full suite **775 passed**, typecheck and `lint:check` clean.

---

### DIFF
--- a/.changeset/bright-tools-recover.md
+++ b/.changeset/bright-tools-recover.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Return connector tool failures as model-visible results so agents can self-correct.
--- a/src/connectors/tools.ts
+++ b/src/connectors/tools.ts
@@ -200,7 +200,42 @@ export function createOpenWikiConnectorTools(): StructuredToolInterface[] {
           ),
         ),
     }),
-  ];
+  ].map(withToolErrorsAsResults);
+}
+
+/**
+ * Return thrown tool errors as the tool *result* instead of rejecting (#427).
+ *
+ * A thrown tool error aborts the agent run, and even when the run survives, the
+ * model never sees the message. That matters because these messages are written
+ * for the model — `callMcpTool` answers a wrong tool name with
+ *
+ *   MCP tool notion-get-page-content was not returned by tools/list for notion.
+ *   Run openwiki_list_mcp_tools first and use an exact discovered name.
+ *
+ * which is precisely the hint needed to retry correctly. Returning it lets the
+ * model self-correct; throwing it just ends the turn.
+ *
+ * Errors are surfaced, not swallowed: the result is prefixed `Tool error:` so a
+ * failure is never mistaken for data, and the run transcript still shows it.
+ */
+export function withToolErrorsAsResults(
+  tool: DynamicStructuredTool,
+): DynamicStructuredTool {
+  const originalFunc = tool.func.bind(tool) as (
+    ...args: Parameters<DynamicStructuredTool["func"]>
+  ) => Promise<unknown>;
+
+  tool.func = (async (...args: Parameters<DynamicStructuredTool["func"]>) => {
+    try {
+      return await originalFunc(...args);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      return `Tool error: ${message}`;
+    }
+  }) as DynamicStructuredTool["func"];
+
+  return tool;
 }
 
 async function listConnectors() {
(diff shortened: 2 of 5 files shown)


Mr-Neutr0n:
Rebased onto the latest `main`; this applies cleanly again.

Mr-Neutr0n:
Thanks, I followed the changeset bot request and added a patch changeset for this user-facing behavior change. I also found one remaining repository test that still expected the MCP-backed guard to reject; it now asserts the labelled Tool error result, while preserving the same message. Prettier, ESLint, and typecheck pass; the focused connector tests pass (28/28), and the broader suite passes 2,030 tests when excluding an unrelated ingestion-run timeout. Commit bb3ccce is pushed for review.

Mr-Neutr0n:
Hi @ousamabenyounes - a gentle nudge on this connector fix (tool errors returned as results so the model can self-correct); still mergeable. Happy to adjust anything, and fine to close if not wanted.

PR #520, opened 2026-07-29
fix(agent): fail the run when output truncation drops a tool call (#458)

### LINKED ISSUE #458: [Bug]: Truncated model tool calls can end runs without an error

Mr-Neutr0n:
Fixes #458.

When a model reaches its output-token limit while streaming a large tool call, the v3 stream reports it twice:

```
content-block-finish   content.type = "invalid_tool_call"
message-finish         reason       = "length"
```

Both were classified as "no text to emit" and dropped:

```ts
if (
  event === "message-start" ||
  event === "message-finish" ||
  event === "content-block-finish" ||
  event === "error"
) {
  return "";
}
```

The incomplete tool call is never executed, so a requested `write_file` / `edit_file` simply does not happen — and the outer run could still finish without an error, leaving the file change missing and nothing explaining why. That is the worst shape a failure can take: silent, and indistinguishable from success.

## The fix

`detectTruncatedOutput()` recognises both signals, and the stream loop throws an actionable message naming the tool that was cut off, suggesting a smaller scope or a higher output limit.

Throwing there lands in the **existing** catch, which already cleans up the temporary plan file and records the run as `"interrupted"` rather than a no-op — so the next update is not skipped against a partial wiki. That was already the right machinery; it just never got invoked for this case.

Both spellings of the finish field are honoured (`reason` and `finish_reason`); providers are not consistent about it.

## Deliberately narrow

Only those two signals throw. A normal stop, an ordinary `tool_use` block, text deltas, and the `tools` channel are all untouched — each with a test. A false positive here would abort healthy runs, which is worse than the bug being fixed.

## Tests

`test/truncated-output-detection.test.ts` — 10 cases, including one that pins the old invisibility: `parseStreamEvent` still returns `null` for both chunks, which is *why* the run could finish reporting success.

Full suite **779 passed**, typecheck and `lint:check` clean.

Related: #474 (configurable output token limit) would give users the lever this error message points at.

---

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -382,6 +382,15 @@ async function runOpenWikiAgentCore(
 
   try {
     for await (const chunk of stream) {
+      // A truncated tool call is not a "no text" chunk to skip: the write the
+      // model asked for never happened, so the run must fail loudly rather than
+      // finish looking successful (#458). Throwing here lands in the catch
+      // below, which records the run as interrupted and cleans up.
+      const truncation = detectTruncatedOutput(chunk);
+      if (truncation) {
+        throw new Error(truncation);
+      }
+
       const event = parseStreamEvent(chunk);
 
       if (event) {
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
One scope decision worth your explicit sign-off, since I'd rather flag it than have you find it.

`message-finish` with `reason: "length"` fires whenever the model hit its output ceiling — **including a long text answer that got clipped with no tool call involved**. Today such a run completes and shows the partial text; with this PR it fails.

I think that's correct for OpenWiki specifically: the command exists to write files, so a clipped response means the page is incomplete either way, and failing loudly beats persisting a half-written wiki. It is nonetheless broader than the issue's tool-call framing, so if you'd rather only `invalid_tool_call` hard-fail and `reason: "length"` merely warn, that's a one-line change and I'll push it.

I also stress-tested the detector against benign shapes to make sure it can't abort healthy runs:

```
no false positives across 11 benign shapes
  message-finish reason=stop | tool_calls | end_turn | (absent)
  message-start
  content-block-finish  type=text | tool_use | (no content) | content=null
  content-block-delta   input_json_delta
  error
all 3 truncation shapes detected
```

A false positive here is worse than the bug, which is why the detector is deliberately narrow and why the `tools` channel isn't inspected at all.

colifran:
Closing in favor of #457

PR #521, opened 2026-07-29
feat: add MiniMax as an inference provider

octo-patch:
Reason: OpenWiki's inference provider registry had no first-class MiniMax provider, so its models were only reachable through the generic OpenAI-compatible endpoint.

## What changed

Adds MiniMax as a first-class, selectable inference provider, following the
existing OpenAI-compatible API-key provider pattern (Baseten, Fireworks, NVIDIA
NIM).

- `src/constants.ts`
  - New `MINIMAX_API_KEY` / `MINIMAX_BASE_URL` env-key constants.
  - `minimax` added to the `OpenWikiProvider` type and the
    `SELECTABLE_OPENWIKI_PROVIDERS` onboarding list.
  - New `PROVIDER_CONFIGS.minimax` entry pinned to the global endpoint
    (`https://api.minimax.io/v1`) with a `MINIMAX_BASE_URL` override, plus the
    `MiniMax-M3` and `MiniMax-M2.7` model presets.
  - `resolveConfiguredProvider` now selects `minimax` when only `MINIMAX_API_KEY`
    is present.
- `src/env.ts` — `MINIMAX_API_KEY` / `MINIMAX_BASE_URL` registered as managed
  env keys, and `MINIMAX_BASE_URL` wired into the base-URL diagnostic warnings
  and treated as a non-secret diagnostic key.
- `src/credentials.tsx` — onboarding article for `minimax` ("a MiniMax").
- `README.md` — provider list and the alternative-base-URL section document
  MiniMax, including how to point `MINIMAX_BASE_URL` at the other documented
  regional endpoint.
- `test/constants.test.ts` — new coverage for the provider config, model
  presets, default model, base-URL resolution/override, and credential
  auto-selection.

The provider resolves through the existing default `ChatOpenAI` construction
path, so no dispatch change in `createModel` is required.

## Checks

- `pnpm run typecheck`
- `pnpm exec vitest run test/constants.test.ts test/create-model.test.ts test/env.test.ts` (105 passed)
- `pnpm exec prettier --check` and `pnpm exec eslint` on the changed files

Native install lifecycle scripts (`better-sqlite3`) were skipped because the
build toolchain is unavailable in this environment; this is unrelated to the
change and does not affect the checks above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ notes.
 
 ## Customizing
 
-OpenWiki supports OpenAI (with an API key or a ChatGPT login), OpenRouter, Gemini (AI Studio), Gemini Enterprise (Vertex AI), Nebius Token Factory, Fireworks, Baseten, NVIDIA NIM, an OpenAI-compatible provider, AWS Bedrock, Anthropic, and GitHub Copilot out of the box. The onboarding default is OpenAI with `gpt-5.6-terra`, and each inference provider also includes pre-defined model options plus support for custom model IDs.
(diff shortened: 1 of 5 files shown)


colifran:
Thanks for the PR - our preference right now is to continue supporting openai compatible providers through the openai compatible provider that is already set up.

PR #529, opened 2026-07-30
feat: add OpenWiki retrieval tools

bracesproul:
## What changed

- Adds the OpenWiki retrieval MCP server and package binary.
- Implements repository indexing, keyword/semantic ranking, change-surface retrieval, and focused search.
- Adds retrieval-focused tests.
- Extends the generated agent guidance with instructions specific to the retrieval tools.

## Why

Give coding agents concise, source-grounded repository evidence without requiring broad exploratory scans.

## Dependency

Stacked on #528 so the generic prompt improvements can merge independently before the retrieval tools and their tool-specific guidance.

## Validation

- 10 retrieval tests
- Typecheck
- Formatting

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,10 @@ When working in this repository, read the OpenWiki quickstart first, then follow
 
 ## OpenWiki
 
-This repository uses OpenWiki for recurring code documentation. Start with `openwiki/quickstart.md`, then follow its links to architecture, workflows, domain concepts, operations, integrations, testing guidance, and source maps.
+This repository has a generated `openwiki/` evidence index. It is optional just-in-time context, not required startup reading.
+
+- Treat source code and tests as authoritative. A brief's unknowns and review items are verification gaps, not automatic requirements.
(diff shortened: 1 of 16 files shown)


Bennoo:
Seems great! But do i understand correctly if i say it is dedicated to other projects that needs the doc?
I mean if I have project A with an openwiki, I am working in project B that uses project A and so, needs the accurate documentation? is this related to that use case?
Thanks

colifran:
@Bennoo I think this is more scoped to retrieval from within a wiki scoped to a single repo.

Bennoo:
Great! Thanks for the clarification. 
I discovered recently Mintlify (https://www.mintlify.com/) which is a documentation service exposing a MCP server for your documentation. That MCP is exposing a tool which is basically a virtual Filesystem exposing the files. I was thinking about the fact that openwiki can be tailored for that, exposing the wiki as a virtual filesystem for agents that needs to dig into the documentation from an external POV.
But maybe this is not the right place to discuss that.

PR #547, opened 2026-07-31
fix: reject malformed OpenWiki managed markers

### LINKED ISSUE #537: [Bug]: Orphaned OpenWiki markers can delete hand-written AGENTS.md or CLAUDE.md content
## Description

The managed-snippet updater finds the first `<!-- OPENWIKI:START -->` and first `<!-- OPENWIKI:END -->` independently. If an instruction file contains an orphaned start marker, the first setup run appends a new complete managed block. On the next setup run, the old start marker is paired with the end marker from the appended block, and all intervening user content is replaced.

This is a destructive, repeatable data-loss bug affecting both `AGENTS.md` and `CLAUDE.md`.

## Affected code

`src/code-mode.ts:181-203`, especially:

- lines 195-196 select the first start and end markers globally;
- lines 197-200 replace everything between them if the end appears after the start;
- malformed or duplicate marker states are otherwise handled by appending another complete block.

## Steps to Reproduce

1. Create `AGENTS.md` with:

```md
# Project instructions

<!-- OPENWIKI:START -->
DO NOT DELETE: hand-written project policy
```

2. Run `ensureCodeModeRepoSetup(repo)` once. A new complete OpenWiki block is appended and the hand-written policy remains.
3. Run `ensureCodeModeRepoSetup(repo)` a second time.
4. Observe that the hand-written policy has disappeared.

Equivalent minimal behavior:

- after run one: the file contains an orphaned start, user content, then a complete start/end block;
- during run two: `indexOf(START)` returns the orphaned marker and `indexOf(END)` returns the appended block's end;
- replacement deletes the entire span.

I reproduced this directly against the current replacement algorithm.

## Expected Behavior

Malformed or ambiguous managed markers must never cause deletion of content OpenWiki cannot prove it owns. Setup should either fail with an actionable message or repair only a demonstrably valid managed block.

## Actual Behavior

A second setup run deletes every byte from the orphaned start marker through the appended end marker.

## Why this matters

These files contain high-value repository instructions and are explicitly meant to combine human-authored content with OpenWiki's managed block. A partial manual edit, merge conflict, interrupted write, or malformed generated file can lead to silent loss on a routine update.

## Suggested direction

Parse marker pairs as a small validated structure:

- exactly one correctly ordered start/end pair: replace it;
- no markers: append the block;
- orphaned, reversed, nested, or duplicate markers: do not rewrite destructively; return an actionable error or use a conservative repair that preserves all non-owned text.

Consider writing atomically after validation as well.

`test/code-mode.test.ts` covers a valid pair, no markers, and idempotence. It has no cases for orphaned starts, orphaned ends, reversed markers, or duplicate blocks.

## Environment

- OS: macOS
- Node.js: v24.14.1

GautamSharma99:
## Summary

Fixes #537.

OpenWiki now refuses to update `AGENTS.md` or `CLAUDE.md` when their managed-block markers are orphaned, reversed, or duplicated. The original file is preserved byte-for-byte, and both files are validated before either write begins.

A patch changeset and regression coverage are included.

## Problem

The existing updater independently selected:

```ts
const startIndex = currentContent.indexOf(OPENWIKI_AGENTS_SNIPPET_START);
const endIndex = currentContent.indexOf(OPENWIKI_AGENTS_SNIPPET_END);
```

It replaced the range whenever the first end marker appeared after the first start marker. Otherwise, it appended a new complete block.

That creates a destructive two-run sequence for an orphaned start marker:

1. The first run sees no usable pair and appends a complete managed block.
2. The file now contains the old orphaned start, user-authored content, a second start, and an end.
3. The second run pairs the first start with the appended end.
4. Everything in that span—including hand-written instructions—is replaced.

Duplicate and reversed marker layouts are similarly ambiguous because first-match selection cannot prove which bytes OpenWiki owns.

## Safety invariant

The updater now accepts only two unambiguous states:

| Marker state | Behavior |
| --- | --- |
| No start and no end marker | Append a fresh managed block |
| Exactly one start followed by exactly one end | Refresh that block in place |
| Orphaned start or end | Reject without writing |
| Reversed markers | Reject without writing |
| Duplicate starts, ends, or complete blocks | Reject without writing |

Malformed states fail with an actionable message naming the affected file and explaining the expected marker structure. The message asks the user to repair or remove the markers and retry.

This deliberately fails rather than guessing. With ambiguous markers, code cannot distinguish OpenWiki-owned content from hand-written repository instructions, so automatic repair risks data loss.

### Validate before write

`writeCodeModeAgentSnippets()` is now a two-phase operation:

1. Read and prepare updates for both `AGENTS.md` and `CLAUDE.md`.
2. Validate each file's marker structure.
3. Only after every preparation succeeds, write the prepared contents.

This means malformed `AGENTS.md` cannot cause a partial setup that creates or refreshes `CLAUDE.md`, and vice versa.

### Exact marker validation

For a file containing markers, the updater requires:

- the first start marker to also be the last start marker;
- the first end marker to also be the last end marker;
- the single end marker to occur after the single start marker.

Only then is the existing replacement behavior used.

Marker-free and valid-pair behavior remains unchanged.

## Reproduction fixed

Given:

```md
# Project instructions

<!-- OPENWIKI:START -->
DO NOT DELETE: hand-written project policy
```

Every setup attempt now:

- rejects with a malformed-marker error;
- leaves the original bytes unchanged;
- does not append a second managed block;
- does not create the sibling agent file as a partial side effect.

## Tests

Added table-driven regression coverage for:

- orphaned start markers;
- orphaned end markers;
- reversed markers;
- duplicate complete managed blocks.

Each case runs setup twice to cover the original second-run deletion sequence and asserts:

- both calls reject with the actionable error;
- `AGENTS.md` remains byte-for-byte identical;
- `CLAUDE.md` is not created before validation completes.

Existing tests continue to verify:

- creation when markers are absent;
- valid managed-block replacement while preserving surrounding content;
- marker-free append behavior;
- idempotence;
- workflow preservation.

## Related work

PR #362 addresses legacy unmarked `## OpenWiki` sections, not malformed managed markers. This PR is intentionally limited to ambiguity/data-loss prevention. Both touch the same updater and may require a small rebase if #362 lands first.

## Validation

- `corepack pnpm test` — 67 files, 795 tests passed
- `corepack pnpm exec vitest run test/code-mode.test.ts` — 13 tests passed
- `corepack pnpm exec changeset status`
- `git diff --check`


colifran (approved):
Thanks @GautamSharma99!

PR #548, opened 2026-07-31
fix: honor language changes on clean updates

### LINKED ISSUE #538: [Bug]: Clean-update no-op skips explicit --language changes
## Description

An explicit language change is ignored when the repository has no source changes since the last OpenWiki update. The no-op predicate considers only whether a user message is present; it does not consider `options.language`.

As a result, a command such as `openwiki code --update --language fr --print` can report that no repository changes were detected and return before the translation plan is created.

## Affected code

- `src/agent/utils.ts:179-181`: `shouldCheckUpdateNoop(options)` returns only `!options.userMessage?.trim()`.
- `src/agent/index.ts:162-181`: the clean-tree early return happens before provider/model setup and before `runOpenWikiAgentCore()`.
- `src/agent/index.ts:310-319`: the translation plan that honors a requested language is therefore never reached.
- `src/startup.ts:100-130`: the pre-credential clean-update shortcut calls the same predicate and can also treat a language-changing command as skippable.

## Steps to Reproduce

1. Initialize an OpenWiki repository in English and commit the generated wiki and `.last-update.json`.
2. Ensure the repository is clean and HEAD matches the saved metadata.
3. Run:
   `openwiki code --update --language fr --print`
4. Observe:
   `No repository changes detected since the last OpenWiki update; skipping agent run.`
5. Inspect the wiki; existing pages remain in English and persisted language does not change to French.

## Expected Behavior

An explicit request to change the wiki language should run the translation pass even when source code and documentation inputs are otherwise unchanged.

## Actual Behavior

The clean-tree optimization returns before language comparison or translation occurs.

## Why this matters

Changing language is a state transition, not a source-code update. The current optimization makes the documented `--language` option unreliable precisely on stable repositories, and scheduled/print runs can exit successfully while doing none of the requested work.

## Suggested direction

Make the no-op decision aware of language intent:

1. Read the persisted wiki language.
2. Normalize the explicitly requested locale.
3. Skip only when no user message exists and the requested language is absent or equivalent to the persisted language.

At minimum, any explicit `options.language` should bypass the no-op shortcut. Apply the same rule in both `runOpenWikiAgent` and the startup pre-credential optimization.

`test/update-noop.test.ts` tests user-message behavior but does not pass a language option. An integration-level test should assert that a clean update with a differing language is not skipped, while the same-language case may still skip.

## Environment

- OS: macOS
- Node.js: v24.14.1

GautamSharma99:
## Summary

Fixes #538.

Clean repository updates now treat an explicit primary-language change as meaningful work. A command such as:

```sh
openwiki code --update --language fr --print
```

will no longer return through the clean-tree no-op path when the persisted wiki language is English. It proceeds through provider setup and reaches the existing whole-wiki translation pass.

This PR also keeps the no-op optimization for equivalent language variants, wires the same decision into the pre-credential startup shortcut, centralizes primary-language comparison, adds regression tests, and includes a patch changeset.

## Problem

The update flow checked for a clean-tree no-op before creating run context or resolving the translation plan:

```ts
if (command === "update" && shouldCheckUpdateNoop(options)) {
  const noopStatus = await getUpdateNoopStatus(cwd, openWikiIgnore);
  if (noopStatus.shouldSkip) return ...;
}
```

`getUpdateNoopStatus()` considered metadata, git HEAD, worktree changes, ignored paths, and interrupted runs, but received no language intent. On a clean tree it returned `shouldSkip: true` even when `options.language` requested a full language switch.

The same bug existed one layer earlier in `resolveStartupCommand()`. A credential-less `--print` run could use the clean-update shortcut and return successfully without requiring credentials, even though translation needed a model call.

### Language-aware no-op status

`getUpdateNoopStatus()` now accepts the requested language and compares it with the persisted `.last-update.json` language before evaluating git state.

When their primary language subtags differ it returns:

```ts
{
  shouldSkip: false,
  reason: "output language changed"
}
```

An absent persisted language is treated as English, matching the existing translation behavior for metadata written by older OpenWiki versions.

The requested language is validated and canonicalized through `resolveLanguage()` before comparison. Invalid programmatic values therefore do not create a false language switch; the CLI already drops them with its existing warning.

### Same semantics as translation

The translation middleware already intentionally compares primary subtags so region-only changes do not trigger a whole-wiki rewrite—for example, `en` and `en-GB` are considered equivalent.

That comparison is now centralized in `getPrimaryLanguageSubtag()` and shared by:

- the clean-update no-op decision;
- `resolveTranslationPlan()`.

This prevents the optimization and translation pass from drifting into different definitions of a language switch.

### Both early-return paths updated

The requested language is passed into no-op evaluation from:

- `runOpenWikiAgent()`, which owns the actual clean-tree early return;
- `canSkipCleanUpdateBeforeCredentials()`, which lets genuine print-mode no-ops exit without configured provider credentials.

A real language change can therefore no longer be skipped at either layer.

## Behavior matrix

| Persisted language | Requested language | Clean tree behavior |
| --- | --- | --- |
| English or absent | none | Skip as before |
| English or absent | `fr` | Run translation/update |
| `fr` | `de` | Run translation/update |
| `en` | `en` | Skip |
| `en` | `en-GB` | Skip; same primary language |
| `pt-BR` | `pt-PT` | Skip; same primary language |
| any | invalid/unrecognized value | Existing warning/fallback behavior; no false switch |

### No-op regression coverage

`test/update-noop.test.ts` now verifies that:

- a clean English wiki requested as French returns `output language changed`;
- an equivalent `en-GB` request remains a no-op.

### Startup regression coverage

`test/startup.test.ts` now verifies that a clean `--print` update requesting French does not use the credential-free no-op shortcut. With no provider key configured, startup correctly reports the missing credential because model-backed translation must run.

### Shared language comparison

`test/language.test.ts` covers:

- region variants mapping to the same primary subtag;
- absent legacy metadata defaulting to English;
- malformed persisted values remaining a safe mismatch.

Existing translation tests continue to verify full-pass versus region-only behavior through the same shared helper.

## Validation

- `corepack pnpm test` — 67 files, 797 tests passed
- `corepack pnpm exec vitest run test/update-noop.test.ts test/startup.test.ts test/translation-middleware.test.ts test/language.test.ts` — 48 tests passed
- `corepack pnpm exec changeset status`
- `git diff --check`


colifran (approved):
Looks good - thanks @GautamSharma99!

PR #550, opened 2026-07-31
fix: sanitize streamed model output before terminal rendering

### LINKED ISSUE #540: [Bug]: Streamed model output can emit terminal control sequences including OSC 52
## Description

The interactive CLI renders model-produced Markdown tokens directly through Ink without removing terminal control characters. Model output is influenced by repository files and connector content, both of which are explicitly treated elsewhere as untrusted evidence.

Control sequences embedded in streamed text can therefore reach the user's terminal. This includes OSC sequences such as OSC 52, which some terminals interpret as a clipboard write, as well as display-manipulation sequences that can spoof or obscure CLI output.

## Affected code

- `src/cli.tsx:1525-1541`: `MarkdownText` lexes the supplied string without first sanitizing it.
- `src/cli.tsx:1624-1627`: plain and escaped token text is returned directly.
- `src/cli.tsx:1653-1655`: code spans are rendered directly.
- `src/cli.tsx:1673-1677`: fallback token content is rendered directly.
- `src/cli.tsx:3648-3667`: a control-character sanitizer already exists, but is used for compact header values rather than streamed assistant text.

## Steps to Reproduce

1. Render an Ink `Text` node containing:
   ```
   before\u001b]52;c;SGVsbG8=\u0007after
   ```
2. Using the project's Ink 5.x dependency, inspect the rendered frame or write it to a compatible terminal.
3. Observe that the OSC 52 sequence remains in the output. I reproduced the rendered frame retaining the full ESC/OSC/BEL sequence.
4. The same content can arrive through an assistant text event after being copied or induced from repository/connector content.

## Expected Behavior

Untrusted model/tool text should be rendered as visible text or removed; it should never be interpreted as terminal control instructions. OpenWiki's own intentional styling should continue to be produced by Ink components.

## Actual Behavior

Raw C0/C1 and ESC-based sequences from model content are passed into Ink. Ink parses ANSI styling but does not remove all terminal control protocols; OSC 52 survives rendering.

## Why this matters

This crosses a trust boundary from repository or remote-source content through an LLM into a developer's terminal. Depending on terminal support and configuration, it can overwrite clipboard data, alter the terminal title, create misleading hyperlinks, erase/reposition visible output, or make a malicious run appear benign.

## Suggested direction

Sanitize every untrusted string at the boundary before Markdown lexing/rendering:

- remove ESC/CSI/OSC/DCS sequences, BEL, and unsafe C0/C1 controls;
- preserve only intentional newline and tab semantics;
- do not run already-styled Ink output through the same sanitizer.

A maintained ANSI/control-sequence stripping implementation is preferable to filtering only the ESC byte because sequences have multiple terminators and encodings.

Add rendering tests containing OSC 52, OSC 8, CSI clear/cursor sequences, BEL, CR, and C1 controls in paragraph text, inline code, fenced code, tables, and raw-token fallbacks. Assert that no unsafe bytes survive the rendered frame.

## Environment

- OS: macOS
- Node.js: v24.14.1
- Ink: 5.x

GautamSharma99:
## Summary

Fixes #540.

Interactive assistant Markdown was passed to Ink after lexing without removing terminal control protocols. Repository and connector content can influence model output, so OSC 52 clipboard writes, OSC 8 hyperlinks, CSI cursor/display controls, and related sequences could reach a user terminal.

## Changes

- Added `stripTerminalControlSequences`, a boundary sanitizer that removes:
  - ESC-prefixed OSC, CSI, DCS, SOS, PM, and APC sequences;
  - their C1 equivalents, including BEL/ST-terminated and unterminated string controls;
  - unsafe C0/C1 controls, DEL, BEL, and carriage returns.
- Preserved newline and tab semantics needed for Markdown layout.
- Applied the sanitizer to the complete Markdown string before `marked.lexer`, covering paragraphs, headings, lists, blockquotes, inline/fenced code, tables, HTML, and fallback tokens without re-sanitizing Ink-generated styling.
- Added regression tests for OSC 52/OSC 8, CSI display/cursor controls, C1 controls, BEL, carriage return, whitespace preservation, and unterminated OSC/DCS payloads.
- Added a patch changeset.

## Security rationale

The model output is untrusted terminal text. Sanitizing before Markdown parsing ensures no rendered token path can accidentally bypass the control-sequence policy, while Ink remains responsible for intentional application styling after the untrusted boundary.

## Validation

- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm test` — 67 files, 795 tests passed


colifran (approved):
Thanks, @GautamSharma99!

PR #552, opened 2026-07-31
fix: preserve parallel connector runs and state

### LINKED ISSUE #542: [Bug]: Millisecond connector run IDs and unlocked state writes lose parallel MCP results
## Description

Connector run IDs are ISO timestamps with millisecond precision. Parallel calls started in the same millisecond receive the same run ID and therefore target the same raw-data directory and fixed filenames.

Connector state is also updated through an unlocked read-modify-write. MCP discovery/tool calls read state before performing network work, then later overwrite the whole state file. Parallel calls can consequently overwrite raw results and discard each other's run history.

These are two parts of the same concurrency failure: connector operations do not have a unique run identity or serialized state commit.

## Affected code

- `src/connectors/io.ts:72-74`: `createRunId()` contains no random or monotonic component.
- `src/connectors/io.ts:33-57`: state is read and written as a whole document without locking or compare-and-merge.
- `src/connectors/io.ts:59-67`: raw paths use `runId` plus a caller-supplied fixed filename.
- `src/connectors/mcp-runtime.ts:47-67`: MCP discovery reads state early, writes `mcp-tools.json`, then writes derived state.
- `src/connectors/mcp-runtime.ts:77-120`: MCP tool calls use the same pattern and fixed `mcp-tool-result.json`.

## Steps to Reproduce

1. Freeze `Date` so two operations see the same millisecond.
2. Start two `callMcpConnectorTool()` calls for the same connector with different results using `Promise.all()`.
3. Allow both calls to complete after both have read the original state.
4. Observe:
   - both calls report the same run ID;
   - both write `<runId>/mcp-tool-result.json`, so one result replaces the other;
   - each derives state from the same old snapshot, and the last state write drops the other run.

Even without a frozen clock, rapid synchronous calls frequently generate repeated timestamp IDs.

## Expected Behavior

Every invocation should have a collision-resistant run ID, preserve its own immutable raw output, and merge its history/cursor update without losing concurrently completed runs.

## Actual Behavior

Parallel invocations can share paths and use last-writer-wins state replacement.

## Why this matters

LangChain/LangGraph agents may issue independent tool calls concurrently. Lost raw evidence breaks provenance and can make the agent read a result from the wrong invocation. Lost state history also undermines cursoring, diagnostics, and latest-run selection.

## Suggested direction

- Generate IDs with `crypto.randomUUID()` or append a cryptographically random/monotonic suffix to the sortable timestamp.
- Serialize state updates per connector, or implement an atomic update API that re-reads and merges immediately before replacement.
- Write state through a unique temp file and atomic rename.
- Ensure run history is deduplicated by unique run ID.

Add a fixed-clock collision test plus a parallel MCP-call test asserting distinct raw paths, preserved payloads, and both runs in `state.json`.

## Environment

- OS: macOS
- Node.js: v24.14.1

GautamSharma99:
## Summary

Fixes #542.

Connector run IDs previously used only an ISO timestamp, so calls in the same millisecond shared raw-data directories and fixed filenames. MCP operations also read connector state before network work and later replaced the whole document from that stale snapshot, dropping parallel run history.

## Changes

- Add a UUID suffix to every connector run ID while retaining the sortable timestamp prefix.
- Add `updateConnectorState`, a process-local FIFO transaction per connector that re-reads state immediately before applying an updater and committing it.
- Move MCP discovery and tool-call run recording to the atomic update API, so parallel operations preserve both run summaries and cursor/state fields.
- Write connector state and raw JSON through unique same-directory temporary files followed by atomic rename, avoiding partial files and temp-path collisions.
- Add regression coverage for distinct run IDs and deterministic concurrent state transactions, asserting both runs survive in `state.json`.
- Add a patch changeset.

## Why this fixes the race

Each invocation now owns a unique raw directory, so `mcp-tools.json` and `mcp-tool-result.json` cannot replace evidence from another invocation. MCP state commits are serialized per connector and each queued commit reads the latest file, so a later completion merges with the prior run instead of writing an old snapshot over it. Atomic replacement preserves the existing file until the new JSON is complete.

## Validation

- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm test` — 68 files, 793 tests passed


AmirF194 (commented):
Nice fix for the MCP race in #542, verified the new test passes on this branch.

One scope question: `readConnectorState`/`writeConnectorState` (src/connectors/io.ts) are unchanged and still used with the same read-before-work, write-after-work shape #542 describes, in every other connector source: gmail.ts (read L102, write L166), slack.ts (L184, L345), git-repo.ts (L64, L118), x.ts (L82, L179), hackernews.ts (L114, L242), langsmith/index.ts (L154, L238), web-search.ts (L75, L156), and mcp.ts (L45, L164, a different file from mcp-runtime.ts). I reproduced the same lost-run-history failure #542 describes against these two functions directly (two reads, then two writes, mirroring gmail.ts's shape): the first run's entry never makes it into state.json, same as the MCP case before this PR.

Was the fix intentionally scoped to the MCP tool bridge from #542's own repro, or would it be worth moving these connectors onto `updateConnectorState` too? It's already generic per connector id, so the wiring looks mechanical if you want it in this PR or a follow-up.

PR #554, opened 2026-07-31
fix: exclude unmeasurable LangSmith runs from outliers

### LINKED ISSUE #544: [Bug]: LangSmith runs without measurable latency are labeled as latency outliers
## Description

LangSmith trace sampling labels a capped portion of non-error root runs as latency outliers. Runs whose latency cannot be computed are assigned the sentinel value `-1`, but they remain eligible for the outlier slice.

If the outlier budget exceeds the number of runs with both start and end timestamps, incomplete/unmeasurable runs are labeled `outlier` despite having no latency evidence. If all candidate roots lack an end time, up to one quarter of them are still labeled as latency outliers.

## Affected code

`src/connectors/sources/langsmith/runs.ts:51-66`:

- `outlierBudget` is based on the entire non-error pool;
- line 61 maps `latencyMs(run) ?? -1`;
- sorting places missing values last but does not remove them;
- `slice(0, outlierBudget)` fills any remaining outlier slots with those missing-latency runs.

The resulting bucket is subsequently described to the synthesis agent as the slowest non-errored roots.

## Steps to Reproduce

1. Pass four non-error root runs into the sampler.
2. Give all four a valid `start_time` but no usable `end_time`.
3. Use caps with `total >= 4` and `outlierCap >= 1`.
4. The computed budget is `floor(4 / 4) = 1`.
5. Observe that one run is returned with `bucket: "outlier"`, although `latencyMs()` was undefined for every run.

A mixed example also fails: with eight candidates, an outlier budget of two, and only one measurable run, the second outlier slot is filled by an unmeasurable run.

## Expected Behavior

Only runs with a valid measured latency should be eligible for the latency-outlier bucket. Unmeasurable runs may be baseline/incomplete samples or excluded, but should not be characterized as slow.

## Actual Behavior

The `-1` sentinel participates in budget filling and produces unsupported outlier labels.

## Why this matters

The LangSmith connector is intended to give coding agents grounded production evidence. Mislabeling active or malformed traces as latency outliers can direct investigation toward the wrong code paths and contaminate generated runtime documentation with a claim the trace does not support.

## Suggested direction

Filter to measurable candidates before both budget calculation and selection:

```ts
const measured = nonErrorRuns
  .filter((run) => !usedIds.has(run.id))
  .map((run) => ({ latency: latencyMs(run), run }))
  .filter((item): item is Measured => item.latency !== undefined);
```

Then cap against `measured.length`. Leave unmeasurable roots available for baseline selection only if that bucket is semantically appropriate, or track them as a separate incomplete category.

`test/langsmith-runs.test.ts` exercises known-latency roots and outlier caps. Add all-missing and partially measurable latency cases and assert that every outlier has a defined latency.

## Environment

- OS: macOS
- Node.js: v24.14.1

GautamSharma99:
## Summary

Fixes #544.

LangSmith sampling used `latencyMs(run) ?? -1` while calculating and filling the latency-outlier slice. When there were fewer measurable runs than the outlier budget, incomplete roots were labeled as slow despite having no latency evidence.

## Changes

- Build the outlier candidate pool from non-error roots with a valid, non-negative measured latency only.
- Calculate the proportional quarter cap from that measurable pool, and select only those candidates.
- Continue using the original non-error order to backfill remaining capacity as baseline, so incomplete runs remain available as context without unsupported outlier labels.
- Add regressions for all-missing latency and mixed measurable/unmeasurable pools.
- Add a patch changeset.

## Behavior

A run with no usable start/end timestamps can now be selected only as baseline (or omitted by the overall cap), never as a latency outlier. Every root in the outlier bucket has a defined measured latency.

## Validation

- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm test` — 67 files, 793 tests passed

### DIFF
--- a/.changeset/measured-langsmith-outliers.md
+++ b/.changeset/measured-langsmith-outliers.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: exclude unmeasurable LangSmith runs from latency outliers
--- a/src/connectors/sources/langsmith/runs.ts
+++ b/src/connectors/sources/langsmith/runs.ts
@@ -21,12 +21,12 @@ export interface SampleCaps {
 /**
  * Selects the sample from two lean root-run pools within the window, biased
  * toward anomalies: errors first (up to errorCap), then latency outliers among
- * the non-errored runs (up to outlierCap, at most a quarter of the non-errored
- * pool so the bucket stays a genuine tail rather than swallowing a small pull,
- * and the remaining budget), then the most-recent non-errored runs to backfill
- * to `total`. With no errors/outliers it degrades to all-baseline — the same
- * recency behavior as before. Runs are deduped by id; `nonErrorRuns` is assumed
- * most-recent-first.
+ * the measurable non-errored runs (up to outlierCap, at most a quarter of the
+ * measurable pool so the bucket stays a genuine tail rather than swallowing a
+ * small pull, and the remaining budget), then the most-recent non-errored runs
+ * to backfill to `total`. With no errors/outliers it degrades to all-baseline —
+ * the same recency behavior as before. Runs are deduped by id; `nonErrorRuns` is
+ * assumed most-recent-first.
  */
 export function selectSampleBuckets(
   errorRuns: Run[],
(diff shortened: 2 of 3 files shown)


colifran:
Thanks for fixing #544. We’re consolidating the measurable-latency fix in #665, which is current, conflict-free, and CI-clean. Closing this PR in favor of #665.

PR #555, opened 2026-07-31
fix: report telemetry send failures accurately

### LINKED ISSUE #545: [Bug]: Telemetry records sent=true after capture rejection or timeout
## Description

The telemetry client's timeout helper intentionally swallows promise rejection and resolves when either the operation settles or the timer fires. Because it returns no outcome, `capture()` unconditionally returns `true` after both failed requests and timeouts.

When `--telemetry-file` is used, the diagnostic file can therefore claim `sent: true` even though the event was rejected or never completed. This contradicts `capture()`'s documented return contract.

## Affected code

- `src/telemetry/client.ts:10-20`: documentation says the function returns whether the event actually sent.
- `src/telemetry/client.ts:35-56`: capture and shutdown are awaited through `withTimeout()`, then the function returns `true`.
- `src/telemetry/client.ts:62-71`: `promise.catch(() => undefined)` converts rejection to success, and the timeout branch also resolves without identifying which branch won.
- `src/telemetry/senders.ts:105-113`: the returned value is written into the local telemetry record as `sent`.

## Steps to Reproduce

Rejection case:

1. Mock `PostHog.captureImmediate()` with `Promise.reject(new Error("network down"))`.
2. Call `capture(event)`.
3. Observe that it resolves to `true`.

Timeout case:

1. Mock `captureImmediate()` with a promise that never settles.
2. Advance fake timers past `FLUSH_TIMEOUT_MS`.
3. Observe that `capture(event)` resolves to `true`.

Calling `recordRun()` with a telemetry file then persists `sent: true` in both cases.

## Expected Behavior

`capture()` should return `true` only when `captureImmediate()` fulfills before the deadline. It should return `false` for rejection and timeout while preserving the rule that telemetry never breaks the main command.

## Actual Behavior

All asynchronously settled outcomes, including failure and timeout, are reported as sent.

A synchronous throw happens to reach `recordRun()`'s outer catch, which is why the current synchronous-failure test does not expose the async bug.

## Why this matters

The telemetry tee is presented as the exact payload/status users can inspect. Incorrect success status makes telemetry loss invisible to maintainers and gives users false information about whether data left the process.

## Suggested direction

Have the timeout helper return a discriminated result such as `fulfilled | rejected | timed-out`, or allow rejection to propagate into a local catch in `capture()`. Return false for rejection/timeout. Keep shutdown best-effort and independent of the send outcome.

If delivery cannot truly be confirmed beyond HTTP request fulfillment, update the field/doc wording from "sent" to the precise guarantee being measured.

`test/telemetry.test.ts` covers success and a synchronous throw. Add:

- `captureImmediate.mockRejectedValue(...)`;
- a never-settling promise with fake timers;
- telemetry-file assertions that both produce `sent: false`;
- a shutdown rejection/timeout case that does not overwrite a successful capture result.

## Environment

- OS: macOS
- Node.js: v24.14.1

GautamSharma99:
## Summary

Fixes #545.

Telemetry `capture()` previously swallowed `captureImmediate()` rejection and timeout outcomes, then unconditionally returned `true`. Telemetry files could therefore claim `sent: true` when PostHog rejected the event or never completed.

## Changes

- Make the timeout helper return a boolean outcome: `true` only for fulfillment before the deadline, `false` for rejection or timeout.
- Convert synchronous SDK throws into the same failed-send outcome so capture remains best-effort without misleading success.
- Keep client shutdown in an independent best-effort path; shutdown rejection or timeout cannot overwrite a successful capture result.
- Clear the timeout when the send settles while still consuming a late rejection after a timeout.
- Add tests for async rejection, never-settling timeout, telemetry-file `sent: false` records, and the existing synchronous-failure guarantee.
- Add a patch changeset.

## Contract

`capture()` now returns `true` only when `captureImmediate()` fulfills before `FLUSH_TIMEOUT_MS`. It returns `false` when no key is configured, the SDK rejects/throws, or the deadline expires. Telemetry remains non-blocking and never breaks the main command.

## Validation

- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm test` — 67 files, 795 tests passed


colifran (approved):
Looks good. Thanks @GautamSharma99!

PR #566, opened 2026-08-01
fix: follow tools/list pagination so all mcp tools are callable

### LINKED ISSUE #565: [Bug]: MCP tools/list ignores nextCursor, so tools past page 1 are undiscoverable and uncallable
### Description

The MCP client never follows `tools/list` pagination. The MCP spec paginates `tools/list` with a top-level `nextCursor`: a client must re-issue the request with `{ "cursor": "<nextCursor>" }` until the field is absent. Both JSON-RPC clients in `src/connectors/mcp-client.ts` call it exactly once with empty params:

```ts
// src/connectors/mcp-client.ts:462 (StdioJsonRpcClient) and :608 (HttpJsonRpcClient)
listTools(): Promise<unknown> {
  return this.request("tools/list", {});
}
```

`extractToolValues` (`:365`) then reads only `result.tools` from that single page, so every tool after the first page is silently dropped from discovery.

That is not just a display gap: `callMcpConnectorTool` hard-rejects any tool name missing from the discovered set (`src/connectors/mcp-runtime.ts:88-92`):

```
MCP tool <name> was not returned by tools/list for <connector>. Run openwiki_list_mcp_tools first and use an exact discovered name.
```

So on any MCP server that paginates, valid read-only tools become permanently uncallable, and the error message points the user at `openwiki_list_mcp_tools` — which is itself truncated to page 1, so the advice can never resolve it.

Both transports are affected (stdio and HTTP).

### Steps to Reproduce

1. Configure an MCP connector against a server whose `tools/list` paginates — page 1 returns some tools plus a `nextCursor`, page 2 returns the rest. (Hosted servers with large tool sets, e.g. Notion, do this.)
2. Run `openwiki_list_mcp_tools` for that connector.
3. Note that only the page-1 tools are listed.
4. Call `openwiki_call_mcp_tool` with the exact name of a read-only tool that lives on page 2.

Minimal reproduction of the underlying behavior — a stub server that returns:

```json
{"tools": [{"name": "page-one-tool"}], "nextCursor": "cursor-2"}
```

then, for a request carrying `{"cursor": "cursor-2"}`:

```json
{"tools": [{"name": "page-two-tool"}]}
```

### Expected Behavior

`listMcpTools` discovers the complete tool set across all pages, so `page-two-tool` is listed and can be called.

### Actual Behavior

Only `page-one-tool` is discovered. The second request is never sent — `tools/list` is issued once with `{}` and `nextCursor` is ignored. Calling `page-two-tool` throws `MCP tool page-two-tool was not returned by tools/list`, even though the tool is valid and read-only.

### Environment

- OS: Windows 11
- Node.js version: v24.18.1
- OpenWiki version: `main` @ 398ebad

### Additional Context

The sibling stdio/HTTP clients both expose an identical private `request(method, params)`, so a single shared pagination helper can fix both without duplicating the loop. Any fix should bound the loop (stop on a repeated cursor and cap total pages) so a misbehaving server that always returns a cursor cannot spin forever.

I have a fix ready and will open a PR referencing this issue.

divya0795:
## Summary

Fixes #565.

The MCP spec paginates `tools/list` with a top-level `nextCursor`: a client must re-issue the request with that cursor until the field is absent. Both JSON-RPC clients in `src/connectors/mcp-client.ts` issued a single request with empty params and read only the first page, so every tool past page 1 was silently dropped from discovery.

That is not only a discovery gap. `callMcpConnectorTool` hard-rejects any tool name missing from the discovered set (`src/connectors/mcp-runtime.ts:88-92`), so on a paginating server valid read-only tools became permanently uncallable with `MCP tool <name> was not returned by tools/list` — and the error points the user at `openwiki_list_mcp_tools`, which was itself truncated to page 1, so the advice could never resolve it.

## Changes

- Added `collectPaginatedTools`, a shared helper that follows `nextCursor` until it is absent and concatenates the pages. It reuses the existing `extractToolValues`, so a malformed page still contributes nothing rather than throwing.
- Pointed both `StdioJsonRpcClient.listTools` and `HttpJsonRpcClient.listTools` at the helper — both transports were affected, and both expose an identical private `request(method, params)`, so neither needed its own loop.
- Bounded the loop twice over so a misbehaving server cannot hang discovery: it stops when a cursor repeats, and at a 100-page cap.
- Added a patch changeset.

## Behavior and safety

The first request is unchanged (`tools/list` with `{}`), and a server that returns no `nextCursor` still results in exactly one round-trip — no extra traffic for the non-paginating case. `listTools` still resolves to `{ tools }`, so `extractTools` and both call sites are untouched. Tool validation and the read-only policy checks in `mcp-runtime.ts` are unchanged; this only widens what discovery sees to what the server actually offers.

## How I tested it

Added four cases to `test/mcp-client.test.ts`, driving the real HTTP client end-to-end through the exported `listMcpTools` against a stubbed MCP server:

- a two-page server ⇒ both tools discovered, and the second request carries `{ cursor: "cursor-2" }`;
- a single-page server ⇒ exactly one `tools/list` request (regression guard against extra traffic);
- a server that repeats the same cursor ⇒ terminates instead of looping;
- a server that always returns a fresh cursor ⇒ terminates at the page cap.

Confirmed these are real regression tests: with the source change reverted, three of the four fail; all four pass with it.

Validation:

- `pnpm run format`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm test` — 872/873 pass. The one failure is `test/openwiki-ignore.test.ts > restricts shell execute while ignore rules are active`, which fails identically on an unmodified `main` on this machine (it shells out to `pwd`, so it is Windows-specific and unrelated to this change).

### DIFF
--- a/.changeset/paged-mcp-tool-lists.md
+++ b/.changeset/paged-mcp-tool-lists.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: follow `nextCursor` when listing MCP tools, so tools on a paginated server past the first page are discovered and callable instead of rejected as "not returned by tools/list"
--- a/src/connectors/mcp-client.ts
+++ b/src/connectors/mcp-client.ts
@@ -7,6 +7,9 @@ import {
 import type { McpConnectorConfig, McpReadOnlyOperation } from "./types.js";
 import { fetchWithResilience } from "./http.js";
 
+/** Upper bound on `tools/list` pages, so a bad cursor chain cannot spin forever. */
+const MAX_TOOL_LIST_PAGES = 100;
+
 type JsonRpcRequest = {
   id?: number;
   jsonrpc: "2.0";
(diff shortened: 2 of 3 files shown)


colifran (approved):
Very nice! Thanks @divya0795!

PR #568, opened 2026-08-01
fix: render array tool-call arguments as a value list

### LINKED ISSUE #567: [Bug]: array tool-call arguments render as 0=…, 1=… in the tool_start display
### Description

When a tool is called with an array as its input, the `tool_start` display renders it as `0=…, 1=…` instead of the intended comma-joined value list.

`formatToolArgs` in `src/agent/index.ts` has an explicit branch for arrays, but it is unreachable:

```ts
function formatToolArgs(input: unknown): string {
  const value = parseStringifiedJson(input);

  if (isRecord(value)) {                    // <- arrays land here first
    return Object.entries(value)
      .map(([key, argValue]) => `${key}=${formatToolValue(argValue)}`)
      .join(", ");
  }

  if (Array.isArray(value)) {               // <- dead branch
    return value.map(formatToolValue).join(", ");
  }
  ...
}
```

The cause is this file's `isRecord`, which does not exclude arrays:

```ts
function isRecord(value: unknown): value is Record<string, unknown> {
  return typeof value === "object" && value !== null;
}
```

`typeof [] === "object"` and an array is non-null, so the `isRecord` branch matches arrays and the `Array.isArray` branch below it can never run. `Object.entries(["a","b"])` then yields the numeric index pairs.

This is display-only — the tool still receives its real input — but the rendered call is misleading, and the dead branch shows the current output is not what was intended.

### Steps to Reproduce

1. Have the model call a tool whose input is a JSON array, e.g. `["a","b"]`.
2. Observe the `tool_start` line rendered for that call.

Equivalent via the exported stream-event parser:

```ts
parseStreamEvent({ event: "on_tool_start", name: "search", input: '["a","b"]' })
```

### Expected Behavior

```
search("a", "b")
```

as the unreachable `Array.isArray` branch intends.

### Actual Behavior

```
search(0="a", 1="b")
```

### Environment

- OS: Windows 11
- Node.js version: v24.18.1
- OpenWiki version: `main` @ 398ebad

### Additional Context

The sibling helper in `src/okf/frontmatter.ts` does exclude arrays in its equivalent predicate, so the two files disagree on what counts as a record. Narrowing the shared `isRecord` here would touch ~22 call sites in `src/agent/index.ts`, so the tightly-scoped fix is to order the `Array.isArray` check ahead of the `isRecord` check inside `formatToolArgs`, leaving every other call site untouched.

I have a fix ready and will open a PR referencing this issue.

divya0795:
## Summary

Fixes #567.

`formatToolArgs` in `src/agent/index.ts` has an explicit branch for array inputs, but it was unreachable. This file's `isRecord` does not exclude arrays, and an array is `typeof "object"` and non-null, so the record branch claimed arrays first and `Object.entries` rendered their numeric indices. A tool called with `["a","b"]` displayed as `search(0="a", 1="b")` instead of `search("a", "b")`.

Display-only — the tool still received its real input — but the rendered call was misleading, and the dead branch shows it was never the intended output.

## Changes

- Ordered the `Array.isArray` check ahead of the `isRecord` check inside `formatToolArgs`, which revives the branch that was already there, plus a comment recording why the order matters.

## Why not fix `isRecord` itself

Narrowing the shared `isRecord` to exclude arrays would be the broader change, and it is what `src/okf/frontmatter.ts` already does in its equivalent predicate — but `isRecord` has ~22 call sites in `src/agent/index.ts`, so that is a different, wider concern than this bug. Keeping the fix inside `formatToolArgs` leaves every other call site untouched, per the one-PR-one-change guidance in CONTRIBUTING. Happy to open that as a separate PR if you'd like it narrowed globally.

## How I tested it

Added `test/tool-args-formatting.test.ts`. `formatToolArgs` is module-private, so the tests drive it the way the CLI does — through the exported `parseStreamEvent`, whose `tools` branch builds the displayed `call` string:

- a stringified array input (`'["a","b"]'`) ⇒ `search("a", "b")`;
- a raw array input ⇒ same;
- mixed member types ⇒ `search(1, "two", true)`, no index keys;
- regression guards that object inputs still render `key=value`, a scalar still renders bare, and a null input still renders an empty argument list.

Confirmed these are real regression tests: with the source change reverted, the three array cases fail with exactly `search(0="a", 1="b")`; all six pass with it. There was no existing test coverage for `tool_start` argument rendering, hence the new file.

Validation:

- `pnpm run format`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm test` — 873/875 pass. The two failures are `test/openwiki-ignore.test.ts > restricts shell execute while ignore rules are active` (shells out to `pwd`) and a 5s-timeout flake in `test/mermaid-validate.test.ts`. Both fail identically on an unmodified `main` on this machine, so they are environmental (Windows / slow mermaid render) and unrelated to this change.

### DIFF
--- a/.changeset/listed-tool-array-args.md
+++ b/.changeset/listed-tool-array-args.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: display array tool-call arguments as a value list instead of `0=…, 1=…`
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1669,14 +1669,16 @@ function formatToolCallName(name: string): string {
 function formatToolArgs(input: unknown): string {
   const value = parseStringifiedJson(input);
 
+  // Checked ahead of isRecord: arrays are `typeof "object"` and non-null, so
+  // the record branch would otherwise claim them and render `0=…, 1=…`.
+  if (Array.isArray(value)) {
+    return value.map(formatToolValue).join(", ");
+  }
+
   if (isRecord(value)) {
     return Object.entries(value)
       .map(([key, argValue]) => `${key}=${formatToolValue(argValue)}`)
       .join(", ");
-  }
-
-  if (Array.isArray(value)) {
-    return value.map(formatToolValue).join(", ");
   }
 
   if (value === undefined || value === null) {
--- a/test/tool-args-formatting.test.ts
+++ b/test/tool-args-formatting.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, test } from "vitest";
+import { parseStreamEvent } from "../src/agent/index.ts";
+
+// `formatToolArgs` is module-private, so these drive it the way the CLI does:
+// through the exported stream-event parser, whose "tools" branch builds the
+// displayed `call` string from the tool's input.
+function makeToolStartChunk(name: string, input: unknown): unknown {
+  return {
+    type: "event",
+    method: "tools",
+    params: {
+      data: { event: "on_tool_start", input, name },
+      namespace: [],
+    },
+  };
+}
+
+function callFor(name: string, input: unknown): string {
+  const event = parseStreamEvent(makeToolStartChunk(name, input));
+
+  expect(event?.type).toBe("tool_start");
+
+  return (event as { call: string }).call;
+}
+
+describe("tool_start argument formatting", () => {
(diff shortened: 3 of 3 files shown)


colifran (approved):
Nice! Thanks @divya0795!

PR #590, opened 2026-08-04
feat: support capping openrouter output tokens

### LINKED ISSUE #588: Support overriding max_tokens for OpenRouter models
## Current behavior

When running OpenWiki with `OPENWIKI_PROVIDER=openrouter`, the agent loop always requests the
model's advertised maximum output tokens (`profile.maxOutputTokens` from
`@langchain/openrouter`'s model-capability table) as `max_tokens` on every request, regardless of
how much output is actually needed for the task.

For example, with `OPENWIKI_MODEL_ID=z-ai/glm-5.2`, every request sends `"max_tokens": 65536`,
because that's the model's ceiling in `@langchain/openrouter`'s bundled profile — not because
OpenWiki, the workflow, or the caller asked for that much.

## Why this is a problem

This prevents using OpenWiki on lower-credit OpenRouter accounts even when a much smaller output
budget (e.g. 8192 or 4096 tokens) would be entirely sufficient for the documentation update being
generated. OpenRouter rejects the request outright with a 402 ("This request requires more
credits, or fewer max_tokens...") rather than truncating or degrading gracefully, so the run fails
completely instead of just producing a smaller update.

We traced this end-to-end (workflow env vars → OpenWiki source → `@langchain/openrouter`'s request
building) and confirmed OpenWiki itself never sets a `maxTokens` field when constructing the model
client (`ChatOpenRouter` in `src/agent/index.ts`), and there's no `OPENWIKI_MAX_TOKENS`-style env
var, CLI flag, or config key anywhere in the codebase to override it.

## Desired behavior

Allow an explicit `max_tokens` override for OpenRouter (and ideally other providers) via one of:

- an environment variable (e.g. `OPENWIKI_MAX_TOKENS`)
- a config file setting
- a CLI flag

so accounts with a smaller credit balance can still run OpenWiki against models with a large
advertised output ceiling, by capping the per-request budget explicitly.

## Environment

- `openwiki` 0.3.0
- `@langchain/openrouter` 0.4.5
- Model: `z-ai/glm-5.2`
- Provider: `openrouter`

pawel-twardziak:
via OPENWIKI_OPENROUTER_MAX_TOKENS

Without an explicit max_tokens, OpenRouter's credit pre-check budgets for the model's full advertised output ceiling and rejects requests with 402 when the balance cannot cover that worst case. Let users cap per-request output tokens with a validated env var, following the existing OPENWIKI_OPENROUTER_PROVIDER_ONLY pattern.

Fixes #588


colifran (approved):
Nice. Thanks, @pawel-twardziak!

PR #599, opened 2026-08-05
fix: pass windows appdata/localappdata into mcp child env

### LINKED ISSUE #598: fix: pass Windows APPDATA/LOCALAPPDATA into MCP stdio child env
### Description

Stdio MCP subprocesses spawned by OpenWiki receive a tightly allowlisted environment (`buildChildEnv` in `src/connectors/mcp-client.ts`). The list includes `USERPROFILE` / `HOMEPATH` / `HOMEDRIVE` but not `APPDATA` or `LOCALAPPDATA`.

Many Windows MCP servers and Node tools expect those variables to resolve config, cache, and native dependency paths under the user profile. Without them, local MCP connectors can fail to start or misbehave only on Windows even though PATH and home vars are present.

### Expected Behavior

`buildChildEnv` forwards `APPDATA` and `LOCALAPPDATA` when set in the parent process (same treatment as `USERPROFILE`), without forwarding OpenWiki provider credentials.

### Actual Behavior

Those two variables are dropped from the child environment.

### Proposed fix

Add `APPDATA` and `LOCALAPPDATA` to `CHILD_ENV_ALLOWLIST` and cover them in `test/mcp-client.test.ts`.

realsudipkandel:
Fixes #598

Stdio MCP children already receive `USERPROFILE` / `HOMEPATH` / `HOMEDRIVE`, but not `APPDATA` or `LOCALAPPDATA`. Many Windows MCP servers need those to resolve config and cache directories.

This adds both keys to `CHILD_ENV_ALLOWLIST` and extends `test/mcp-client.test.ts`. Credential filtering is unchanged.

Verified with `pnpm exec vitest run test/mcp-client.test.ts` (6 passed).


colifran (approved):
Thanks @sudipawtg!

Issue #372, opened 2026-07-16
Post-generation consistency pass: dead internal links/anchors and cross-page drift in regenerated wikis


rstover-fo:
## The gap

The wiki-as-memory loop this tool implements is canonically *ingest → query → lint → write back* — but openwiki currently ships without the lint step. The synthesis agent updates pages locally without re-verifying the wiki's cross-reference graph, so regenerations reliably ship a class of internal-consistency drift that review bots then catch one item per round:

- **Stale counts/enumerations**: a page opening still said "the five existing dimensions" two paragraphs above freshly regenerated content documenting the sixth; a "planned but not yet authored" list contradicted by the same run's own additions.
- **Dead internal anchors**: a heading edited during regeneration left links in sibling pages pointing at the old anchor.
- **Repo-state claims that drift from the committed tree**: generated pages describing files that exist only in the working tree (e.g., the generator's own uncommitted CI scaffold), which reads as broken guidance to anyone at the committed tree.

Observed across several 0.1.2 code-mode wikis (both `--update` and `--init` paths) and equally applicable to personal-brain wikis. Still present in 0.2.0: OKF adds frontmatter validation and deterministic index generation, but no link/anchor validation of page bodies.

## Proposal

Two tiers, cheapest first:

1. **Deterministic validator (no LLM)** — after generation completes, before `.last-update.json` is written and before the generated CI workflow opens its PR: resolve every relative `[text](path.md#anchor)` link against files that exist and anchors GitHub actually generates (lowercase → strip non `[\w\- ]` → each space becomes one hyphen, so `"a + b"` → `a--b`; duplicate headings get `-1`, `-2` suffixes; root-relative `/path.md` resolves from the repo root). ~60 lines of stdlib Python/TS; fail the run or surface the list to the agent for a self-repair turn. As of 0.2.0 there's a natural home for this: the `afterAgent` hook in `src/agent/index-middleware.ts` where `synchronizeWikiIndexes` already does deterministic post-run work over the whole wiki.

2. **LLM lint pass (the missing loop step)** — a cheap final turn over the diff'd pages asking specifically for cross-page contradictions the local edits introduced: counts/enumerations that no longer match, claims about repo state, "planned" lists contradicted by shipped content. This is much cheaper as a dedicated pass than as N rounds of external review-bot findings.

Happy to contribute the tier-1 validator as a PR if there's interest — we're running exactly this as a post-generation gate in our repos' openwiki CI workflows and it took the review-round churn to zero.

[...]

PR #608, opened 2026-08-07
fix: resolve root-relative markdown links in generated wiki pages

### LINKED ISSUE #602: Generated page bodies use root-relative links (`/src/foo.ts`, `/openwiki/page.md`) that no renderer resolves — including `openwiki visualize`'s own graph
## Summary

Page bodies written by the synthesis agent link with a leading `/` — for example `[/src/main.tsx](/src/main.tsx)` and `[Viewer Runtime](/openwiki/rendering/viewer.md)` — as if the renderer will resolve `/` to the repo root. No common renderer does that, and neither does OpenWiki's own graph builder.

On a freshly generated 19-page code-mode wiki (v0.3.1), **152 of 169 links are unresolvable**. The 17 that work are exactly the relative links in the deterministically-generated `index.md` files.

This is not a rendering preference — the same convention breaks the `visualize` node graph and backlinks feature, so it is observable entirely inside OpenWiki.

## Evidence

`dist/visualize/graph.js:161` (source presumably `src/visualize/graph.ts`) resolves link targets with:

```js
const target = toId(wikiRoot, path.resolve(fileDir, link));
```

`path.resolve` **discards** `fileDir` when `link` is absolute, so `/openwiki/rendering/viewer.md` resolves to the filesystem path `/openwiki/rendering/viewer.md`. `byId.get(target)` then misses and `linkNodes` silently drops the edge ("links to unknown pages … are dropped").

Replicating that resolution over the generated wiki:

```js
// run from the repo root that openwiki was pointed at
const fs=require("fs"), path=require("path");
const wikiRoot=path.resolve("openwiki");
let total=0, resolved=0, dropped=0;
(function walk(d){for(const e of fs.readdirSync(d,{withFileTypes:true})){const p=path.join(d,e.name);
 if(e.isDirectory())walk(p); else if(e.name.endsWith(".md")){
  const body=fs.readFileSync(p,"utf8"), fileDir=path.dirname(p);
  for(const m of body.matchAll(/\]\(([^)#]+)(#[^)]*)?\)/g)){
    const link=m[1]; if(/^https?:/.test(link))continue; total++;
    const target=path.resolve(fileDir,link);          // graph.js:161
    (target.startsWith(wikiRoot)&&fs.existsSync(target))?resolved++:dropped++;
  }}}})(wikiRoot);
console.log({total, resolved, dropped});
```

```
{ total: 169, resolved: 17, dropped: 152 }
```

Link shapes emitted (counts from the same wiki):

| Prefix | Count |
|---|---|
| `/openwiki/...` (wiki-internal) | 90 |
| `/src/...` | 36 |
| `/examples/...` | 9 |
| `/CLAUDE.md` | 9 |
| `/index.html`, `/.nvmrc`, `/vite.config.ts`, `/scripts/...`, `/docs/...` | 8 |

## Where it breaks

1. **GitHub / Bitbucket blob view** — root-relative markdown links resolve against the *domain*, not the repository, so every one 404s. This is the main way a committed `openwiki/` folder gets read.
2. **Editor preview** — VS Code resolves `/x` against the open workspace folder, so links only work if the workspace happens to be the exact directory OpenWiki treated as the repo root.
3. **`openwiki visualize`** — 152/169 edges dropped, per above. The interactive graph shows ~19 near-isolated nodes and backlinks are almost entirely empty.

## Three components disagree about the convention

This looks less like one bug and more like a missing shared contract:

- **The deterministic index writer** emits **relative** links (`](overview.md)`, `](quickstart.md)`). These are the only 17 that resolve. All 6 `index.md` files use this form.
- **The synthesis agent** emits **root-absolute** links in every page body.
- **The link validator** emitted exactly **one** `<!-- openwiki: broken internal link ... -->` comment on this wiki, and it is a **false positive**: it flagged `[/examples](/examples)` as `file "/examples" does not exist` when `examples/` exists with 7 files — apparently because it checks for a file and not a directory. Meanwhile it flagged **none** of the 152 links its own graph builder cannot resolve.

So the validator and the graph builder also disagree with each other about what resolves.

## Aggravating case: `repoRoot` ≠ git root

The onboarding "Use this repository?" step allows editing the path, so `repoRoot` can legitimately be a subdirectory of the git repo (here: repo root `…/tools`, OpenWiki `repoRoot` `…/tools/a2ui-react`). In that configuration the implied link root matches *nothing* any tool derives by default — not `git rev-parse --show-toplevel`, not the editor workspace, not the forge. An agent resolving `/src/...` against the git root lands on a path that doesn't exist.

## Suggested fix

Preferred: **emit relative links from page bodies too**, matching what `index.md` already does. That is the only form that resolves in all three contexts above, needs no renderer cooperation, and makes the generator self-consistent.

If root-relative is intentional, then at minimum:
- teach `linkNodes` to resolve a leading `/` against `repoRoot` before falling back, so `visualize` works on the tool's own output; and
- fix the validator's directory check, and have it use the same resolution as the graph builder so the two cannot disagree.

Either way, a deterministic post-generation link normalization/validation step (the `afterAgent` hook alongside `synchronizeWikiIndexes` looks like the natural home) would keep the three components from drifting again.

## Not a duplicate of #372

#372 ("Post-generation consistency pass: dead internal links/anchors and cross-page drift") is adjacent but assumes this convention is correct — its proposal specifies that a validator should treat "root-relative `/path.md` resolves from the repo root." That would bless the emitted form rather than fix it.

That assumption is also self-contradictory against the shipped code. A single sentence of the proposal asks the validator to resolve *relative* links against the containing file **and** *root-relative* links against the repo root — two different roots for one link syntax — while `linkNodes` has exactly one resolver (`path.resolve(fileDir, link)`) that cannot do both, and the index writer emits only the relative form. So codifying root-relative would promote the three-way disagreement described above into a specification, and leave the graph builder non-conforming to it. Picking one root (the relative one, which already works everywhere) removes the contradiction instead of enshrining it.

This issue is about the emitted convention being unresolvable by the renderers the wiki is actually read in, including OpenWiki's own.

## Environment

- openwiki **0.3.1**, installed globally
- Node **24.16.0**, macOS (arm64)
- mode: `code`, invoked as `openwiki --init`
- provider `bedrock`, model `us.anthropic.claude-sonnet-5`
- generated wiki: 19 pages, ~11.8k words
- `repoRoot` set to a subdirectory of the git repo via the onboarding "Edit path" step

ANUBprad:
## Summary

Fix inconsistent handling of root-relative Markdown links.

### Changes

- Update the generation prompt to emit relative Markdown links instead of root-relative links.
- Preserve compatibility with previously generated root-relative links in the visualizer graph.
- Improve internal link validation for root-relative paths.
- Fix Windows CRLF bookkeeping that incorrectly reported stamped files.
- Add regression tests covering graph resolution and link validation.

Fixes #602

## Validation

- `pnpm run typecheck`
- `pnpm vitest run test/wiki-link-validator.test.ts`
- `pnpm vitest run test/wiki-link-validator-dogfood.test.ts`
- `pnpm vitest run test/visualize-graph.test.ts`

### Note

The existing Mermaid/jsdom timeout failures reproduce on upstream `main` and are unrelated to this change.

### DIFF
--- a/.changeset/curly-llamas-jump.md
+++ b/.changeset/curly-llamas-jump.md
@@ -1,0 +1,7 @@
+---
+"openwiki": patch
+---
+
+fix: emit relative markdown links so every renderer resolves them
+
+Generated page bodies used root-relative links (`/openwiki/page.md`, `/src/foo.ts`) that no renderer resolves. Page bodies are now required to use paths relative to the linking file, the validator stamps any root-relative destination so the next update repairs it, and the visualizer graph resolves leftover root-relative links instead of dropping their edges.
--- a/src/agent/prompt.ts
+++ b/src/agent/prompt.ts
@@ -115,8 +115,9 @@ Output language:
 export function createLinkIntegrityInstructions(): string {
   return `
 Link integrity:
-- Prefer relative Markdown links to existing wiki pages and stable heading anchors. Do not invent destinations that are not written in the same run.
-- OpenWiki validates relative internal links and heading anchors after the run. Broken links are left in place and marked with an HTML comment starting with "openwiki: broken internal link", so the run completes and a later update can self-correct. If you find such a comment, repair the href or restore the target page using the reason in the comment, then delete the comment.
+- Write Markdown link destinations relative to the linking file, never rooted at /. The /-rooted virtual paths in the filesystem-tool instructions are only for tool calls. Use the relative form the generated index.md files use, for example ./architecture/overview.md, ../agent/workflow.md, or ../../src/main.ts; a root-relative destination such as /openwiki/page.md or /src/main.ts resolves in no renderer (GitHub, editor preview, or the OpenWiki visualizer graph) and is treated as broken.
+- Prefer links to existing wiki pages and stable heading anchors. Do not invent destinations that are not written in the same run.
+- OpenWiki validates internal links and heading anchors after the run. Broken links — root-relative ones included — are left in place and marked with an HTML comment starting with "openwiki: broken internal link", so the run completes and a later update can self-correct. If you find such a comment, repair the href or restore the target page using the reason in the comment, then delete the comment.
 `;
 }
 
--- a/src/agent/wiki-link-validator.ts
+++ b/src/agent/wiki-link-validator.ts
@@ -124,7 +124,7 @@ export async function validateWikiInternalLinks(
 
     report.issuesFound += issues.length;
     const stamped = stampBrokenLinks(cleaned, issues);
-    if (stamped === original) {
+    if (stamped === normalizeLineEndings(original)) {
       continue;
     }
 
@@ -168,6 +168,16 @@ export function stripBrokenLinkStamps(content: string): string {
     .split(/\r?\n/u)
     .filter((line) => !BROKEN_LINK_STAMP_PATTERN.test(line))
     .join("\n");
+}
+
+/**
+ * Normalizes CRLF and lone-CR line endings to LF, so a rewrite comparison is
+ * immune to the host platform's line-ending convention. `stripBrokenLinkStamps`
+ * already returns LF, so a file whose only difference from the clean content is
+ * line endings must not be rewritten (or reported as stamped).
+ */
+function normalizeLineEndings(content: string): string {
+  return content.replace(/\r\n?/gu, "\n");
 }
 
 /**
@@ -221,6 +231,19 @@ async function validateLink(
     return null;
   }
 
+  // A leading-slash destination is root-relative: no renderer (GitHub, editor
+  // preview, or the visualizer graph) resolves it, so it cannot be relied on.
+  // Flag it so a later update run rewrites it as a path relative to this file,
+  // matching the relative form the deterministic index.md files already use.
+  if (path.posix.isAbsolute(href)) {
+    return {
+      href,
+      line,
+      message: `root-relative link "${href}" resolves in no renderer; rewrite it as a path relative to ${sourcePath}`,
+      sourcePath,
+    };
+  }
+
   const { anchor, path: linkPath } = parseLinkDestination(href);
   if (!linkPath) {
     if (!anchor) {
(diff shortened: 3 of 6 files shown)


WH-sa (approved):

maxitoon:
Rechecked `4baf17f` against current main `64164ce`. A few focused changes would help this land:

- **Preserve existing-input compatibility separately from the new emission rule.** In `src/agent/wiki-link-validator.ts`, `validateLink` now rejects every absolute href before target lookup. Main's `resolveRepoLinkPath` (lines 442–452) and `test/agent/wiki-link-validator.test.ts` (lines 54–83) still explicitly support repository-root links, as specified in #372 (now closed). Emitting relative links addresses #602, but stamping all existing `/openwiki/...` and `/src/...` links changes that compatibility contract. Please retain validation of those inputs unless maintainers explicitly agree on a migration policy.
- **Update the active repository-generation prompt.** Main's `createSystemPrompt`/`createUserPrompt` in `src/agent/prompt.ts` reject non-chat repository generation; repository page workers use `createRepositoryPagePrompt` in `src/agent/repository-prompts.ts:115–178`. Editing only `createLinkIntegrityInstructions` misses that path. Put the relative Markdown destination rule on the active page prompt and test it there.
- **Avoid guessing the virtual root from node existence.** At this PR's `src/visualize/graph.ts:307–323`, `resolveNodeId('/repo/openwiki', '/repo/openwiki', '/README.md', nodes)` returns `README` when the wiki has a `README` node. In repository mode that href names `/repo/README.md`, not `/repo/openwiki/README.md`: an incorrect edge. I reproduced this with the extracted resolver. Pass explicit root/mode semantics instead of trying wiki root first.

The current-main merge also conflicts in `src/visualize/graph.ts` and the moved `test/visualize/visualize-graph.test.ts`. Please preserve #917's single URL-decoding behavior during the update.

Minimal regression checklist: relative-link prompt assertion; existing root-relative links + anchors; slash-less directories; positive repository/local-wiki graph links; negative ambiguous `/README.md`; existing encoded-filename cases. Run the focused `test/agent/{repository-prompts,wiki-link-validator}.test.ts` and `test/visualize/visualize-graph.test.ts` suites, then format/typecheck. The recorded Format failure is still from August 7, not a current-main result.

My verification was source review plus an isolated Node resolver/static probe only—not project tests (Node 22.22+ and project dependencies unavailable). No claim of a green current-main suite.

PR #610, opened 2026-08-07
fix: preserve exec bit on dist/cli.js after build

Tomaskobel:
`prebuild` removes `dist/`, and `tsc` regenerates `cli.js` without the executable bit. For a globally linked install (`npm link`), the `bin` symlink then points at a non-executable file, so `openwiki` fails with `permission denied` after every rebuild. Contributors hit this on each edit → build → run iteration.

Adding a `postbuild` step restores the bit.

## Before

```
$ pnpm run build && ls -l dist/cli.js
-rw-r--r--  dist/cli.js
```

## After

```
$ pnpm run build && ls -l dist/cli.js
-rwxr-xr-x  dist/cli.js
```

## Checks

- `pnpm run format:check` — all files match Prettier
- `pnpm run lint:check` — clean
- `pnpm test` — 935 passed (76 files)

## Note on changesets

No changeset included: this affects the local build only, not published runtime behavior, since npm sets the exec bit on `bin` entries at install time. Happy to add one if you'd prefer.


colifran (approved):
Thanks @Tomaskobel! In the future please make sure you add a changeset using `pnpm changeset`

Issue #240, opened 2026-07-09
feat: offer provider-aware CI workflow scaffolding after --init


SadriG91:
## Problem

`--init` ends with generated docs and a pointer in `AGENTS.md`/`CLAUDE.md` — but keeping the wiki fresh requires the scheduled CI job, and today that's a fully manual last mile: find `examples/openwiki-update.yml`, copy it, replace the OpenRouter-specific env block with your provider's auth, and figure out the auth model yourself. For API-key providers that's a secret + two env vars. For cloud-credential providers it's substantially harder — e.g. Claude via Vertex AI (#179) needs `google-github-actions/auth` with workload identity federation, `id-token: write` permissions, and a service account with `roles/aiplatform.user`. That's exactly the step where setup silently stalls: the docs exist, but the automation that keeps them alive never gets wired up.

## Proposal

After a successful `--init` in an interactive terminal (before auto-exit), one final onboarding step:

```
Set up scheduled documentation updates? (writes .github/workflows/openwiki-update.yml) [Y/n]
```

On yes:

1. **Detect the forge** from `git remote get-url origin` — GitHub → `.github/workflows/openwiki-update.yml`, GitLab → print the `.gitlab-ci.yml` job to include; unknown remote → print a link to `examples/` and continue.
2. **Write the workflow pre-configured for the provider the user just set up**, not the generic template: key-based providers get an env block referencing `${{ secrets.<PROVIDER>_API_KEY }}` and the selected `OPENWIKI_MODEL_ID`; keyless/cloud providers get their auth step scaffolded (for vertex: the WIF auth action plus `OPENWIKI_PROVIDER`/`GOOGLE_CLOUD_PROJECT` env, with clearly marked placeholders for the pool/provider/service-account).
3. **Print the provider-specific follow-ups** the CLI can't do itself — "add secret `ANTHROPIC_API_KEY` in repo settings", or for vertex the three-line WIF checklist. This guidance is arguably the highest-value part even if file-writing is out of scope.

Non-interactive runs (`--print`, no TTY) never prompt — same gating as the existing credential setup. `--update` is unaffected.

## Design sketch

Per-provider CI knowledge (env block for the workflow + post-setup hint text) lives alongside `PROVIDER_CONFIGS` in `src/constants.ts`, consistent with how provider support is already centralized — every future provider then ships its CI recipe with its config entry. The two files under `examples/` become the templates this feature parameterizes.

## Minimal fallback

[...]

PR #621, opened 2026-08-09
feat: generate the ci workflow env block from the configured provider

danielsogl:
Refs #240.

## What and why

`createCodeModeWorkflow` emitted a fixed OpenRouter provider block no matter which provider the operator picked during setup. Anything other than an OpenRouter setup got a workflow whose first scheduled run failed on a secret the repo was never told about, and because the local run authenticates from `~/.openwiki/.env`, the mismatch only surfaced in CI.

The block is now derived from the resolved provider:

- the provider's credential secret, plus the paired secret and region for AWS-SDK providers
- a required endpoint or cloud project as `vars.` rather than `secrets.`, since neither is a credential
- the selected model ID, quoted (Cloudflare Workers AI IDs lead with `@`, a reserved YAML indicator that fails to parse unquoted), and omitted entirely when the provider has no preset and none is configured (Bedrock)
- for browser-login providers, a comment instead of a secret reference: the stored access token is refreshed in place, so pinning it as a repo secret breaks on the first rotation

Copilot before and after, which is what prompted this:

```yaml
# before
OPENWIKI_PROVIDER: openrouter
OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
OPENWIKI_MODEL_ID: z-ai/glm-5.2

# after
OPENWIKI_PROVIDER: copilot
COPILOT_API_KEY: ${{ secrets.COPILOT_API_KEY }}
OPENWIKI_MODEL_ID: "gpt-5.6-terra"
```

`ensureCodeModeRepoSetup` takes an optional `env` (defaulting to `process.env`) so the block is derived from the same environment the run authenticates with, and so tests do not depend on the developer's own setup. Neither call site needed changing.

## Scope

Deliberately limited to the provider half of the block. The generated workflow also hardcodes `LANGCHAIN_TRACING_V2: "true"` and the LangSmith tracing secrets even when tracing was declined at setup, which is the same class of bug but a separate change; I will open it separately rather than bundle it here. `OPENWIKI_LANGSMITH_API_KEY` is untouched on purpose, it is the connector read key rather than a tracing setting.

The `examples/` pipelines still show OpenRouter. They are hand-copied templates, not generated output, so they are out of scope here.

## How it was tested

Five tests added to `test/ingestion/code-mode.test.ts`, covering the configured-provider block, `vars.` versus `secrets.` placement, the Bedrock credential pair plus omitted model line, the browser-login comment, and model-ID quoting. The existing workflow tests still pass unchanged, including the one asserting `OPENWIKI_LANGSMITH_API_KEY` stays wired in.

Full suite: 2044 tests across 138 files pass, plus `typecheck`, `build`, `format:check` and `lint:check`. I also rendered the block for every provider category and checked the YAML by hand.


colifran (approved):
Looks good. Thanks - @danielsogl!

Issue #633, opened 2026-08-12
[Bug]: skill sync fails with EACCES when bundled skills are read-only (Nix store, containers) + misleading auth panel


giau50:
## Summary
Every `openwiki` agent run fails with:

```
Error EACCES: permission denied, unlink '.../.openwiki/skills/.write-connector-staging-XXX/staged/SKILL.md'
\# How to fix
  Your provider rejected the credentials for this run.
  1. Re-enter your key: re-run openwiki --init, or edit ~/.openwiki/.env.
```

when openwiki is installed from a location whose bundled `skills/` are read-only (e.g. the Nix store, immutable containers/images). The auth panel is a red herring.

## Root cause
`syncBundledSkills()` → `replaceSkillDirectory()` in `src/agent/skills.ts` copies bundled skills with `fs.cp(src, staged, { recursive: true })`, which preserves the source's mode bits. When the source is read-only (`dr-xr-xr-x` dirs, `-r--r--r--` files — exactly how the Nix store ships `lib/node_modules/openwiki/skills/`), the staged copy is read-only too. That breaks the atomic swap twice:

1. `rename(staged, target)` fails with `EACCES` (the staged dir has no write bit),
2. the `finally` cleanup `rm(scratchDir, { force: true, recursive: true })` fails to `unlink` `staged/SKILL.md` (unlink needs write permission on the containing dir).

Every run after the first leaves a leaked `.write-connector-staging-*/` / `.mermaid-diagrams-staging-*/` directory behind (they also can't be removed with plain `rm`).

## Minimal reproduction (Linux, Node 22)
```js
mkdir -p src/skills/write-connector && echo hi > src/skills/write-connector/SKILL.md
chmod 555 src/skills/write-connector && chmod 444 src/skills/write-connector/SKILL.md
node -e 'const {cp}=require(node:fs/promises); cp(src/skills/write-connector,out/staged,{recursive:true})'
node -e 'const {rm}=require(node:fs/promises); rm(out,{recursive:true,force:true})'
# → EACCES: permission denied, unlink '.../out/staged/SKILL.md'
```

## Misleading diagnostics
`isAuthError()` in `src/platform/diagnostics.ts` matches the message text `permission denied`, so the EACCES failure is classified as a credential rejection and the CLI shows the "re-enter your key" panel even though the provider credentials are valid.

## Suggested fix
[...]

PR #635, opened 2026-08-12
fix: sync bundled skills from read-only installations

Bubblegunn:
## Summary

- recursively add the owner-write bit to staged bundled-skill copies before the atomic replacement
- preserve all existing permission bits and skip symlinks
- add regression coverage for read-only top-level and nested directories, cleanup, and repeated syncs
- add a patch changeset

Addresses the skill-sync EACCES path in #633. The misleading authentication classification described in the same issue is intentionally excluded under the repository’s one-change-per-PR rule.

## Problem

`fs.cp(..., { recursive: true })` preserves mode bits from bundled skills. When an installation source is read-only, the staged tree remains read-only: the cross-parent directory rename can fail with `EACCES`, and the `finally` cleanup cannot unlink files from nested read-only directories. This also leaks staging directories and breaks subsequent syncs.

The issue’s suggested `chmod(staged, 0o755)` repairs the top level but does not cover nested `references/` or `scripts/` directories.

## Implementation

The staged tree is traversed before the swap and receives only the owner-write bit (`mode | 0o200`). Existing executable, group, and other bits are preserved, and symlinks are skipped so the traversal cannot modify targets outside the staging tree.

Normal writable installations remain a no-op.

## Verification

- 3 new regression tests fail on unchanged `main`
- 7 focused skill-sync tests pass with the fix
- full test suite: 2,072 tests pass
- type checks pass
- lint passes
- formatting passes

## Attribution

Thanks to @giau50 for the original report, reproduction details, and top-level workaround suggestion. This PR extends the fix to nested read-only trees and adds regression coverage for cleanup and repeated syncs.

## AI assistance

AI-assisted tooling was used during investigation and implementation. I personally reviewed the design, verified the permission semantics, and ran the reported tests.


colifran (approved):
Nice - thanks, @Bubblegunn!

PR #640, opened 2026-08-13
fix: generate CLAUDE.md as a pointer to AGENTS.md on init

### LINKED ISSUE #628: [Bug]: CLAUDE.md is not generated when the file doesn't exist at the project
### Description

When I use the `/openwiki:wiki` inside a Claude code session, it generates AGENTS.md but no CLAUDE.md is generated even though the README.md here in openwiki says that it maintains both AGENTS.md and CLAUDE.md. 

### Steps to Reproduce

1. Prepare a project with some implementations but without AGENTS.md and CLAUDE.md files
2. Run `claude` on terminal to enter a new session
3. Run `/openwiki:wiki init` inside that session
4. Verify the generated files
### Expected Behavior

The ideal behavior is to generate both CLAUDE.md and AGENTS.md files. Inside CLAUDE.md, just put reference link to AGENTS.md in order to keep it simple and manageable. In this way, users can use whatever AI tools without manually adding required files.

### Actual Behavior

It only generates AGENTS.md file.

```shell
git status                                  
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        AGENTS.md
        openwiki/

nothing added to commit but untracked files present (use "git add" to track)
```

<img width="799" height="85" alt="Image" src="https://github.com/user-attachments/assets/f4f7ef78-c3e8-4393-90f1-e243fb6e2364" /

### Environment

- OS: macOS Tahoe Version 26.5.2
- Node.js version: v22.19.0
- OpenWiki version: openwiki@openwiki-cc (v0.1.1, user)

### Additional Context

_No response_

Christian-Sidak:
## Summary

- When neither AGENTS.md nor CLAUDE.md exists, the code generates both files
- AGENTS.md receives the full agent-instruction snippet as before
- CLAUDE.md now receives a shorter snippet that simply references AGENTS.md, rather than duplicating the full content

This makes AGENTS.md the single source of truth for agent instructions while still giving Claude Code a file to read at startup. The pointer approach avoids duplication and is simpler to maintain.

## Test plan

- [x] Existing test "creates both AGENTS.md and CLAUDE.md when neither exists" still passes (both files are created with markers and `## OpenWiki`)
- [x] New test "CLAUDE.md is a simple reference to AGENTS.md" verifies CLAUDE.md contains a link to AGENTS.md and is shorter than AGENTS.md
- [x] All 2326 tests pass

Fixes #628


colifran (approved):
Thanks @Christian-Sidak!

PR #647, opened 2026-08-14
fix: refresh .last-update.json timestamp on no-op updates

IstPlayer:
## Problem

`.last-update.json` is only written when the wiki content actually changes. When a repo has no new commits since the last run, `openwiki --update` is a no-op and the timestamp stays at the last *content change*. External freshness checks (e.g. a Pi Agent startup reminder that warns "OpenWiki N days not updated") then misreport the wiki as stale even though OpenWiki just ran.

**Repro:** run `openwiki personal --update` twice in a row with no repo changes. The second run reports "no-op, nothing edited" and `.last-update.json` keeps its old `updatedAt`.

## Root cause

`persistRunMetadataIfChanged` returns early without writing metadata when the content snapshot is unchanged:

```ts
if (snapshotBefore === (await createOpenWikiContentSnapshot(cwd, outputMode))) {
  // ... returns false, no metadata write
}
```

The same applies to the fast-skip path in `runOpenWikiAgent` (`getUpdateNoopStatus().shouldSkip`), which returns before any metadata write.

## Fix

- `persistRunMetadataIfChanged` now always writes metadata for non-chat runs, so an unchanged wiki still records the actual last run time (and a completed run still clears a previous `interrupted` status).
- The fast-skip no-op path in `runOpenWikiAgent` also refreshes the metadata (best-effort, wrapped in try/catch so it never blocks the no-op path).

## Tests

Updated `test/agent/run-metadata.test.ts` to assert the new refresh behaviour instead of the previous no-write behaviour. All 403 agent tests pass; `tsc --noEmit` is clean.


colifran (approved):
Nice - thanks @IstPlayer!

Issue #651, opened 2026-08-15
Provider proposal: Claude Code CLI (subscription login, no API key)


carlitoswillis:
## Motivation

OpenWiki's providers all require an API key or cloud credentials. Many Claude users have a Claude Pro/Max subscription and the Claude Code CLI installed, but no Anthropic API account. A `claude-cli` provider lets them run OpenWiki through `claude -p`, authenticated by the CLI's own login — zero keys to paste.

## Approach

`claude -p` is one-shot text-in/text-out and cannot surface tool calls mid-turn, so the provider emulates tool use as structured output: tools are described in the prompt, a `--json-schema` forces a `{kind: "text" | "tool_calls"}` response, the harness executes the requested call, and the transcript is replayed on the next invocation.

Two things proved load-bearing in real use:

1. **Explain the indirection, don't command trust.** Capable models (Claude Sonnet 5) inspect their deliberately-empty native tool roster, conclude the session is misconfigured, and refuse to proceed. A framing that walks through *how* the harness executes their structured-output calls eliminated those refusals entirely. An explicit no-human-in-the-loop note also stops smaller models (Haiku) from ending turns to ask permission.
2. **A generous per-call timeout (600s).** Late-run turns replay the full transcript and routinely exceed 3 minutes during verification passes.

## Evidence

Used it to generate wikis for 14 repositories (~360 pages) in one day on Claude Haiku 4.5, including 4-way parallel runs. Known rough edges, disclosed up front: models occasionally end a turn with narrated intent instead of a tool call (an inherent risk of prompt-emulated tool use — retry-resume recovers it), and one repository consistently hit "No such tool available" on resumed runs (still investigating).

Draft PR with the implementation to follow. Happy to adapt the approach to maintainer preferences.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Young-Joe:
I need too

colifran:
Closed with #685

PR #652, opened 2026-08-15
feat: add claude-cli provider (Claude Code subscription auth)

carlitoswillis:
Implements the provider proposed in #651 — inference through the locally installed Claude Code CLI (`claude -p`), authenticated by the user's existing subscription login. No API key required.

## How it works

- `claude -p` can't surface tool calls mid-turn, so tool use is emulated as structured output: tools are rendered into the prompt, `--json-schema` forces a `{kind: "text" | "tool_calls"}` response, the harness executes the call, and the transcript replays on the next invocation. Prompt goes over stdin (argv limits).
- The tool framing *explains* the harness mechanism rather than instructing the model to trust it — without this, capable models find their empty native roster and refuse the session as misconfigured. Includes a no-human-in-the-loop note that stops smaller models from pausing to ask permission.
- Per-call timeout is 600s; late-run turns replay the full transcript and exceed 3 minutes during verification.
- Auth probe checks only for the `claude` binary (verifying login would cost a full round trip per setup screen); an expired login surfaces as a clear error on first generation.

## Testing

Generated wikis for 14 real repositories (~360 pages) on `claude-haiku-4-5`, including 4-way parallel runs; also exercised with `claude-sonnet-5`. Known limitations disclosed in #651.

Models offered: Opus 5 / Sonnet 5 / Haiku 4.5.

Marked draft pending maintainer feedback on the emulation approach — happy to add tests/docs in whatever shape you prefer.

### DIFF
--- a/.changeset/claude-cli-tool-refusal.md
+++ b/.changeset/claude-cli-tool-refusal.md
@@ -1,0 +1,15 @@
+---
+"openwiki": patch
+---
+
+claude-cli provider: recover when the model reports its tools are missing.
+
+`claude -p` exposes no native tools, so this provider passes OpenWiki's toolset
+in the prompt and reads tool calls back out of the structured reply. A model
+that distrusts that framing answers with prose about the missing tools instead
+of the calls that would have done the work — and because that prose is a
+well-formed `kind:"text"` reply, the agent loop accepts it as a finished turn.
+The run reports success having written nothing.
+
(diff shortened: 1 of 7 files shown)


colifran:
I think we will run into ToS issues here. Planning to pursue coding agent integrations via: https://github.com/langchain-ai/openwiki/pull/685

PR #656, opened 2026-08-15
feat: let the openai-compatible provider force the streaming transport

### LINKED ISSUE #655: [Feature]: Allow the openai-compatible provider to force the streaming transport
## Summary

Add an opt-in environment flag (`OPENWIKI_OPENAI_COMPATIBLE_STREAMING=true`) that makes the `openai-compatible` provider use the streaming transport, so OpenWiki can talk to gateways that only serve SSE.

## Motivation

Some OpenAI-compatible gateways serve only the streaming transport. A non-streaming request to them is either rejected outright (`Stream must be set to true`) or — worse — answered with HTTP 200 and an empty `content`.

OpenWiki issues non-streaming requests internally: DeepAgents' agent node calls `.invoke()`, and the streaming output visible in the TUI is LangGraph's message-level stream, not the HTTP transport. So against such a gateway every generation comes back empty, the agent writes nothing, and the run exits 0 with a blank wiki and no error message anywhere. There is no error string to search for and no diagnostic pointing at the cause, which makes this one of the harder failures to attribute — users are likely to conclude OpenWiki is broken rather than that their endpoint needs SSE.

The existing `OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API` does not help: it switches the endpoint (`/responses` instead of `/chat/completions`) but not the transport, so the non-streaming request is still non-streaming.

This is not hypothetical for the codebase either — the repo already recognises the failure mode. `createModel`'s `openai-chatgpt` branch hardcodes `streaming: true` with the comment: *"The Codex backend rejects non-streaming requests ("Stream must be set to true"), so force the streaming transport for every generation — including the non-streaming `.invoke()` calls DeepAgents' agent node issues internally."* The same remedy simply isn't reachable for users of `openai-compatible`, which is the provider most likely to point at a self-hosted or proxied endpoint.

## Proposed Solution

Add `OPENWIKI_OPENAI_COMPATIBLE_STREAMING`, defaulting to off, mirroring the existing `OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API` opt-in exactly (constant, resolver, managed-env registration, diagnostics). When enabled, `createModel` passes `streaming: true` for the `openai-compatible` provider.

It should stay **opt-in rather than becoming the default**, because `openai-compatible` points at arbitrary third-party endpoints where SSE is not guaranteed to survive proxies and load balancers. `openai-chatgpt` can hardcode it precisely because it targets one known backend. Turning it on for everyone would be a behaviour change for existing working deployments.

## Alternatives Considered

- **Defaulting `openai-compatible` to streaming.** Smallest diff, but a behaviour change for every existing user, on a provider that by definition points anywhere.
- **Detecting empty responses and retrying with streaming.** Implicit behaviour, and the heuristic is hard to pin down (an empty response is legitimately possible).
- **`OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API`.** Already exists; changes the endpoint, not the transport.

## Additional Context

Related reports from self-hosted / gateway deployments: #510 (self-hosted LLM/NIM) and #626 (self-hosted vLLM).

Two side effects worth stating up front, since enabling streaming changes more than the transport in `@langchain/openai`: the request body gains `stream_options: {include_usage: true}`, and the aggregated result reports estimated rather than server-reported token counts. Both are consequences of LangChain's streaming path, not of this flag's implementation, but they are reasons to keep it opt-in.

I have a patch ready (config + agent wiring + the generated CI workflow + tests + changeset) and am happy to open a PR if this direction looks right.

Amzp:
Closes #655.

## What

Adds `OPENWIKI_OPENAI_COMPATIBLE_STREAMING`, an opt-in boolean that makes the `openai-compatible` provider use the streaming transport. Off by default; when enabled, `createModel` passes `streaming: true`. The flag mirrors the existing `OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API` opt-in through the same constant / resolver / managed-env / diagnostics path, and is forwarded into the generated GitHub Actions workflow so a scheduled run behaves like the local one.

## Why

Some OpenAI-compatible gateways serve only the streaming transport, answering a non-streaming request with HTTP 200 and empty `content` (or rejecting it with `Stream must be set to true`). DeepAgents' agent node calls `.invoke()` internally, so against such a gateway every generation returns empty, nothing is written, and the run exits 0 with a blank wiki and no error — a silent failure with no string to search for. `OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API` doesn't help: it changes the endpoint, not the transport.

The remedy already exists in the repo — the `openai-chatgpt` branch hardcodes `streaming: true` for the Codex backend, whose comment describes this exact problem — but it isn't reachable for `openai-compatible`, the provider most likely to point at a self-hosted or proxied endpoint.

**It stays opt-in rather than becoming the default** because `openai-compatible` points at arbitrary third-party endpoints, where SSE is not guaranteed to survive proxies and load balancers; `openai-chatgpt` can hardcode it only because it targets one known backend. Enabling streaming in `@langchain/openai` also adds `stream_options: {include_usage: true}` to the request body and switches usage reporting to estimated token counts — further reasons not to impose it on working deployments. Both are noted in the README section this adds.

## How tested

- New unit tests: `providerUsesStreaming` (opt-in on/off, plus explicit negative assertions for the six other providers sharing the `ChatOpenAI` fallthrough), resolver parsing (`"true"` / `" TRUE "` / `"false"`), credential diagnostics, and that the generated workflow carries the flag only when opted in.
- Mutation-checked the new logic: reverting the injection point, the workflow forwarding, or the provider guard each makes a specific test fail — the assertions aren't vacuous.
- Verified end-to-end against a gateway exhibiting this failure mode: with the flag off, `.invoke()` returns empty content; with it on, the same request succeeds, and a full `openwiki code --init` completes with `status: "complete"` and a fully populated wiki. This exercises the multi-turn tool-calling path under SSE, not just the constructor.
- `pnpm run format`, `pnpm run lint`, `pnpm test` (2344 passed) all clean.


colifran (approved):
Nice - thanks @Amzp!

Issue #213, opened 2026-07-08
[0.0.2] Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware"


lewandowskim1988:
When I try to run `openwiki` with GLM models I get following error when running `openwiki --init`

```node
* Error Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware": expected AIMessage or Command, got object
```

My configuration is as follows
```bash
OPENAI_COMPATIBLE_API_KEY="..."
OPENAI_COMPATIBLE_BASE_URL="https://api.z.ai/api/coding/paas/v4"
OPENWIKI_PROVIDER="openai-compatible"
OPENWIKI_MODEL_ID="glm-5.2"
LANGSMITH_API_KEY=""
LANGCHAIN_TRACING_V2="false"
```

This is some misconfiguration on my side or some bug in `openwiki`?
Thanks for help!

serbinyo:
I have the same problem
```
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ >_ OpenWiki v0.0.2 agent docs for codebases                                                                                        │
│ provider: OpenAI-compatible                                                                                                        │
│ model: glm-5.2                                                                                                                     │
│ directory: D:\CodeProjectsSerbin\YandexGames\1. penguin-jumper                                                                     │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

- LangSmith tracing disabled - Run failed
Tip: ask for a docs change, or use /exit when you are done.

* Error Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware": expected AIMessage or Command, got object
# Credential Diagnostics
  Raw secret values are intentionally not printed.

  OPENWIKI_PROVIDER source=process.env over ~/.openwiki/.env
  length=17 preview="openai-compatible"
  warnings=none

  BASETEN_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  FIREWORKS_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  OPENAI_API_KEY source=unset
  length=unset preview=<unset>
  warnings=none

  OPENAI_COMPATIBLE_API_KEY source=process.env over ~/.openwiki/.env
  length=49 preview="1925e8...N4Gb"
  warnings=none

  OPENAI_COMPATIBLE_BASE_URL source=process.env over ~/.openwiki/.env
[...]

PR #660, opened 2026-08-16
fix: stream updates instead of messages for openai-compatible providers

### LINKED ISSUE #659: [Bug]: GLM (z.ai) openai-compatible runs crash with Invalid response from "wrapModelCall" — regression of #213 on 0.3.x
## Description

`openwiki --init` fails on the first model call when the provider is z.ai (GLM) via `openai-compatible`:

```
* Error Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware": expected AIMessage or Command, got object
```

This is a **regression / incomplete fix of #213** (closed as fixed by the v3 streaming protocol in #22). The failure persists on 0.3.3 because the run path never opts into the v3 handler — details below.

## Steps to Reproduce

1. Configure z.ai GLM as the provider:

```bash
OPENAI_COMPATIBLE_API_KEY="..."
OPENAI_COMPATIBLE_BASE_URL="https://api.z.ai/api/coding/paas/v4"
OPENWIKI_MODEL_ID="glm-5.3"   # glm-5.2 in the original #213 report
```

2. Run `openwiki --init` in any repository.
3. The first agent turn crashes with the error above. No wiki is produced.

## Expected Behavior

The run completes and the wiki is generated. GLM's stream shape (reasoning deltas before the first assistant delta) should not break the agent loop.

## Actual Behavior

Crash on the first model call. Root cause traced through the installed packages:

**1. `streamMode: ["messages", ...]` forces streaming on `.invoke()` internals.**
`src/agent/index.ts` (~line 529 on `main`) streams with `streamMode: ["messages", "tools"]`. The legacy `StreamMessagesHandler` (`@langchain/langgraph` `pregel/messages.js`) sets `lc_prefer_streaming = true`, which makes `@langchain/core`'s `invoke()` take the `_streamResponseChunks` aggregation path instead of `_generate` (`core/dist/language_models/chat_models.js`, the `hasStreamingHandler` branch).

**2. GLM streams `reasoning_content` deltas before the first `role: "assistant"` delta.**
In `@langchain/openai`'s `convertCompletionsDeltaToBaseMessageChunk` (`converters/completions.js`), a roleless delta falls through the `assistant` branch to the generic `ChatMessageChunk` fallback (`role = delta.role ?? defaultRole` is `undefined` on the first reasoning delta).

**3. Chunk aggregation keeps the first chunk's class.**
`concat` on message chunks preserves the first chunk's constructor. A roleless-first stream therefore aggregates to a `ChatMessageChunk`, not an `AIMessageChunk`/`AIMessage`.

Minimal repro against openwiki 0.3.3's own dependency tree (`@langchain/core` 1.2.8):

```js
import { AIMessageChunk, ChatMessageChunk, AIMessage } from "@langchain/core/messages";

// z.ai GLM stream: roleless reasoning delta arrives BEFORE the first assistant delta
const roleless = new ChatMessageChunk({ content: "", role: undefined,
  additional_kwargs: { reasoning_content: "thinking..." } });
const assistant = new AIMessageChunk({ content: "Hello" });

const glmOrder = roleless.concat(assistant);
glmOrder.constructor.name;               // => "ChatMessageChunk"
AIMessage.isInstance(glmOrder);          // => false   ← breaks the agent loop

// OpenAI/Anthropic order (assistant-role delta first) is unaffected
const openaiOrder = assistant.concat(roleless);
AIMessage.isInstance(openaiOrder);       // => true
```

Raw output from the repro run:

```
GLM order  (roleless first): ChatMessageChunk | AIMessage.isInstance: false
OpenAI order (assistant first): AIMessageChunk | AIMessage.isInstance: true
```

**4. The validator rejects the aggregated object.**
`langchain`'s `AgentNode` validates every `wrapModelCall` return with `AIMessage.isInstance(...) || isCommand(...) || {structuredResponse, messages}`. The `ChatMessageChunk` fails all three → `expected AIMessage or Command, got object`. `patchToolCallsMiddleware` is named in the error only because it is the outermost `wrapModelCall` in the deepagents stack — it passes the object through untouched; the middleware itself is not at fault.

## Environment

- OS: Linux (Ubuntu)
- Node.js version: v22.15.0
- OpenWiki version: 0.3.3 (present on `main` @ 60aada6)
- Deps at fault: `@langchain/core` 1.2.8, `@langchain/openai` 1.5.8, `langchain` 1.5.9, `@langchain/langgraph` 1.4.10, `deepagents` 1.12.0

## Additional Context

- Why the #213 fix didn't cover this path: the v3 streaming protocol handler (`StreamProtocolMessagesHandler`) is only installed when `options.version === "v3"` (langgraph `pregel/index.js` line ~1027). OpenWiki's run path calls `agent.stream(input, { streamMode: [...], subgraphs: true })` with no `version`, so it always gets the legacy handler that forces streaming. (Even the v3 handler routes through `_streamChatModelEvents`/chunk aggregation, so the class-identity bug in step 3 would still need fixing at the `@langchain/openai` or `@langchain/core` layer.)
- Community-verified workaround from #213: patch `dist/agent/index.js` (or `src/agent/index.ts`) to `streamMode: ["updates", "tools"]`. Without `"messages"`, langgraph does not force streaming, the model is called via `.invoke()` → `_generate`, and the non-streaming code path returns a proper `AIMessage`. Cost: no live token streaming in the TUI. I can confirm the workaround unblocks wiki generation on z.ai.
- Longer-term fixes could live in `@langchain/openai` (default `role: "assistant"` for chat-completions assistant streams — arguably correct for any assistant-targeted request) or `@langchain/core` (aggregation promoting to the most-specific message class), but a short-term openwiki-side mitigation is also possible (e.g. a provider/env flag to drop `"messages"` from streamMode for openai-compatible endpoints that stream reasoning first).
- Same failure reported by two users in #213 (z.ai GLM-5.2, Windows + Linux), so this is not single-machine.

JayDataEngineer:
## Summary

Fixes #659 (regression of #213). `openwiki --init` crashes on the first model call when the `openai-compatible` provider targets an endpoint that streams reasoning deltas before the first `role: "assistant"` delta (z.ai GLM via `https://api.z.ai/api/coding/paas/v4`):

```
* Error Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware": expected AIMessage or Command, got object
```

## Root cause (verified against openwiki 0.3.3's installed dependency tree)

1. `src/agent/index.ts` streams with `streamMode: ["messages", "tools"]`. The `"messages"` mode installs langgraph's legacy `StreamMessagesHandler` (`lc_prefer_streaming = true`), which routes `@langchain/core`'s internal `invoke()` through `_streamResponseChunks` chunk aggregation instead of `_generate`.
2. GLM emits `reasoning_content` deltas **before** the first `role: "assistant"` delta. A roleless delta falls through `@langchain/openai`'s converter to the generic `ChatMessageChunk` branch.
3. Chunk `concat` preserves the **first** chunk's class, so the aggregated result is a `ChatMessageChunk`, not an `AIMessage`.
4. langchain's `AgentNode` validates `wrapModelCall` returns with `AIMessage.isInstance() || isCommand() || {structuredResponse, messages}` → the `ChatMessageChunk` fails all three → the reported error. `patchToolCallsMiddleware` is only named because it's the outermost wrapper; it passes the object through untouched.

Minimal repro of step 3 (run against openwiki 0.3.3's own `node_modules`, `@langchain/core` 1.2.8):

```js
import { AIMessageChunk, ChatMessageChunk, AIMessage } from "@langchain/core/messages";

const roleless  = new ChatMessageChunk({ content: "", role: undefined,
  additional_kwargs: { reasoning_content: "thinking..." } });
const assistant = new AIMessageChunk({ content: "Hello" });

AIMessage.isInstance(roleless.concat(assistant));  // false ← GLM order
AIMessage.isInstance(assistant.concat(roleless));  // true  ← OpenAI order
```

Why the #213 fix didn't cover this: the v3 `StreamProtocolMessagesHandler` is only installed for `streamEvents(..., { version: "v3" })`. The run path calls `agent.stream()` with no version, so it always gets the legacy handler.

## Change

- `openai-compatible` runs now stream with `["updates", "tools"]` instead of `["messages", "tools"]`. Without `"messages"`, the model stays on the non-streaming `_generate` path, which returns a proper `AIMessage`.
- `OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true` opts back into live token streaming for endpoints known to emit a role-bearing first delta (mirrors the existing `OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API` opt-in convention).
- All other providers keep `messages` mode — this is scoped to the provider whose endpoint is arbitrary.
- `updates` chunks are inert for the TUI: `parseAgentStreamChunk`'s type guard only accepts `"messages" | "tools"` tuples, so they're ignored (at most 3 debug lines in `--debug`). Tool activity still streams via `"tools"` mode.

Trade-off: openai-compatible runs lose live streamed text in the TUI (the same trade-off as the community workaround in #213, which this makes official and reversible). The proper long-term fix belongs in `@langchain/openai` (default `role: "assistant"` for assistant-stream deltas) or `@langchain/core` (aggregation promoting to the most specific message class).

## Testing

- New `test/agent/stream-modes.test.ts` covers the resolver: default off, `true`/`TRUE` with whitespace on, garbage values off.
- Full suite: **2337 passed, 3 skipped, 0 failed** (`pnpm test`), `pnpm run lint` and `pnpm run format` clean.
- Manual: workaround verified end-to-end on z.ai GLM-5.3 — with `messages` dropped from streamMode the run proceeds and generates the wiki (matches the #213 report's verification on Windows/GLM-5.2).

## Notes for reviewers

One PR = one change: this is deliberately the openwiki-side mitigation only. The class-identity aggregation bug is upstream in `@langchain/openai`/`@langchain/core` and worth a separate fix there; happy to open that follow-up too.

### DIFF
--- a/.changeset/lucky-pumas-relax.md
+++ b/.changeset/lucky-pumas-relax.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Fix `openwiki --init` crashing with `Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware"` when the openai-compatible provider targets an endpoint that streams reasoning deltas before the first assistant-role delta (e.g. z.ai GLM). OpenAI-compatible runs now stream with `updates` instead of `messages` mode by default; set `OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true` to restore live token streaming on endpoints known to emit a role-bearing first delta.
--- a/.changeset/lucky-pumas-relax.md
+++ b/.changeset/lucky-pumas-relax.md
@@ -2,4 +2,4 @@
 "openwiki": patch
 ---
 
-Fix `openwiki --init` crashing with `Invalid response from "wrapModelCall" in middleware "patchToolCallsMiddleware"` when the openai-compatible provider targets an endpoint that streams reasoning deltas before the first assistant-role delta (e.g. z.ai GLM). OpenAI-compatible runs now stream with `updates` instead of `messages` mode by default; set `OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true` to restore live token streaming on endpoints known to emit a role-bearing first delta.
+fix: stream updates instead of messages for openai-compatible providers
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -116,6 +116,7 @@ import {
   providerUsesExternalCliAuth,
   providerUsesResponsesApi,
   resolveConfiguredProvider,
+  resolveOpenAiCompatibleStreamMessages,
   resolveOpenRouterMaxTokens,
   resolveOpenRouterProviderOnly,
   resolveProviderBaseUrl,
@@ -518,7 +519,21 @@ async function runOpenWikiAgentCore(
     ],
   };
 
-  emitDebug(options, "stream=opening modes=messages,tools subgraphs=true");
+  // "messages" stream mode forces @langchain/core to route the model's
+  // `.invoke()` through chunk aggregation. Providers that stream reasoning
+  // deltas before the first `role: "assistant"` delta (z.ai GLM) aggregate to
+  // a ChatMessageChunk, which the agent loop's wrapModelCall validator
+  // rejects: `expected AIMessage or Command, got object` (issue #659). The
+  // openai-compatible provider can point at any endpoint, so it gets the
+  // safe "updates" mode by default; known-good endpoints can opt back in
+  // with OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true.
+  const streamMessagesEnabled =
+    provider !== "openai-compatible" || resolveOpenAiCompatibleStreamMessages();
+  const streamModes = streamMessagesEnabled
+    ? (["messages", "tools"] as const)
+    : (["updates", "tools"] as const);
+  const streamModesLabel = streamModes.join(",");
+  emitDebug(options, `stream=opening modes=${streamModesLabel} subgraphs=true`);
   const stream = await inStage(
     "build",
     () =>
@@ -526,12 +541,12 @@ async function runOpenWikiAgentCore(
         configurable: {
           thread_id: threadId,
         },
-        streamMode: ["messages", "tools"],
+        streamMode: [...streamModes],
         subgraphs: true,
       }),
     { errorClass: "build_error", errorDetail: "stream_open" },
   );
-  emitDebug(options, "stream=started modes=messages,tools subgraphs=true");
+  emitDebug(options, `stream=started modes=${streamModesLabel} subgraphs=true`);
 
   // Register with the crash guard for exactly the stream-consumption window: a
   // subagent rejection surfaces on the microtask queue during streaming and escapes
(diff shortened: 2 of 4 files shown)


colifran (approved):
Nice - thanks, @JayDataEngineer!

PR #668, opened 2026-08-18
fix(agent): record ended_early when a clean exit leaves the skeleton behind

### LINKED ISSUE #653: `.last-update.json` records `status: "complete"` when the agent exits without finishing
## What happens

`<repo>/openwiki/.last-update.json` gets `"status": "complete"` whenever the agent's final turn ends without an error — including runs where the agent produced little or nothing. Observed repeatedly during a 14-repo batch:

- A run whose final output was a *plan* ("The skeleton is comprehensive… ready to hand off to the critic") with only `index.md` written → recorded `complete`.
- A run that ended with "Would you like me to continue?" after writing the skeleton → recorded `complete`.
- A run that *announced* "Now deleting _skeleton.md" without actually doing it → recorded `complete`, skeleton still on disk.

## Why it matters

Anything that keys on the status — resume logic, CI docs jobs, batch scripts, the `--update` no-op check's `interrupted` branch — treats these as finished wikis and skips them. In our batch we had to layer external heuristics (page count, leftover `_skeleton.md`) because the status field couldn't be trusted.

## Suggestion

Gate `complete` on verifiable state rather than a clean exit — e.g. `_skeleton.md` no longer present, or a minimum artifact check — and record something like `"status": "ended_early"` otherwise. Happy to help with a PR if maintainers agree on the direction.

Seen on v0.3.2/v0.3.3, `init` runs, multiple providers.

yzxcj797:
## Summary

Fixes #653.

`.last-update.json` recorded `status: "complete"` whenever the agent's final turn ended without an error — including runs that ended on a plan, a "would you like me to continue?", or an announced-but-not-performed cleanup. Anything keying on the status (resume logic, CI docs jobs, batch scripts, the `--update` no-op check) then treated a mostly-unwritten wiki as finished and skipped it.

## What this does

Gates `complete` on the one piece of verifiable state the run controls. The agent is instructed to delete `_skeleton.md` once every wiki file has been written, so on a clean exit a **leftover skeleton means the run ended before finishing**:

- New `UpdateRunStatus` value **`"ended_early"`** (distinct from `interrupted`, which keeps its stream/owner-failure meaning).
- The success path checks `hasLeftoverSkeletonFile(cwd, outputMode)` and records `ended_early` instead of `complete` when it's present — catching the "announced deleting `_skeleton.md` without doing it" case regardless of what the final message claimed.
- `getUpdateNoopStatus` treats `ended_early` like `interrupted`: never skip, with its own reason (`"previous run ended before finishing the wiki"`).
- `readLastUpdate` preserves the new value; metadata written before the status field existed still reads as `complete`, so upgrades don't force spurious re-runs.
- The gate emits a visible warning (run event + stderr, the same dual-channel pattern as the model-mismatch warning) naming the leftover file, so batch operators see why the wiki isn't marked complete without opening the JSON.

## Tests

- `test/agent/skeleton-status.test.ts`: skeleton detection in both output modes (repository `openwiki/_skeleton.md`, `local-wiki` root), negative case, and a metadata write/read round-trip for `ended_early`.
- `test/agent/update-noop.test.ts`: the no-op check does not skip after an `ended_early` run.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -97,6 +97,7 @@ import {
   getProviderSecretKeyEnvKey,
   getProvidersForKnownModelId,
   isModelIdForOtherProvider,
+  DEFAULT_PROVIDER_RETRY_ATTEMPTS,
   DEFAULT_VERTEX_LOCATION,
   GOOGLE_CLOUD_PROJECT_ENV_KEY,
   isValidModelId,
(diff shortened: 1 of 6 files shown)


colifran:
Thanks for investigating #653. Repository completion and interruption handling have since moved from the skeleton-based state machine to the durable page-job lifecycle introduced in #713 and hardened in #732 and #740. This implementation no longer applies to the current architecture, so I’m closing the PR. We can keep #653 open and retest it against the current lifecycle if the behavior is still reproducible.

PR #624, opened 2026-08-09
fix: close visualizer resources on Ctrl-C

HwangJohn:
## Summary

- end active live-reload SSE responses before closing the visualizer HTTP server
- close the recursive file watcher and pending debounce timer, including the initial-scan shutdown race
- add shutdown-path regression coverage and a patch changeset

## Testing

- `corepack pnpm exec vitest run test/visualize/visualize-shutdown.test.ts` (1 passed)
- `corepack pnpm exec vitest run test/visualize/server.test.ts test/visualize/visualize-shutdown.test.ts` (17 passed)
- `corepack pnpm exec vitest run test/visualize/server.test.ts test/visualize/visualize-shutdown.test.ts test/visualize/visualize-server.test.ts test/visualize/static-export.test.ts test/visualize/page.test.ts test/visualize/visualize-client-lib.test.ts` (45 passed)
- `corepack pnpm run typecheck`
- `corepack pnpm run lint:check`
- `corepack pnpm run format:check`
- manual build path: clean `dist`, run both TypeScript projects, copy visualizer assets, and chmod the CLI entrypoint
- `git diff --check`

`corepack pnpm exec vitest run test/visualize` reaches the existing local Windows symlink permission issue in `visualize-graph.test.ts:119` before that test body runs (`EPERM` creating a symlink). The shutdown regression and the other visualizer tests listed above pass locally.

Fixes #623

PR #669, opened 2026-08-18
fix(visualize): tear down SSE clients and watcher on SIGINT

### LINKED ISSUE #623: openwiki visualize can hang on Ctrl-C while a live-reload client is connected
## Bug description

`openwiki visualize` can remain running after Ctrl-C when a browser still has an open `/events` Server-Sent Events connection.

## Reproduction

1. Run `openwiki visualize`.
2. Open the visualizer in a browser and keep the page open.
3. Press Ctrl-C in the terminal.
4. The process prints `stopped.` but can remain alive.

A minimal Node.js reproduction confirms that `server.close()` does not invoke its callback while the SSE response remains open. Additional implementation discovery confirmed that the active `fs.watch()` handle also keeps the event loop alive.

## Root cause

`src/visualize/server.ts` tracks live SSE responses in `sseClients` and starts a recursive file watcher. The SIGINT handler calls `server.close()` but does not end those responses or close the watcher first. The HTTP server waits for persistent connections, while the watcher independently keeps the process alive.

## Expected behavior

Ctrl-C should end all live-reload SSE responses, close the file watcher and pending debounce timer, stop the HTTP server, and let the process exit without requiring the browser tab to close.

## Proposed scope

- End tracked SSE responses during visualizer shutdown.
- Close the file watcher and pending debounce timer.
- Prevent the watcher from starting after shutdown when the initial scan is still running.
- Keep normal client-disconnect cleanup unchanged.
- Add regression coverage for shutdown resources.
- Add a patch changeset because this changes published CLI behavior.

yzxcj797:
## Summary

Fixes #623.

## The hang

`openwiki visualize` could stay alive after Ctrl-C while a browser kept an open `/events` SSE connection:

- `server.close()`'s callback never fires while the SSE response is open, and
- the recursive `fs.watch()` handle independently keeps the event loop alive,

so the process printed `stopped.` and hung until the browser tab was closed.

## Fix (the issue's proposed scope)

- SIGINT now runs a dedicated **`shutdownVisualizer`** that, in order: **ends every tracked SSE response** (their open connections hold `server.close()` back), **closes the fs watcher with its pending debounce timer**, and only then stops the HTTP server.
- `startWatch` returns a **`WatchState` disposer** instead of `undefined`.
- A `shuttingDown` flag prevents the watcher (and the banner/browser launch) from starting after Ctrl-C while the initial scan is still running.
- Client-disconnect cleanup (`req.on("close")` → `sseClients.delete`) is unchanged.

## Tests

`test/visualize/visualize-shutdown.test.ts`:

- the shutdown sequence ends all tracked responses, empties the set, disposes the watcher, and resolves only after `server.close()`'s callback fires (ordering asserted);
- the disposer closes cleanly with no pending timer and is idempotent on double-close;
- a real server with no clients/watcher closes and stops listening through `shutdownVisualizer`.

## Changeset

Patch changeset included — this changes published CLI behavior.

### DIFF
--- a/.changeset/spotty-geckos-listen.md
+++ b/.changeset/spotty-geckos-listen.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: end SSE responses, close the file watcher, and stop the server on Ctrl-C so `openwiki visualize` cannot hang with a live-reload client connected
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -97,6 +97,7 @@ import {
   getProviderSecretKeyEnvKey,
   getProvidersForKnownModelId,
   isModelIdForOtherProvider,
+  DEFAULT_PROVIDER_RETRY_ATTEMPTS,
   DEFAULT_VERTEX_LOCATION,
   GOOGLE_CLOUD_PROJECT_ENV_KEY,
   isValidModelId,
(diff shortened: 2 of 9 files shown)


colifran:
Thanks for addressing #623. We’re consolidating the visualizer shutdown fix in #624, which is rebased, CI-clean, and scoped specifically to the visualizer resources and regression tests. Closing this PR in favor of #624.

PR #674, opened 2026-08-19
feat: add a resizable, collapsible graph panel to the visualizer

### LINKED ISSUE #658: [Feature]: Add toggle button/parameter to hide node-graph panel in UI visualizer
### Summary

Add a native option to collapse or hide the interactive node-graph visualizer panel when running the openwiki visualize command. This feature allows users to switch easily between a split-screen map view and a clean, full-width text layout.

### Motivation

The visualizer currently uses a fixed side-by-side split screen that displays the node-graph and markdown documentation at the same time. While the network graph is great for understanding page relationships, it creates a constant visual distraction when a user needs to deeply focus on reading and reviewing the actual documentation content.

There is currently no way to temporarily hide the graph canvas to create a clean, distraction-free reading environment, forcing users to look at moving visual nodes while trying to comprehend long-form text.

### Proposed Solution

We propose adding a flexible layout system with two implementation options:
1. UI Toggle Button (Frontend):Add a clickable button (e.g., "Focus Mode" or "Hide Graph") in the user interface. Clicking this button toggles a CSS class that sets display: none on the network graph container and expands the markdown documentation panel to full width for immersive reading.
2. CLI Configuration Flag (Backend):Introduce an optional command-line parameter such as openwiki visualize --text-only. Passing this flag configures the local web server to launch directly into the documentation text view without loading or rendering the graph panel by default.

### Alternatives Considered

_No response_

### Additional Context

_No response_

BenjiKo14:
## Summary

Closes #658. Adds a draggable splitter between the graph and the reader, plus
a topbar button (⛶) that collapses the graph entirely so the reader takes the
full width — matching the "UI Toggle Button" option described in the issue.
Both the splitter width and the collapsed state persist in `localStorage`
across reloads.

## Why the splitter's hit area is wider than its visible line

While building this I hit a real interaction bug worth calling out: the
splitter's line is only 2px, and a near-miss press that lands on the graph
canvas instead is read by force-graph as a background click — which clears
the currently open page (`onBackgroundClick(clearSelection)`). This only
showed up on a **stationary** press+release; a moving release is read by
force-graph as a pan instead, which is why it was intermittent and easy to
miss. Fixed by widening the splitter's hit area well past its visible line
(a ::after overlay bleeding 8px into each neighbour) and by disabling pointer
events on `#graph` for the duration of a drag as a second line of defense.

## How I tested it

- `pnpm run typecheck`, `pnpm run build`, and `pnpm run lint` all pass.
- `pnpm run coverage` fails to start in my environment
  (`ERR_PACKAGE_IMPORT_NOT_DEFINED: "#module-evaluator"` from vitest's own
  entrypoint) — I confirmed this is pre-existing and unrelated to this change
  by reproducing the identical failure on a clean, unmodified `main` checkout
  in the same environment (stashed my diff, reran, same error). I wasn't able
  to get a green vitest run locally to double-check against, so please treat
  the diff as typecheck/build/lint-verified rather than test-verified.
- Manually verified the actual interaction with Playwright driving a real
  Firefox build (not just Chromium) against the compiled output, since the
  bug above was Firefox-specific in how it first surfaced:
  - Dragging the splitter resizes the graph column and the resize survives a
    window resize (percentage-based, not raw pixels).
  - The toggle button collapses the graph to 0 width and restores it.
  - A stationary press+release near the graph/splitter boundary (the exact
    repro for the bug above) no longer clears the open page.
  - A deliberate click well inside the graph background still clears the
    open page — the pre-existing behavior is preserved, only the accidental
    trigger is fixed.

## Changeset

Added (`minor`, new backward-compatible feature).


colifran (approved):
Very nice! Thanks, @BenjiKo14!

PR #675, opened 2026-08-19
fix: omit unsupported GPT-5.6 cache retention

### LINKED ISSUE #671: [Bug]: `openai-chatgpt` provider fails with GPT-5.6 models due to unsupported `prompt_cache_retention`
### Description

When using OpenWiki with the `openai-chatgpt` provider and GPT-5.6 models, the request fails with the following error:

```text
Error 400 prompt_cache_retention is not supported on this model
```

I reproduced the same issue with both:

* `gpt-5.6-luna`
* `gpt-5.6-sol`

The important detail is that this occurs when using the **`openai-chatgpt` provider with ChatGPT OAuth authentication**.

I have not confirmed whether the same issue occurs with the regular `openai` provider using an API key.

This appears to be related to the request configuration used by the `openai-chatgpt` provider for GPT-5.6 models. In particular, `prompt_cache_retention` may be included in the request even though the target model or ChatGPT/Codex endpoint does not support that parameter.

### Steps to Reproduce

1. Configure OpenWiki to use the ChatGPT provider:

```bash
export OPENWIKI_PROVIDER=openai-chatgpt
```

2. Authenticate through the ChatGPT OAuth flow.

3. Configure GPT-5.6 Luna:

```bash
export OPENWIKI_MODEL_ID=gpt-5.6-luna
```

4. Run:

```bash
openwiki code --init
```

5. The request fails with:

```text
Error 400 prompt_cache_retention is not supported on this model
```

6. Change the model to GPT-5.6 Sol:

```bash
export OPENWIKI_MODEL_ID=gpt-5.6-sol
```

7. Run again:

```bash
openwiki code --init
```

8. The same HTTP 400 error occurs.

### Expected Behavior

OpenWiki should successfully use supported GPT-5.6 models through the `openai-chatgpt` provider.

If `prompt_cache_retention` is not supported for GPT-5.6 models through the ChatGPT/Codex request path, the provider should omit that parameter.

### Actual Behavior

Both `gpt-5.6-luna` and `gpt-5.6-sol` fail with:

```text
Error 400 prompt_cache_retention is not supported on this model
```

As a result, `openwiki code --init` cannot proceed when using these models through the `openai-chatgpt` provider.

### Environment

* OS: macOS
* Node.js version: `v24.11.1`
* OpenWiki version: `0.3.3`
* Provider: `openai-chatgpt`
* Authentication: ChatGPT OAuth
* Models tested:

  * `gpt-5.6-luna`
  * `gpt-5.6-sol`

### Additional Context

Since both GPT-5.6 Luna and GPT-5.6 Sol reproduce the same error, this may be a GPT-5.6 compatibility issue rather than a problem with one specific model.

green3sf:
Closes #671.

## What changed

- Strip the deprecated `prompt_cache_retention` field from GPT-5.6 requests at the ChatGPT/Codex fetch boundary.
- Preserve the required `store: false` behavior and Luna-specific request framing.
- Keep other models and non-Codex endpoints unchanged.
- Add regression coverage for the GPT-5.6 alias, Terra, Luna, and Sol, plus scope guards.
- Add a patch changeset.

## Why

The ChatGPT-backed Codex endpoint rejects `prompt_cache_retention` for GPT-5.6 models. LangChain can still serialize the legacy Responses option, so filtering it at the existing final request adapter fixes the HTTP 400 without disabling `zdrEnabled` or changing request behavior elsewhere.

## Validation

- `pnpm exec vitest run test/agent/openai-chatgpt-oauth.test.ts` — 40 tests passed
- `pnpm exec tsc --noEmit -p tsconfig.json`
- `pnpm exec tsc --noEmit -p tsconfig.client.json`
- `pnpm exec eslint src/agent/openai-chatgpt-oauth.ts test/agent/openai-chatgpt-oauth.test.ts`
- `pnpm exec prettier --check src/agent/openai-chatgpt-oauth.ts test/agent/openai-chatgpt-oauth.test.ts .changeset/calm-caches-rest.md`
- `git diff --check`


colifran (approved):
Looks good! Thanks, @green3sf!

green3sf:
Thank you for the review and merge! I really appreciate it.

PR #678, opened 2026-08-19
chore: serve the visualizer styles as a standalone stylesheet

### LINKED ISSUE #677: [Feature]: Extract the visualizer's inline CSS into a standalone stylesheet
### Summary

The visualizer's entire visual design lives in a 182-line `<style>` block inside a TypeScript template literal in `src/visualize/page.ts`. Moving it to `src/visualize/styles.css`, served at `/styles.css` and written as a sibling by the static export, would make it editable as CSS without changing what the page renders.

### Motivation

Three concrete problems today:

1. **No CSS tooling.** Restyling means editing a TypeScript string — no syntax highlighting, no language server, no CSS-aware formatting.
2. **Style changes are indistinguishable from logic changes in a diff**, since both live in `page.ts`.
3. **Static-export consumers can't restyle the output.** Anyone publishing with `openwiki visualize --export` has to patch OpenWiki and regenerate.

The token layer already in there (`--lc-*` primitives → semantic `--bg`/`--panel`/`--edge` → a `[data-theme="light"]` override) is well structured and would be a genuinely pleasant thing to work with as a file.

### Proposed Solution

Treat the stylesheet as a fourth browser asset, exactly like `client.js` and `client-lib.js` are handled now: one field on `VisualizerAssets`, one on `RequestHandlerDeps`, one fixed `/styles.css` route, one sibling write in the export.

Because `tsc` doesn't copy non-TypeScript files, a small cross-platform Node script copies it into `dist/` as a build step and fails the build if the result is missing or empty.

No visual change — the CSS moves verbatim.

### Alternatives Considered

Keeping the CSS in TypeScript as `export const CSS = /* css */ \`…\`` in its own module. Smaller diff, no build or server changes, and editors do highlight tagged template literals — but it stays a string, and static-export consumers still can't restyle the output. The extra plumbing for a real `.css` file seemed worth it.

### Additional Context

Happy to open a PR.

One thing worth flagging: this conflicts with #674, which adds CSS inside the same `<style>` block. I think **#674 should land first** — I've verified the rebase on top of it is clean, with their `.splitter` rules carrying into the new file unchanged.

timaxorum:
Closes #677

## What

Moves the visualizer's 182-line inline `<style>` block from `src/visualize/page.ts` into `src/visualize/styles.css`, served at `/styles.css` by the local server and written as a sibling file by the static export.

## Why

The visualizer's whole visual design currently lives inside a template literal, so any style tweak is a TypeScript change with no CSS tooling, and anyone publishing a static export can't restyle it without patching OpenWiki. See #677 for the fuller motivation.

## How it fits the existing design

The stylesheet becomes a fourth browser asset, handled exactly like the three that already exist:

- `stylesCss` field on `VisualizerAssets`, read in `loadVisualizerAssets()`
- same field on `RequestHandlerDeps`, served at a fixed `/styles.css` route with `text/css; charset=utf-8`
- one more `writeFile` in `exportStaticVisualizer()`
- `scripts/copy-visualize-assets.cjs` copies it into `dist/` (following the precedent of `scripts/stamp-build-channel.cjs`), using only `node:fs` so it works on the Windows job too, and exiting non-zero if the source is missing or the copy lands empty

No new abstraction, and routing stays a fixed allowlist — no filesystem path is derived from `req.url`.

**No visual change.** Normalising whitespace, hex case and leading zeros, the extracted CSS and the former inline block are identical, character for character.

**CSP is unchanged.** `style-src` keeps `'unsafe-inline'` — not for a `<style>` element any more, but because `client.ts` writes inline `style=` attributes for the legend swatches and sidebar dots. I updated the now-stale comment in `server.ts` that explained the old reason.

## How I tested it

- `pnpm run format`, `pnpm run lint`, `pnpm test` — 2437 passed, 3 skipped
- **New:** `test/copy-visualize-assets.test.ts` covering the copy step, matching the style of the existing `test/stamp-build-channel.test.ts`
- **Updated:** `page.test.ts` (both pages link the stylesheet; neither carries a `<style>` element), `server.test.ts` (the route serves `text/css`, and `/styles.css/../../../secret` still 404s), `static-export.test.ts` (the export writes it)
- Verified each new guard can actually fail: making the copy a no-op turns 4 of its 5 tests red with the positive control still green; pasting a `<style>` back into the template turns the no-inline guard red; widening the route match from `===` to `startsWith` turns the traversal test red
- Manually ran the server and a static export, checked both themes

## Relationship to #674

This conflicts with #674, which adds `.splitter` and `.icon-btn.active` rules inside the same `<style>` block. **I think #674 should merge first** — it was open before this and it's a user-facing feature, whereas this is a mechanical refactor.

I've already verified the rebase: one conflict in `page.ts`, resolved by re-running the extraction against #674's version of the block, with their rules carried into `styles.css` unchanged. Lint, typecheck, format, tests and build all pass on top of it. The resolved branch is here if it's useful:

https://github.com/langchain-ai/openwiki/compare/main...timaxorum:openwiki:preview/stylesheet-on-674

Happy to rebase whenever #674 lands, or to close this if you'd rather not take the refactor.


colifran (approved):
Love it. Thanks, @timaxorum!

PR #703, opened 2026-08-22
fix: isolate onboarding tests from the real ~/.openwiki on windows

### LINKED ISSUE #695: [Bug]: onboarding tests write into the real ~/.openwiki on Windows and corrupt it
### Description

`test/setup/onboarding.test.ts` isolates itself by pointing `process.env.HOME` at a temp directory, but the code it exercises resolves its home through `os.homedir()`. On Windows `os.homedir()` reads `USERPROFILE` and ignores `HOME`, so the isolation is a no-op and the tests write into the developer's real `~/.openwiki`. Nothing cleans up afterwards: `afterEach` only removes the temp directories it created.

Two files get overwritten with test fixtures. `INSTRUCTIONS.md` is replaced with whitespace and `onboarding.json` with the literal string `{ not valid json`.

The same pollution makes the file fail against itself. On a full run I get 2 failed / 40 passed, and both failures pass in isolation against a clean home directory.

Path chain at `98ccf03`:

- `test/setup/onboarding.test.ts:17` sets only `process.env.HOME`
- `src/setup/onboarding.ts:11-16` builds `openWikiOnboardingPath` and `openWikiInstructionsPath` from `openWikiHomeDir`
- `src/config/openwiki-home.ts:39` evaluates `openWikiHomeDir` at module load
- `src/config/openwiki-home.ts:11-14` falls back to `path.join(os.homedir(), ".openwiki")` when `OPENWIKI_CONFIG_DIR` is unset

### Steps to Reproduce

On Windows, at `98ccf03`:

1. Put recognizable content into `%USERPROFILE%\.openwiki\INSTRUCTIONS.md` and `%USERPROFILE%\.openwiki\onboarding.json` so the damage is visible.
2. Run the file: `npx vitest run test/setup/onboarding.test.ts`
3. Read those two files back.

Back up `~/.openwiki` first. `.env` lives in the same directory.

### Expected Behavior

The tests write only inside the temp directory created by `createTempHome`, and the real `~/.openwiki` is untouched on every platform.

### Actual Behavior

Both files are overwritten and stay overwritten after the run:

```
INSTRUCTIONS.md   "my real wiki goal"   ->  "   \n"
                  md5 311c8502cfe02637a305bf22b9666366 -> c444d514eccc3f524b8f681379056739

onboarding.json   {"real":"state"}      ->  { not valid json
                  md5 1e14ce4397eff8eb29520fa2a399957d -> f28f9ec6ff5dedf1c999d4f1bd8c2db5
```

The `INSTRUCTIONS.md` content is byte-exact for the `seedHomeInstructions(onboarding, "   \n")` call at line 815. The `onboarding.json` content comes from the `seedRawOnboardingText` calls at lines 651, 823 and 901.

Two tests fail on the full-file run:

```
x readOpenWikiOnboardingConfig missing and error paths > returns a bare empty config when neither file exists
x readOpenWikiOnboardingConfig missing and error paths > rethrows non-ENOENT errors from the instructions file
```

Both pass on their own against a clean home directory. The first is self-inflicted: earlier tests in the same file create those files in the real home, so "neither file exists" is no longer true by the time it runs.

I only caught this because my own `INSTRUCTIONS.md` turned out to have been sitting at 4 bytes of whitespace since 16 August, which is when I last ran this suite.

### Environment

- Windows 11, Node v24.12.0, pnpm 11.8.0 on PATH (repo pins `pnpm@10.33.2` via `packageManager`)
- `main` at `98ccf03`
- vitest as configured in the repo

### Additional Context

The mechanism on its own:

```
> process.env.HOME = "C:\Temp\fakehome"
> os.homedir()
'C:\Users\zelys'
```

#274 is what makes this cheap to close. It added `OPENWIKI_CONFIG_DIR`, which `resolveOpenWikiHomeDir` checks before falling back to `os.homedir()` and which behaves the same on both platforms. Setting that instead of `HOME` in `loadOnboardingModule` is a one-line change and avoids mocking `os.homedir()`. `openWikiHomeDir` is a module-load constant, so it has to be set before the dynamic import, which is already where `vi.resetModules()` sits.

Worth grepping for other tests that isolate through `process.env.HOME`, since any of them would have the same problem.

Happy to send a PR if the approach looks right.

Bubblegunn:
Fixes #695

## Summary

- point `OPENWIKI_CONFIG_DIR` at the per-test temp home in `loadOnboardingModule`, so the isolation no longer depends on `os.homedir()` following `HOME`
- restore the previous `OPENWIKI_CONFIG_DIR` value in `afterEach` alongside the existing `HOME` restore
- no changeset: the only file touched is `test/setup/onboarding.test.ts`

This is a test-infrastructure fix. No shipped runtime behaviour changes, and no source file under `src/` is modified.

## Problem

`test/setup/onboarding.test.ts` isolates itself by setting `process.env.HOME` to a temp directory. The code under test resolves its config directory through `resolveOpenWikiHomeDir`, which falls back to `path.join(os.homedir(), ".openwiki")` when `OPENWIKI_CONFIG_DIR` is unset.

On POSIX `os.homedir()` follows `$HOME`, so the isolation happens to work. On Windows it reads `USERPROFILE` and ignores `HOME`, so the isolation is a no-op and the suite writes into the developer's real config directory. Two files are overwritten and are not restored afterwards: `INSTRUCTIONS.md` and `onboarding.json`.

## Implementation

`resolveOpenWikiHomeDir` already honours `OPENWIKI_CONFIG_DIR` ahead of the `os.homedir()` fallback, so the test helper now sets that variable to the temp home as well. The isolation then holds on every platform instead of only where `os.homedir()` happens to track `HOME`. `afterEach` restores the original value, including the "was unset" case.

## Verification

Reproduced on macOS by running the existing suite with `os.homedir()` stubbed to a stand-in home directory, which reproduces the Windows resolution order:

| | before | after |
|---|---|---|
| `test/setup/onboarding.test.ts` | 2 failed / 40 passed | 42 passed |
| stand-in `INSTRUCTIONS.md` | overwritten | unchanged (md5 identical) |
| stand-in `onboarding.json` | overwritten with `{ not valid json` | unchanged (md5 identical) |

Full suite on this branch: 214 test files, 2839 tests, 0 failed. The unmodified base produces the same counts.

## Attribution

Thanks to @Zelys-DFKH for the report. The `os.homedir()` / `HOME` path chain in the issue is exactly right, and the recorded md5s made the reproduction straightforward to confirm.

## AI assistance

AI-assisted tooling was used during investigation and implementation. I personally reviewed the change, confirmed the Windows resolution order it depends on, and ran the reported tests.


colifran (approved):
Thanks, @Bubblegunn!

Issue #719, opened 2026-08-25
[Bug]: CLAUDE.md symlinked to AGENTS.md — concurrent snippet writes clobber each other and corrupt the managed block (since 0.3.3)


NeilsC:
### Description

When `CLAUDE.md` is a symlink to `AGENTS.md` — a common layout for repos that want one canonical agent-instruction file — `writeCodeModeAgentSnippets` destroys it.

`ingestion/code-mode.js` builds a *different* snippet for each name (`AGENTS.md` gets the full block, `CLAUDE.md` gets the one-line pointer added in 0.3.3) and then writes both concurrently:

```js
const updates = await Promise.all(CODE_MODE_AGENT_FILES.map((fileName) =>
  prepareCodeModeAgentSnippet(path.join(cwd, fileName), snippetByFile[fileName] ?? agentsSnippet)));
await Promise.all(updates.map(({ agentsPath, nextContent }) => writeFile(agentsPath, nextContent, "utf8")));
```

Both paths resolve to one inode. Both `writeFile` calls open with `O_TRUNC`, so the longer AGENTS.md body extends the file to its full length and the shorter CLAUDE.md body then overwrites only the front of it. What survives is the CLAUDE.md snippet followed by the tail of the AGENTS.md snippet — including a second `<!-- OPENWIKI:END -->`, and a line that starts mid-word.

Nothing detects this. Both `prepare` calls read the file *before* either write, so validation passes and the run exits 0 on a file it just corrupted. The failure only surfaces on a later invocation, when `prepareCodeModeAgentSnippet` reads back the duplicated markers and refuses to touch the file at all.

Two releases are load-bearing here:

- **0.3.0** (2026-08-04) moved the writes into `Promise.all`. Harmless at the time: both files got byte-identical content, so whichever landed second, the result was correct.
- **0.3.3** (2026-08-14) added `createCodeModeClaudeSnippet()`. The payloads now differ in length, and the race became destructive.

### Steps to Reproduce

```bash
mkdir repro && cd repro
printf '# Test\n\nSome existing instructions.\n' > AGENTS.md
ln -s AGENTS.md CLAUDE.md

cat > t.mjs <<'JS'
import { ensureCodeModeRepoSetup } from "openwiki/dist/ingestion/code-mode.js";
for (const i of [1, 2]) {
  try { await ensureCodeModeRepoSetup("."); console.log(`run ${i}: ok`); }
[...]

PR #723, opened 2026-08-26
fix: prevent corruption when agent files share a target

### LINKED ISSUE #725: Managed AGENTS.md block is corrupted when CLAUDE.md is a symlink to AGENTS.md
## Summary

In a repository where `CLAUDE.md` is a symlink to `AGENTS.md` — a common way to keep one canonical agent guide readable under both names — every `openwiki_begin` corrupts the managed block. The following run then fails outright, because the corruption it left behind is exactly what `prepareCodeModeAgentSnippet` rejects.

Reproduced on `openwiki@0.4.0` (current npm latest) and on `main` @ `27d835c`.

## Cause

`writeCodeModeAgentSnippets` treats the two entries of `CODE_MODE_AGENT_FILES` as two files:

```js
const CODE_MODE_AGENT_FILES = ["AGENTS.md", "CLAUDE.md"];   // src/ingestion/code-mode.ts

const updates = await Promise.all(          // both prepared from the same pre-write content
  CODE_MODE_AGENT_FILES.map((fileName) =>

await Promise.all(                          // both written, concurrently
  updates.map(({ agentsPath, nextContent }) => writeFile(agentsPath, nextContent, "utf8")));
```

Each name gets a **different** snippet — the full block for `AGENTS.md`, the pointer for `CLAUDE.md` — and both resolve to the same inode. The two payloads race onto one file.

## Reproduction

```bash
mkdir repro && cd repro
cat > AGENTS.md <<'MD'
# Repo guide

Our real instructions live here.
MD
ln -s AGENTS.md CLAUDE.md
# then run openwiki against this directory (any init or update run)
```

`AGENTS.md` afterwards:

```markdown
# Repo guide

Our real instructions live here.

<!-- OPENWIKI:START -->

## OpenWiki

<!-- OPENWIKI:END -->

- Treat source code and tests as authoritative. …

<!-- OPENWIKI:END -->
```

The `CLAUDE.md` pointer sits inside the block, `time context, not required startup reading.` is an orphaned tail of the `AGENTS.md` snippet, and `OPENWIKI:END` appears twice.

## Why the next run then fails

`prepareCodeModeAgentSnippet` requires exactly one ordered marker pair, so once the file is in the state above, the following run throws:

```
Cannot update AGENTS.md because its OpenWiki managed markers are malformed or
duplicated. … the file was left unchanged.
```

Setup stays broken until a human repairs the markers by hand. We hit this twice in the same repository and only got a second successful run because the file happened to be repaired in between.

## Expected

Two names that resolve to the same file should be written once. `AGENTS.md` should keep the full snippet, and the `CLAUDE.md` pointer should be skipped — the symlink already gives that host the canonical instructions.

## Notes

- `.openwikiignore` does not apply here: `src/ingestion/code-mode.ts` never consults it.
- There is no configuration for the target list; `CODE_MODE_AGENT_FILES` is a module-level constant.
- Not host-specific — `ensureCodeModeRepoSetup` runs on every `beginRepositoryRun`, for `init` and `update` alike.

PR with a fix and two regression tests follows.

Hexecu:
Follow-up to #719 after #777. Fixes #725.

The original report was about an existing `CLAUDE.md -> AGENTS.md` link. #777 now detects that case and also makes a distinct `CLAUDE.md` use the correct `@AGENTS.md` import.

There is still one bootstrap path where the two managed names are not recognized as the same target:

```text
CLAUDE.md -> AGENTS.md
AGENTS.md does not exist yet
```

`stat` cannot compare the files in that state. Setup therefore prepares the full `AGENTS.md` block and the shorter `CLAUDE.md` import from the same empty target, then writes both through the two paths concurrently.

This update resolves each managed path to a physical identity before preparing content. It follows symbolic links through a missing final target, uses device and inode for existing files and hard links, and prepares and writes each distinct target once. `AGENTS.md` remains first in the existing ordered list, so a shared target receives the full instructions. Distinct files keep the `@AGENTS.md` behavior from #777.

The same resolution also rejects a symbolic-link cycle without replacing either link. Existing validation still happens for every distinct file before any write.

### Before and after

I reran the dangling-link scenario on current `main` at `cf0700f` with Node 24.20.0 and real temporary files. On 100 fresh repositories, 75 were rejected on the second setup because the first left duplicated or malformed markers. The exact count varies with filesystem timing, which is consistent with the race.

With this update, 100 of 100 repositories completed both setup calls with one ordered marker pair, identical content through both paths and the symlink preserved.

### Tests

The focused suite now covers the dangling first-run case across repeated fresh repositories, the existing symlink behavior from #777, hard links, cycle rejection, distinct files, idempotence and malformed-marker safety. Following @vieko's suggestion, the existing code-mode suite is also included in the Windows portability job.

- Node 22.21.0: 30 focused tests passed
- Node 24.20.0: 30 focused tests passed
- Full suite: 225 files passed and 2 skipped. 3,069 tests passed and 3 skipped
- `pnpm run format:check`, `pnpm run lint:check`, `pnpm exec changeset status` and `git diff --check` passed

I have not reproduced the Windows environment locally. The added Windows CI coverage is intended to verify the same filesystem tests there before merge.

AI assisted with analysis and acted as a copilot.

### DIFF
--- a/.changeset/calm-files-share.md
+++ b/.changeset/calm-files-share.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Prevent code-mode setup from corrupting agent instructions when managed paths resolve to the same file.
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -146,7 +146,11 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Run Windows portability tests
-        run: pnpm exec vitest run test/agent/skills.test.ts test/config/env-behavior.test.ts test/agent/repository-source-fingerprint.test.ts
+        run: >-
+          pnpm exec vitest run test/agent/skills.test.ts
+          test/config/env-behavior.test.ts
+          test/agent/repository-source-fingerprint.test.ts
+          test/ingestion/code-mode.test.ts
 
   audit:
     name: Audit
--- a/src/ingestion/code-mode.ts
+++ b/src/ingestion/code-mode.ts
@@ -1,4 +1,11 @@
-import { mkdir, readFile, stat, writeFile } from "node:fs/promises";
+import {
+  lstat,
+  mkdir,
+  readFile,
+  readlink,
+  realpath,
+  writeFile,
+} from "node:fs/promises";
 import path from "node:path";
 import {
   getProviderAuthMethod,
@@ -191,27 +198,39 @@ async function readLastUpdatedAt(
 
 async function writeCodeModeAgentSnippets(cwd: string): Promise<void> {
   const agentsSnippet = createCodeModeAgentsSnippet();
-  // Some repositories make CLAUDE.md a link to AGENTS.md. There the import
-  // would point the file at itself, so the block carries the instructions
-  // instead of referring to them.
-  const claudeSnippet = (await resolvesToSameFile(
-    path.join(cwd, "AGENTS.md"),
-    path.join(cwd, "CLAUDE.md"),
-  ))
-    ? agentsSnippet
-    : createCodeModeClaudeSnippet();
   const snippetByFile: Record<string, string> = {
     "AGENTS.md": agentsSnippet,
-    "CLAUDE.md": claudeSnippet,
+    "CLAUDE.md": createCodeModeClaudeSnippet(),
   };
-  // Prepare and validate both files before writing either one. If one file has
-  // malformed markers, setup fails without partially refreshing its sibling.
+  const targets = await Promise.all(
+    CODE_MODE_AGENT_FILES.map(async (fileName) => {
+      const agentsPath = path.join(cwd, fileName);
+      return {
+        agentsPath,
+        fileIdentity: await resolveCodeModeAgentFileIdentity(agentsPath),
+        snippet: snippetByFile[fileName] ?? agentsSnippet,
+      };
+    }),
+  );
+
+  // AGENTS.md comes first intentionally. When another managed path resolves to
+  // the same file, the shared target receives the full instructions once rather
+  // than an import that would point back to itself.
+  const seenFileIdentities = new Set<string>();
+  const uniqueTargets = targets.filter(({ fileIdentity }) => {
+    if (seenFileIdentities.has(fileIdentity)) {
+      return false;
+    }
+    seenFileIdentities.add(fileIdentity);
+    return true;
+  });
+
+  // Prepare and validate every distinct file before writing any of them. If one
+  // file has malformed markers, setup fails without partially refreshing its
+  // sibling.
   const updates = await Promise.all(
-    CODE_MODE_AGENT_FILES.map((fileName) =>
-      prepareCodeModeAgentSnippet(
-        path.join(cwd, fileName),
-        snippetByFile[fileName] ?? agentsSnippet,
-      ),
+    uniqueTargets.map(({ agentsPath, snippet }) =>
+      prepareCodeModeAgentSnippet(agentsPath, snippet),
     ),
   );
 
@@ -222,6 +241,57 @@ async function writeCodeModeAgentSnippets(cwd: string): Promise<void> {
         : writeFile(agentsPath, nextContent, "utf8"),
     ),
   );
+}
+
+/**
+ * Resolve the physical identity of a managed agent file, including a symbolic
+ * link whose final target has not been created yet.
+ */
+async function resolveCodeModeAgentFileIdentity(
+  agentsPath: string,
+  visitedPaths = new Set<string>(),
+): Promise<string> {
+  const absolutePath = path.resolve(agentsPath);
+
+  try {
+    const fileStats = await lstat(absolutePath, { bigint: true });
+    if (fileStats.isSymbolicLink()) {
+      if (visitedPaths.has(absolutePath)) {
+        throw new Error(
+          `Cannot update ${path.basename(agentsPath)} because its symbolic link chain contains a cycle.`,
+        );
+      }
+      visitedPaths.add(absolutePath);
+      const target = await readlink(absolutePath);
+      return resolveCodeModeAgentFileIdentity(
+        path.resolve(path.dirname(absolutePath), target),
+        visitedPaths,
+      );
+    }
+
+    // Device and inode identify regular paths and hard links. Some filesystems
+    // report no useful inode, so use the canonical path as a safe fallback.
+    return fileStats.ino === 0n
+      ? `path:${await realpath(absolutePath)}`
+      : `inode:${fileStats.dev}:${fileStats.ino}`;
+  } catch (error) {
+    if (!isFileNotFoundError(error)) {
+      throw error;
+    }
+
+    // A missing target has no inode yet. Its canonical parent still lets a
+    // dangling link and the path it names share one stable identity.
+    let canonicalParent: string;
+    try {
+      canonicalParent = await realpath(path.dirname(absolutePath));
+    } catch (parentError) {
+      if (!isFileNotFoundError(parentError)) {
+        throw parentError;
+      }
+      canonicalParent = path.resolve(path.dirname(absolutePath));
+    }
+    return `path:${path.join(canonicalParent, path.basename(absolutePath))}`;
+  }
 }
 
 async function prepareCodeModeAgentSnippet(
@@ -454,21 +524,4 @@ function createCodeModeClaudeSnippet(): string {
 @AGENTS.md
 
 ${OPENWIKI_AGENTS_SNIPPET_END}`;
-}
-
-/**
- * Whether two paths are the same file on disk, following symlinks. Inode 0 is
- * treated as unknown because some Windows filesystems report it for every
- * file, which would otherwise make unrelated paths compare equal.
- */
-async function resolvesToSameFile(
-  first: string,
-  second: string,
-): Promise<boolean> {
-  try {
-    const [a, b] = await Promise.all([stat(first), stat(second)]);
-    return a.ino !== 0 && a.ino === b.ino && a.dev === b.dev;
-  } catch {
-    return false;
-  }
 }
--- a/test/ingestion/code-mode.test.ts
+++ b/test/ingestion/code-mode.test.ts
@@ -1,7 +1,10 @@
 import {
+  lstat,
+  link,
   mkdir,
   mkdtemp,
   readFile,
+  readlink,
   rm,
   symlink,
   writeFile,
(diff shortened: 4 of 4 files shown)


vieko:
@Hexecu, I have a tests-only patch you’re welcome to fold into this PR. It adds automated coverage around your existing fix rather than another implementation.

Five additional cases cover:
- Reverse `AGENTS.md → CLAUDE.md` aliases.
- Two relative symlink chains sharing a third policy file.
- A differently cased regular-file path on a case-insensitive volume (explicitly skipped elsewhere).
- Exact CRLF/trailing-space preservation around an existing managed block.
- Init → finish → update → interrupted resume → finish, with `CLAUDE.md` still linked to `AGENTS.md`.

It also adds the setup tests and targeted lifecycle case to the existing Windows CI job. Windows execution is **not yet verified**; my runs were on macOS. I saw your manual topology probes, so this is an offer to keep more of that coverage in the suite, not a claim those layouts went untested.

No source, dependency, changeset or supported-layout changes. Your hard-link and dangling-target behavior stays intact. The patch is based on this PR’s current head, `d11bee2a431413b369858eb00d5487b23c16d048`, and is three files, `+261/-2`.

Validation on that head plus the patch:
- Full Node 24.20.0 suite: **3,048 passed, 3 existing skips**. Used npm 10.9.4 and `--coverage --maxWorkers=2 --testTimeout=20000 --hookTimeout=30000` for the local run.
- Node 22.21.0: all 34 code-mode tests plus the new lifecycle case passed.
- Build, source/client and scoped test typechecks, full-repository lint and formatting passed.
- On unchanged upstream main (`1e6d54c`), four of the five new cases failed in the comparison run. The chain case happened to pass, consistent with the original timing-dependent race; all five pass here.

This was AI-assisted and independently reviewed/tested locally. The apply-ready diff is below, so there’s no dependency on an unpublished branch. Save the code block as `alias-coverage.patch`, then use `git apply --check alias-coverage.patch` followed by `git apply alias-coverage.patch` on this PR’s head. Take whichever cases fit; I’m keeping the broader write-safety work separate.

<details>
<summary>Apply-ready tests and CI diff</summary>

```diff
diff --git a/.github/workflows/checks.yml b/.github/workflows/checks.yml
index 690f656..2c39f5f 100644
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -146,7 +146,18 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Run Windows portability tests
-        run: pnpm exec vitest run test/agent/skills.test.ts test/config/env-behavior.test.ts test/agent/repository-source-fingerprint.test.ts
+        run: >-
+          pnpm exec vitest run test/agent/skills.test.ts
+          test/config/env-behavior.test.ts
+          test/agent/repository-source-fingerprint.test.ts
+          test/ingestion/code-mode.test.ts
+          test/ingestion/code-mode-aliases.test.ts
+
+      - name: Check aliased instructions across the repository lifecycle
+        run: >-
+          pnpm exec vitest run test/generation/repository-run.test.ts
+          -t "keeps CLAUDE.md linked to AGENTS.md through init, update, and resume"
+          --testTimeout=20000
 
   audit:
     name: Audit
diff --git a/test/generation/repository-run.test.ts b/test/generation/repository-run.test.ts
index 215ff8a..9db08a5 100644
--- a/test/generation/repository-run.test.ts
+++ b/test/generation/repository-run.test.ts
@@ -1,6 +1,15 @@
 import { execFile } from "node:child_process";
 import { randomUUID } from "node:crypto";
-import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import {
+  lstat,
+  mkdir,
+  mkdtemp,
+  readFile,
+  readlink,
+  rm,
+  symlink,
+  writeFile,
+} from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import { promisify } from "node:util";
@@ -1528,6 +1537,81 @@ describe("finishRepositoryRun", () => {
     });
   });
 
+  test("keeps CLAUDE.md linked to AGENTS.md through init, update, and resume", async () => {
+    const root = await createRepository();
+    const agents = path.join(root, "AGENTS.md");
+    const claude = path.join(root, "CLAUDE.md");
+    const humanPolicy = "# Human policy\nKeep this guidance.";
+    await writeFile(agents, humanPolicy, "utf8");
+    await rm(claude);
+    await symlink("AGENTS.md", claude);
+    const run = requireActiveRun(
+      await beginRepositoryRun({
+        root,
+        mode: "init",
+        actor: ACTOR,
+        now: () => new Date(STARTED_AT),
+      }),
+    );
+    await submitRepositoryPlan(run, {
+      pages: [
+        {
+          path: "/openwiki/quickstart.md",
+          title: "Quickstart",
+          purpose: "Document the repository entry point.",
+        },
+      ],
+    });
+    await completeCurrentPage(run, "Quickstart");
+    await finishRepositoryRun(run);
+    const canonical = await readFile(agents, "utf8");
+    expect(canonical.startsWith(humanPolicy)).toBe(true);
+    expect(canonical).toContain(
+      "Treat source code and tests as authoritative.",
+    );
+    expect(canonical).not.toContain("See [AGENTS.md](AGENTS.md)");
+    expect(canonical.split("<!-- OPENWIKI:START -->")).toHaveLength(2);
+    expect(canonical.split("<!-- OPENWIKI:END -->")).toHaveLength(2);
+    expect((await lstat(claude)).isSymbolicLink()).toBe(true);
+    expect(await readlink(claude)).toBe("AGENTS.md");
+
+    const update = await beginForcedUpdate(root);
+    expect(await readFile(agents, "utf8")).toBe(canonical);
+    const runId = update.state.runId;
+    await submitRepositoryPlan(update, {
+      pages: [
+        {
+          path: "/openwiki/quickstart.md",
+          title: "Quickstart",
+          purpose: "Refresh the repository entry point.",
+        },
+      ],
+    });
+    const pending = await nextRepositoryPage(update);
+    if (pending.status !== "pending") throw new Error("Expected pending page.");
+    const snapshot = await captureRepositoryPageSnapshot(
+      update,
+      pending.job.id,
+    );
+    await skipRepositoryPage(update, snapshot);
+    const resumedResult = await beginRepositoryRun({
+      root,
+      mode: "update",
+      actor: OTHER_ACTOR,
+    });
+    const resumed = requireActiveRun(resumedResult);
+    expect(resumed.state.runId).toBe(runId);
+    await completeCurrentPage(resumed, "Quickstart");
+    await finishRepositoryRun(resumed);
+
+    expect(await readFile(path.join(root, "AGENTS.md"), "utf8")).toBe(
+      canonical,
+    );
+    expect(await readFile(claude, "utf8")).toBe(canonical);
+    expect((await lstat(claude)).isSymbolicLink()).toBe(true);
+    expect(await readlink(claude)).toBe("AGENTS.md");
+  }, 20_000);
+
   test("completes init with canonical bytes and final Claims durable", async () => {
     const root = await createRepository(["old.md"]);
     const result = await beginRepositoryRun({
diff --git a/test/ingestion/code-mode-aliases.test.ts b/test/ingestion/code-mode-aliases.test.ts
new file mode 100644
index 0000000..1a2dde3
--- /dev/null
+++ b/test/ingestion/code-mode-aliases.test.ts
@@ -0,0 +1,164 @@
+import {
+  lstat,
+  mkdir,
+  mkdtemp,
+  readFile,
+  readlink,
+  rm,
+  symlink,
+  writeFile,
+} from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, test } from "vitest";
+import { ensureCodeModeRepoSetup } from "../../src/ingestion/code-mode.ts";
+
+const START = "<!-- OPENWIKI:START -->";
+const END = "<!-- OPENWIKI:END -->";
+const repos: string[] = [];
+
+async function repo(): Promise<string> {
+  const root = await mkdtemp(
+    path.join(tmpdir(), "openwiki-code-mode-aliases-"),
+  );
+  repos.push(root);
+  return root;
+}
+
+afterEach(async () => {
+  await Promise.all(
+    repos.splice(0).map((root) => rm(root, { recursive: true, force: true })),
+  );
+});
+
+function expectCanonicalBlock(content: string): void {
+  expect(content.split(START)).toHaveLength(2);
+  expect(content.split(END)).toHaveLength(2);
+  expect(content.indexOf(START)).toBeLessThan(content.indexOf(END));
+  expect(content).toContain("Treat source code and tests as authoritative.");
+  expect(content).not.toContain("See [AGENTS.md](AGENTS.md)");
+}
+
+describe("code-mode agent aliases", () => {
+  test("uses the full canonical block for a reverse AGENTS.md alias", async () => {
+    const root = await repo();
+    const agents = path.join(root, "AGENTS.md");
+    const claude = path.join(root, "CLAUDE.md");
+    await writeFile(claude, "# Human policy\n\nKeep this guidance.\n", "utf8");
+    await symlink("CLAUDE.md", agents);
+
+    await ensureCodeModeRepoSetup(root);
+    const first = await readFile(claude, "utf8");
+    await ensureCodeModeRepoSetup(root);
+    const second = await readFile(claude, "utf8");
+
+    expect(first).toBe(second);
+    expect(await readFile(agents, "utf8")).toBe(first);
+    expectCanonicalBlock(first);
+    expect((await lstat(agents)).isSymbolicLink()).toBe(true);
+    expect(await readlink(agents)).toBe("CLAUDE.md");
+  });
+
+  test("refreshes a multi-hop relative chain without replacing either alias", async () => {
+    const root = await repo();
+    const policy = path.join(root, "policy.md");
+    const links = path.join(root, "links");
+    await writeFile(policy, "# Existing policy\n", "utf8");
+    await mkdir(links);
+    await symlink(
+      path.join("..", "policy.md"),
+      path.join(links, "agents-hop.md"),
+    );
+    await symlink(
+      path.join("..", "policy.md"),
+      path.join(links, "claude-hop.md"),
+    );
+    await symlink(
+      path.join("links", "agents-hop.md"),
+      path.join(root, "AGENTS.md"),
+    );
+    await symlink(
+      path.join("links", "claude-hop.md"),
+      path.join(root, "CLAUDE.md"),
+    );
+    const linksBefore = await Promise.all([
+      readlink(path.join(root, "AGENTS.md")),
+      readlink(path.join(root, "CLAUDE.md")),
+      readlink(path.join(links, "agents-hop.md")),
+      readlink(path.join(links, "claude-hop.md")),
+    ]);
+
+    await ensureCodeModeRepoSetup(root);
+    const first = await readFile(policy, "utf8");
+    await ensureCodeModeRepoSetup(root);
+    const second = await readFile(policy, "utf8");
+
+    expect(second).toBe(first);
+    expect(await readFile(path.join(root, "AGENTS.md"), "utf8")).toBe(first);
+    expect(await readFile(path.join(root, "CLAUDE.md"), "utf8")).toBe(first);
+    expect(
+      await Promise.all([
+        readlink(path.join(root, "AGENTS.md")),
+        readlink(path.join(root, "CLAUDE.md")),
+        readlink(path.join(links, "agents-hop.md")),
+        readlink(path.join(links, "claude-hop.md")),
+      ]),
+    ).toEqual(linksBefore);
+    expectCanonicalBlock(first);
+  });
+
+  test("resolves a lower-case target on a case-insensitive volume", async (context) => {
+    const root = await repo();
+    await writeFile(
+      path.join(root, "agents.md"),
+      "# Existing policy\n",
+      "utf8",
+    );
+    // Exercise a regular AGENTS.md entry with different on-disk casing, not
+    // a second symlink that would also work on a case-sensitive volume.
+    try {
+      await lstat(path.join(root, "AGENTS.md"));
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code !== "ENOENT") throw error;
+      context.skip();
+      return;
+    }
+    expect((await lstat(path.join(root, "AGENTS.md"))).isFile()).toBe(true);
+    await symlink("agents.md", path.join(root, "CLAUDE.md"));
+
+    await ensureCodeModeRepoSetup(root);
+    const content = await readFile(path.join(root, "agents.md"), "utf8");
+    await ensureCodeModeRepoSetup(root);
+
+    expectCanonicalBlock(content);
+    expect(await readFile(path.join(root, "AGENTS.md"), "utf8")).toBe(content);
+    expect(await readFile(path.join(root, "CLAUDE.md"), "utf8")).toBe(content);
+    expect(await readlink(path.join(root, "CLAUDE.md"))).toBe("agents.md");
+  });
+
+  test("refreshes managed bytes while preserving CRLF and trailing-space human text", async () => {
+    const root = await repo();
+    const file = path.join(root, "AGENTS.md");
+    const existing = `Human prefix with spaces  \r\n\r\n${START}\r\nstale\r\n${END}\r\n\r\nHuman suffix with spaces  \r\n`;
+    await writeFile(file, existing, "utf8");
+    const claude = path.join(root, "CLAUDE.md");
+    await symlink("AGENTS.md", claude);
+    const beforePrefix = existing.slice(0, existing.indexOf(START));
+    const beforeSuffix = existing.slice(existing.indexOf(END) + END.length);
+
+    await ensureCodeModeRepoSetup(root);
+    const first = await readFile(file, "utf8");
+    await ensureCodeModeRepoSetup(root);
+    const second = await readFile(file, "utf8");
+
+    expect(second).toBe(first);
+    const start = first.indexOf(START);
+    const end = first.indexOf(END);
+    expect(first.slice(0, start)).toBe(beforePrefix);
+    expect(first.slice(end + END.length)).toBe(beforeSuffix);
+    expectCanonicalBlock(first);
+    expect(first).not.toContain("stale");
+    expect(await readFile(claude, "utf8")).toBe(first);
+    expect(await readlink(claude)).toBe("AGENTS.md");
+  });
+});
```

</details>

Hexecu:
Thanks @colifran for the note on #726, and @danielsogl for #777. I checked this again against current `main` before rebasing anything.

#777 now handles the original case where `CLAUDE.md` points to an existing `AGENTS.md`, so I agree that #723 should no longer present that case as unresolved.

I did find one remaining corruption path:

```text
CLAUDE.md -> AGENTS.md
AGENTS.md does not exist before the first code-mode setup
```

In this state the `stat` comparison cannot identify the shared target, so the full `AGENTS.md` block and the `@AGENTS.md` block are still prepared and written concurrently through the two paths.

Using real temporary files on current `main`, with Node 24.20.0 on macOS, 87 of 100 fresh repositories were left with duplicated or malformed markers and were rejected on the second setup. The current #723 head remained stable in 100 of 100 runs. This is not Windows verification, but it is a reproducible bootstrap state that Git can preserve.

Would you be open to me rebasing and narrowing #723 to this remaining case, while preserving the `@AGENTS.md` behavior introduced by #777? If this topology is not considered supported, I am also happy to close it rather than retain unnecessary filesystem complexity.

Hexecu:
I’ve rebased and narrowed the PR to the remaining dangling-symlink bootstrap case, while preserving #777’s AGENTS.md behavior. The updated branch includes focused regression coverage and adds the code-mode suite to Windows CI.

PR #726, opened 2026-08-26
fix: dedupe managed agent files by resolved path

### LINKED ISSUE #725: Managed AGENTS.md block is corrupted when CLAUDE.md is a symlink to AGENTS.md

leonardsellem:
Fixes #725.

## What

`writeCodeModeAgentSnippets` treats the two entries of `CODE_MODE_AGENT_FILES` as two files. When `CLAUDE.md` is a symlink to `AGENTS.md` they are one file, and the two different snippets race onto a single inode:

```js
const updates = await Promise.all(   // both prepared from the same pre-write content
  CODE_MODE_AGENT_FILES.map((fileName) => prepareCodeModeAgentSnippet(...)));

await Promise.all(                   // both written, concurrently
  updates.map(({ agentsPath, nextContent }) => writeFile(agentsPath, nextContent, "utf8")));
```

This collapses the target list by resolved path before any update is prepared, keeping the first name so `AGENTS.md` wins and the `CLAUDE.md` pointer is skipped — the symlink already gives that host the canonical instructions. A name that does not resolve stays its own target, since it will be created at exactly that path; anything other than `ENOENT` still propagates.

## Why

Symlinking `CLAUDE.md` to `AGENTS.md` is a common way to keep one canonical guide. Today every run in such a repo leaves a block with the pointer text inside it, an orphaned snippet tail, and a duplicated `OPENWIKI:END`.

That is not only cosmetic. `prepareCodeModeAgentSnippet` rejects malformed or duplicated markers, so the **next** run fails outright on a file the previous run corrupted:

```
Cannot update AGENTS.md because its OpenWiki managed markers are malformed or
duplicated. … the file was left unchanged.
```

Setup then stays broken until someone repairs the markers by hand. #725 has the full reproduction and the resulting file.

## How it was tested

Two tests added to `test/ingestion/code-mode.test.ts`, both verified failing on `main` before the fix and passing after:

- *keeps a CLAUDE.md symlinked to AGENTS.md intact and well-formed* — asserts one marker pair, the preserved hand-written policy, no pointer text in `AGENTS.md`, and both names reading identical content. Fails before the fix with `expected 3 to be 2` on the `OPENWIKI:END` count.
- *stays idempotent across runs when CLAUDE.md is a symlink to AGENTS.md* — runs setup twice and compares. Fails before the fix with the malformed-markers error, which is the follow-on breakage above.

`pnpm run format`, `pnpm run lint`, and `pnpm test` all pass: 2908 tests (2906 on `main` plus these 2), 219 files.

One caveat, reported rather than hidden: on the first full run `test/cli/components/chat.test.tsx` failed once on the reasoning-effort assertion. It passes in isolation on both this branch and clean `main`, and passes on a re-run of the full suite here; a clean-`main` full run is also green. It looks like a timing flake in that Ink rendering test, unrelated to this change, but flagging it in case it is known.

## Scope

One change: the duplicate-target bug, its regression tests, and a patch changeset. No behavior change for repos where `AGENTS.md` and `CLAUDE.md` are distinct files — they still receive their existing separate snippets.

### DIFF
--- a/.changeset/symlinked-claude-md-managed-block.md
+++ b/.changeset/symlinked-claude-md-managed-block.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: stop corrupting the managed block when CLAUDE.md is a symlink to AGENTS.md
--- a/src/ingestion/code-mode.ts
+++ b/src/ingestion/code-mode.ts
@@ -1,4 +1,4 @@
-import { mkdir, readFile, writeFile } from "node:fs/promises";
+import { mkdir, readFile, realpath, writeFile } from "node:fs/promises";
 import path from "node:path";
 import {
   getProviderAuthMethod,
@@ -195,10 +195,12 @@ async function writeCodeModeAgentSnippets(cwd: string): Promise<void> {
     "AGENTS.md": agentsSnippet,
     "CLAUDE.md": claudeSnippet,
   };
+  const targetFiles = await resolveDistinctAgentFiles(cwd);
+
   // Prepare and validate both files before writing either one. If one file has
   // malformed markers, setup fails without partially refreshing its sibling.
   const updates = await Promise.all(
-    CODE_MODE_AGENT_FILES.map((fileName) =>
+    targetFiles.map((fileName) =>
       prepareCodeModeAgentSnippet(
         path.join(cwd, fileName),
         snippetByFile[fileName] ?? agentsSnippet,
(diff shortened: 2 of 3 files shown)


colifran:
Thanks for addressing the symlink corruption bug. #723 now covers the same case plus dangling symlinks, symlink chains, hard links, cycles, malformed shared targets, and repeated-run behavior. We’re consolidating the fix there, so I’m closing this PR in favor of #723.

PR #743, opened 2026-08-27
fix(bedrock): set maxTokens to avoid Converse API 4096-token cap

### LINKED ISSUE #722: Bedrock provider sets no maxTokens → Converse caps output at 4096, truncating pages mid-write
### Summary

On the `bedrock` provider, OpenWiki builds the model with no `maxTokens`, so the Bedrock **Converse API caps output at its 4096-token default**. A from-scratch `--init` on a non-trivial repo then reliably writes only ~4 short pages and stops mid-sentence, because a turn that hits the cap truncates **mid-`write_file` tool call** and the agent graph ends cleanly (exit 0). It looks like a page-count/runtime limit but is purely the output-token cap.

### Version

- `openwiki@0.2.5`, Node 22
- Provider: `bedrock`, model `us.anthropic.claude-sonnet-5`, region `us-east-1`

### Where

`dist/agent/index.js`, `createModel()`:

```js
if (provider === "bedrock") {
        model: modelId,
        ...retryOptions,
    });
}
```

No `maxTokens` is passed. `@langchain/aws`'s `ChatBedrockConverse` leaves `maxTokens` undefined and only attaches an `inferenceConfig` when one of maxTokens/temperature/topP/stop is set — so it sends none, and the Bedrock Converse API applies its **4096-token default**. Every other provider path is fine: `@langchain/anthropic` defaults these models to 16384 via `defaultMaxOutputTokensForModel`; only the Bedrock path passes nothing.

### Proof

Raw Bedrock Converse call to `us.anthropic.claude-sonnet-5` with no inference config, asking for a long answer:

```
stopReason: "max_tokens"
outputTokens: 4096
```

Setting `inferenceConfig.maxTokens` higher removes the cap.

### Symptoms observed

- `--init` on a medium repo writes ~4 short pages (quickstart + a few) then exits 0, mid-sentence.
- A run given a steering `[message]` spends the 4096 on upfront reasoning and writes **0 files**.
- Not a step ceiling: deepagents sets `recursionLimit: 1e4`; and the "at most 8 documentation pages" prompt line is soft guidance, not the limiter.

### Suggested fix

Set an explicit `maxTokens` on the Bedrock model (e.g. 16000, matching `@langchain/anthropic`'s built-in default for these models), or expose it via an env/config knob. Injecting `maxTokens: 16000` into that constructor took a real repo from 4 → 8 complete pages with no other change.

Christian-Sidak:
## Summary

- The Bedrock provider passed no `maxTokens` to `ChatBedrockConverse`, so the Converse API applied its 4096-token default ceiling
- Long wiki pages were silently truncated mid-write with `stopReason: max_tokens` and exit 0, looking like a page-count or runtime limit
- Sets a default of 16000 (matching `@langchain/anthropic`'s built-in ceiling for Claude models)
- Adds `OPENWIKI_BEDROCK_MAX_TOKENS` env var override for models with a lower ceiling (Haiku 3, Titan/Llama variants)

Fixes #722

## Test plan

- [x] `resolveBedrockMaxTokens` unit tests: default returns 16000, env override is parsed, invalid values throw with a clear error message
- [x] `createModel` Bedrock integration tests: default maxTokens is 16000, env override passes through, invalid value throws
- [x] All existing Bedrock credential tests still pass
- [x] Full test suite: 81 tests pass across affected files


colifran (dismissed):
Nice. Thanks, @Christian-Sidak!

colifran (approved):

PR #745, opened 2026-08-27
fix: default Bedrock Claude runs to a 16,384-token output ceiling

### LINKED ISSUE #722: Bedrock provider sets no maxTokens → Converse caps output at 4096, truncating pages mid-write

Payal2000:
## What and why

On the `bedrock` provider a default run sends no `inferenceConfig` at all, so the Converse API applies its own **4,096-token** output ceiling. A turn that hits the cap truncates mid-`write_file` and the agent graph ends cleanly (exit 0), which reads as a page-count or runtime limit rather than an output-token cap — `--init` on a medium repo writes ~4 short pages and stops mid-sentence.

`ChatBedrockConverse` attaches an `inferenceConfig` only when one of `maxTokens`/`temperature`/`topP`/`stop` is set. Since #459, OpenWiki does pass `maxTokens` on the Bedrock path — but only when `OPENWIKI_MAX_OUTPUT_TOKENS` or an explicit run option is configured. Out of the box nothing is, so nothing is sent. The direct Anthropic path has no such gap: `resolveAnthropicMaxOutputTokens` gives modern Claude families 16,384 by default. `us.anthropic.claude-sonnet-5` over Bedrock gets nothing.

This mirrors that helper on the Bedrock path, so the same Claude model gets the same ceiling however it is reached.

## Why the ceiling is model-derived rather than flat

Bedrock serves every vendor through this one client, and a ceiling above a model's own maximum is a `ValidationException`. So the default is raised only for the modern Anthropic families known to accept it — non-Anthropic vendors (Nova, Llama, Mistral), Claude 3.x (`claude-3-5-sonnet-…`, `claude-3-haiku-…`), and custom IDs keep the provider default they work with today. Bare, cross-region inference-profile (`us.`/`eu.`/`apac.`/`us-gov.`), and ARN-qualified IDs all normalize to the same decision, so the ceiling doesn't depend on which form the operator pasted.

Configuration is unchanged: `OPENWIKI_MAX_OUTPUT_TOKENS` (and an explicit run option) still overrides, including for models with no known ceiling. No new env var — the provider-neutral one from #459 already reaches this path.

## Relationship to #743

#743 targets the same issue. It was written against a pre-#459 tree — its base Bedrock constructor has neither `maxTokensOptions` nor `streamIdleTimeoutOptions` — and currently conflicts. Against today's `main` it would add `OPENWIKI_BEDROCK_MAX_TOKENS` as a third max-tokens env var alongside the provider-neutral one, reimplement `resolvePositiveIntegerSetting`, and apply a flat `16000` to every Bedrock model including those with a lower ceiling. This PR is an alternative built on current `main`; happy to defer if maintainers prefer that direction instead. Credit to @Christian-Sidak for getting to it first, and to @tshappee-veeva for the root-cause writeup in #722 — the `stopReason: max_tokens` / `outputTokens: 4096` proof is what makes this diagnosable at all.

## How I tested it

`test/agent/bedrock-model.test.ts` — 18 tests, all passing:

- **Raises to 16,384** for bare foundation-model IDs, `us.`/`eu.`/`apac.`/`us-gov.` inference profiles, and inference-profile ARNs
- **Leaves the provider default in place** for non-Anthropic vendors, Claude 3.5, Claude 3 Haiku, and unrecognized IDs
- **Explicit override wins** — both the run option and `OPENWIKI_MAX_OUTPUT_TOKENS`, the latter verified on a model with no known ceiling
- Updated the existing "preserves the provider default" test to a non-Anthropic model; it previously pinned `anthropic.claude-sonnet-5` to *no* `maxTokens`, which is the behavior this fixes
- Added `OPENWIKI_MAX_OUTPUT_TOKENS` to the suite's env isolation list

Full gate green: `pnpm run format`, `pnpm run lint`, `pnpm test` (220 files, 2951 tests). Changeset included (patch).

> Note, unrelated: `pnpm test` prints `stamp-build-channel failed: expected exactly one BUILD_CHANNEL assignment to stamp, found 0`. It reproduces on unmodified `main` and is not caused by this change — flagging it rather than fixing it here, per the one-PR-one-change rule.

Fixes #722

### DIFF
--- a/.changeset/bedrock-default-max-output-tokens.md
+++ b/.changeset/bedrock-default-max-output-tokens.md
@@ -1,0 +1,12 @@
+---
+"openwiki": patch
+---
+
+fix: default Bedrock runs on modern Claude models to a 16,384-token output ceiling
+
+The Bedrock Converse API caps output at 4,096 tokens when a request sends no
+`inferenceConfig`, which truncated long pages mid-write and ended the run
+cleanly, looking like a page-count limit. Bedrock runs on modern Claude models
+now use the same 16,384-token ceiling as the direct Anthropic provider. Models
+without a known-wide output window keep the provider default, and
+`OPENWIKI_MAX_OUTPUT_TOKENS` still overrides both.
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1215,10 +1215,17 @@ export function createModel(
   }
 
   if (provider === "bedrock") {
+    const bedrockMaxTokens = resolveBedrockMaxOutputTokens(
+      modelId,
+      configuredMaxOutputTokens,
+    );
+
     return new ChatBedrockConverse({
       model: modelId,
       region: resolveProviderRegion(provider),
-      ...maxTokensOptions,
+      ...(bedrockMaxTokens !== undefined
+        ? { maxTokens: bedrockMaxTokens }
+        : {}),
       ...streamIdleTimeoutOptions,
       ...retryOptions,
     });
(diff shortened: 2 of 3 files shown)


Payal2000:
Closing this — retracting on the author's side, not because of anything in the review. The root cause and the analysis in #722 still stand if anyone wants to pick it up.

PR #767, opened 2026-08-30
fix: tolerate human-readable not-found errors from deepagents backends

### LINKED ISSUE #765: bug: restoreRepositoryPageMarkdown crashes on new-page skip due to DeepAgents error format mismatch
## Summary

`restoreRepositoryPageMarkdown` (introduced in #732) crashes the **entire run** when a page worker fails and the skip-restore path attempts to delete a non-existent file. The root cause is an **error format mismatch** between openwiki and DeepAgents: openwiki expects error code `"file_not_found"`, but DeepAgents returns a human-readable string `"Error: File '<path>' not found"`.

This turns an otherwise recoverable single-page failure into a fatal crash that aborts the entire generation run, discarding progress on all other pages.

## Reproduction

1. Run `openwiki --init` on a non-trivial repository
2. If **any** new page worker fails (model timeout, network error, output validation failure, etc.), the run crashes instead of skipping that page and continuing
3. Re-running `openwiki --init` resumes, but the same page fails again -> same crash -> infinite loop

Observed on v0.4.3 with 19 planned pages. After 3 consecutive runs, the run crashed at different pages each time (pages 10, 16, and 16 again), never completing.

## Root Cause

In `src/generation/repository-run.ts`, `restoreRepositoryPageMarkdown`:

```typescript
async function restoreRepositoryPageMarkdown(
  run: ActiveRepositoryRun,
  snapshot: RepositoryPageSnapshot,
): Promise<void> {
  const result =
    snapshot.markdown === null
      ? await run.backend.delete(snapshot.path)
      : await run.backend.write(snapshot.path, snapshot.markdown);
  if (
    result.error &&
    !(snapshot.markdown === null && result.error === "file_not_found")
  ) {
    throw new RepositoryRunError(
      "invalid_state",
      `Could not restore ${snapshot.path}: ${result.error}`,
    );
  }
}
```

**The trigger chain:**

1. A new page worker fails (e.g. model timeout, API error)
2. `skipRepositoryPage` is called to roll back the failed page
3. `captureRepositoryPageSnapshot` captures the snapshot -- for a **new page that was never written**, `snapshot.markdown` is `null`
4. `restoreRepositoryPageMarkdown` enters the `delete` branch (`snapshot.markdown === null`)
5. DeepAgents' `delete()` returns `{ error: "Error: File '<path>' not found" }` for a file that does not exist
6. The tolerance check `result.error === "file_not_found"` fails because the actual value is `"Error: File '...' not found"` (a human-readable string, not the error code `"file_not_found"`)
7. The `RepositoryRunError` is thrown -> **entire run crashes**

**Evidence from DeepAgents dist** (`node_modules/deepagents/dist/langsmith-hz83LfzA.js:828`):

```javascript
// DeepAgents delete() returns a human-readable string, not an error code:
if (!(filePath in this.files))
  return { error: `Error: File '${filePath}' not found` };
```

## Impact

- **Every single new-page worker failure** crashes the entire run, regardless of how many other pages succeeded
- 15/19 pages were generated and saved to disk, but the run never finalizes (`.run.json` is not deleted), so those pages cannot be used as a baseline for `--update`
- Re-running always re-plans and re-generates from scratch (or resumes and crashes again at the same failing page), wasting tokens

## Suggested Fix

The tolerance check should accept both the error code `"file_not_found"` and any human-readable "not found" message:

```typescript
if (
  result.error &&
  !(snapshot.markdown === null &&
    (result.error === "file_not_found" || result.error.includes("not found")))
) {
  throw new RepositoryRunError(
    "invalid_state",
    `Could not restore ${snapshot.path}: ${result.error}`,
  );
}
```

Alternatively, the check could be simplified -- when `snapshot.markdown === null`, **any** delete error on a non-existent file should be tolerated, since the intent is "ensure the file is gone" and it already is:

```typescript
// When markdown is null (delete path), tolerate ALL errors
if (result.error && snapshot.markdown !== null) {
  throw new RepositoryRunError(
    "invalid_state",
    `Could not restore ${snapshot.path}: ${result.error}`,
  );
}
```

## Context

This bug was introduced by #732 (`fix: skip failed page workers without aborting updates`, merged 2026-08-26). The irony is that #732's stated goal was exactly this -- allowing page worker failures without aborting the run -- but the implementation introduced a secondary crash in the restore path.

The fix was merged before v0.4.3 was released (0.4.3: 2026-08-27, #732: 2026-08-26), so v0.4.3 contains this code.

## Workaround

Patch the installed dist file at `node_modules/openwiki/dist/generation/repository-run.js`:

```diff
-        !(snapshot.markdown === null && result.error === "file_not_found")) {
+        !(snapshot.markdown === null &&
+          (result.error === "file_not_found" || result.error.includes("not found")))) {
```

This patch must be reapplied after each `npm install -g openwiki` upgrade.

forrinzhao:
## Summary

Fixes the run-aborting crash described in #765: DeepAgents filesystem backends return a **human-readable** `"Error: File '...' not found"` string — not the `"file_not_found"` error code — when a `delete`/`readRaw` targets a file that never existed. The restore/skip path compared only against the error code, so **any** page-worker failure on a new page threw `invalid_state` and aborted the whole generation run.

## Problem

In `restoreRepositoryPageMarkdown` (introduced in #732), the tolerance check was:

```typescript
!(snapshot.markdown === null && result.error === "file_not_found")
```

When a new-page worker fails, `skipRepositoryPage` rolls back by `delete`-ing the page. For a page that was never written, `snapshot.markdown` is `null` and the backend returns `"Error: File '...' not found"` — which does **not** equal `"file_not_found"` — so the rollback threw and the entire run crashed. 15/19 pages could be on disk and the run still never finalized, wasting tokens on every retry.

## Change

Centralize not-found classification in `isNotFoundBackendError` (matches both the error code and the human-readable string) and apply it to all four backend `file_not_found` tolerance sites in `src/generation/repository-run.ts`:

- `captureRepositoryPageSnapshot` read
- `restoreRepositoryPageMarkdown` (the crash site)
- `applyAbandonedGeneratedPageDeletions`
- `applyPlannedDeletions`

Plus a regression test: mock `backend.delete` to return the human-readable not-found error and assert `skipRepositoryPage` resolves instead of throwing.

## Verification

- `pnpm run typecheck` — pass
- `npx vitest run test/generation/repository-run.test.ts` — 44 passed
- `eslint` and `prettier` on changed files — pass

Fixes #765

### DIFF
--- a/.changeset/tolerant-not-found-restore.md
+++ b/.changeset/tolerant-not-found-restore.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: tolerate human-readable "not found" errors from DeepAgents backends when rolling back failed page workers and deleting non-existent pages, instead of aborting the whole run
--- a/src/generation/repository-run.ts
+++ b/src/generation/repository-run.ts
@@ -544,6 +544,17 @@ export async function beginRepositoryRun(
 }
 
 /**
+ * True when a backend result error indicates a file does not exist.
+ *
+ * Matches both the standard `"file_not_found"` error code used by some backends
+ * and the human-readable `"Error: File '...' not found"` string returned by
+ * DeepAgents' filesystem backends (see #765).
+ */
+function isNotFoundBackendError(error: string): boolean {
+  return error === "file_not_found" || error.includes("not found");
+}
+
+/**
  * Reconstructs a durable run and invalidates its complete plan on source drift.
  *
  * @param input - Current begin request used to validate the durable owner.
@@ -1014,7 +1025,7 @@ export async function captureRepositoryPageSnapshot(
   let markdown: string | null = null;
   try {
     const read = await run.backend.readRaw(current.path);
-    if (read.error && read.error !== "file_not_found") {
+    if (read.error && !isNotFoundBackendError(read.error)) {
       throw new RepositoryRunError(
         "invalid_state",
         `Could not snapshot ${current.path}: ${read.error}`,
@@ -1118,7 +1129,7 @@ async function restoreRepositoryPageMarkdown(
       : await run.backend.write(snapshot.path, snapshot.markdown);
   if (
     result.error &&
-    !(snapshot.markdown === null && result.error === "file_not_found")
+    !(snapshot.markdown === null && isNotFoundBackendError(result.error))
   ) {
     throw new RepositoryRunError(
       "invalid_state",
@@ -1581,7 +1592,7 @@ async function applyAbandonedGeneratedPageDeletions(
   for (const page of await store.discoverPages()) {
     if (initial.has(page) || planned.has(page)) continue;
     const result = await run.backend.delete(page);
-    if (result.error && result.error !== "file_not_found") {
+    if (result.error && !isNotFoundBackendError(result.error)) {
       throw new RepositoryRunError(
         "invalid_state",
         `Could not remove page abandoned by an invalidated plan ${page}: ${result.error}`,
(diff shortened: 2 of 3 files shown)


colifran (approved):
Nice. Thanks for fixing!

PR #777, opened 2026-08-31
fix: import AGENTS.md into the managed CLAUDE.md block

### LINKED ISSUE #719: [Bug]: CLAUDE.md symlinked to AGENTS.md — concurrent snippet writes clobber each other and corrupt the managed block (since 0.3.3)

### LINKED ISSUE #776: [Bug]: The managed CLAUDE.md block is inert in Claude Code
### Summary

The managed block OpenWiki writes into `CLAUDE.md` points at `AGENTS.md` with a
plain Markdown link. Claude Code does not follow it, and does not read
`AGENTS.md` on its own while a `CLAUDE.md` sits beside it. The result is that
the OpenWiki instructions reach Codex but never reach Claude Code — the one host
`CLAUDE.md` exists for.

Written into `CLAUDE.md` (0.4.0 through 0.4.3):

```markdown
<!-- OPENWIKI:START -->

## OpenWiki

<!-- OPENWIKI:END -->
```

Claude Code expands only its own `@path` import syntax. A Markdown link is inert
text to the loader.

### Reproduction

In a repo with both files, ask Claude Code which instruction files it loaded:

```console
$ claude -p "List only the file paths of the project instruction files given to you as claudeMd context."
/Users/…/.claude/CLAUDE.md
/Users/…/repo/CLAUDE.md
/Users/…/repo/.claude/rules/*.md
```

No `AGENTS.md`. Then the direct test — `AGENTS.md` holds a token nothing else
mentions:

| `CLAUDE.md` contains | Token reachable? |
| --- | --- |
| `See [AGENTS.md](AGENTS.md) for OpenWiki agent instructions.` | no — the model guesses |
| `@AGENTS.md` | yes |

Claude Code 2.1.251, macOS. `AGENTS.md` discovery is hardcoded in the binary but
only applies as a fallback when no `CLAUDE.md` is present, which an OpenWiki run
guarantees there is.

### Suggested fix

Write the import form into the `CLAUDE.md` block:

```markdown
<!-- OPENWIKI:START -->

@AGENTS.md

<!-- OPENWIKI:END -->
```

Or, if a self-contained block is preferable to an import, put the same
instruction text in `CLAUDE.md` that `AGENTS.md` already carries, rather than a
reference to it. Either way the current form is the one that cannot work.

Worth noting for whoever picks this up: the fix interacts with #719 and #725,
where `CLAUDE.md` is a symlink to `AGENTS.md`. In that setup `@AGENTS.md` inside
the block would make the file import itself, so the writer needs the same
symlink check those issues are about.

### Workaround

Put `@AGENTS.md` in `CLAUDE.md` *outside* the managed markers — inside, the next
`openwiki code --update` overwrites it, since the block is rewritten wholesale.

### Environment

- openwiki 0.4.0 and 0.4.3, `openwiki code --update --print`
- Claude Code 2.1.251
- four repos, scheduled GitHub Actions workflow

danielsogl:
Fixes #776.

### What was wrong

The managed block written into `CLAUDE.md` pointed at `AGENTS.md` with a
Markdown link. Claude Code expands only its own `@path` import syntax, and it
falls back to reading `AGENTS.md` itself only when no `CLAUDE.md` sits beside
it — which, after code-mode setup, is never. So the block was inert: the
instructions reached Codex and never reached Claude Code, the host `CLAUDE.md`
exists for.

Verified against Claude Code 2.1.251 with a token that appears only in
`AGENTS.md`:

| `CLAUDE.md` contains | Token reachable? |
| --- | --- |
| `See [AGENTS.md](AGENTS.md) for OpenWiki agent instructions.` | no — the model guesses |
| `@AGENTS.md` | yes |

### The change

`createCodeModeClaudeSnippet()` writes `@AGENTS.md`. The block keeps its shape
and `AGENTS.md` stays the single source of the instructions; only the reference
form changes.

The one case where the import is wrong is a repository that makes `CLAUDE.md` a
link to `AGENTS.md` — the setup in #719 and #725 — because the file would then
import itself. `writeCodeModeAgentSnippets` compares the two paths by device and
inode and writes the full agents snippet to both when they are the same file.
Inode `0` counts as unknown, since some Windows filesystems report it for every
file and would otherwise make unrelated paths compare equal.

This does not attempt to fix #719 itself; it only avoids introducing a
self-import into that setup.

### Tests

- The existing "CLAUDE.md is a simple reference to AGENTS.md" test now asserts
  the import form and rejects the Markdown link.
- New: with `CLAUDE.md` symlinked to `AGENTS.md`, the block carries the
  instructions and contains no `@AGENTS.md`.

`pnpm run format:check`, `pnpm run lint:check`, `pnpm run typecheck` and
`vitest run test/ingestion/code-mode.test.ts` (26 passed) are green locally.

### Note for existing repositories

The block is rewritten wholesale on the next run, so existing repositories pick
this up automatically. Anyone who added `@AGENTS.md` outside the markers as a
workaround will want to drop it afterwards to avoid importing the file twice.


colifran (approved):
Thanks, @danielsogl!

PR #788, opened 2026-09-01
feat: let openai-compatible opt into reasoning effort

### LINKED ISSUE #686: [Feature]: Let openai-compatible opt into OPENWIKI_REASONING_EFFORT
### Summary

The `openai-compatible` provider can never use `OPENWIKI_REASONING_EFFORT`. `resolveReasoningConfig` looks up an exact `(provider, modelId)` pair in `REASONING_CAPABILITIES` and throws on a miss, but `openai-compatible` ships `modelOptions: []` by design, so no model ID a user types there will ever be in the table.

I would like to add an opt-in escape hatch, mirroring how the same provider already opts into the Responses API.

### Motivation

Calling `resolveReasoningConfig` against `upstream/main` with `OPENWIKI_REASONING_EFFORT=high`:

```
THROW openai-compatible  Qwen/Qwen3.7-235B   OPENWIKI_REASONING_EFFORT is not supported
                                             for provider "openai-compatible" and model
                                             "Qwen/Qwen3.7-235B".
```

The throw is unconditional for this provider, whatever the model ID.

That is unfortunate, because `openai-compatible` is the provider most likely to be pointed at a reasoning model. Users reach it with a local vLLM or a gateway serving Qwen or DeepSeek, and #256 is already open about thinking-model handling on this same code path. The model list is empty on purpose here: OpenWiki cannot know what the endpoint serves, so it cannot decide whether the model reasons.

This differs from a curated model that is genuinely known not to support an effort value. There, failing before a request is sent is the documented and correct behaviour. The gap is only about the case where OpenWiki has no basis for either answer, and currently resolves that uncertainty by refusing.

### Proposed Solution

Let the user assert the capability, exactly as they already do for the Responses API on this provider:

```ts
// src/config/constants.ts
"openai-compatible": {
  responsesApi: resolveOpenAiCompatibleUseResponsesApi(),  // OPENWIKI_OPENAI_COMPATIBLE_USE_RESPONSES_API
  ...
}
```

A matching `OPENWIKI_OPENAI_COMPATIBLE_REASONING_EFFORT_SUPPORTED` (or similar) would make `resolveReasoningConfig` accept any effort value for this provider and forward it, leaving the endpoint to reject what it does not understand. That is the same trust boundary the project already applies to custom base URLs.

The transport needs no new decision: `providerUsesResponsesApi(provider, modelId)` already resolves it per provider and model, including this provider's opt-in.

Everything else stays as it is. Curated providers keep the strict table and the early failure.

### Alternatives Considered

**Leave it as is.** Users can still configure reasoning by other means at the gateway itself. But `OPENWIKI_REASONING_EFFORT` is documented as provider-neutral, and one provider being structurally excluded from it is surprising.

**Match model IDs by pattern instead of exact string.** Useful for self-hosted deployments that rename a known model, and there is precedent for it (`responsesApi: /^gpt-5/u` on copilot). It does not help here though, since an arbitrary gateway's model IDs follow no pattern OpenWiki can predict.

### Additional Context

- `OPENWIKI_REASONING_EFFORT` was added in #634, closing #625.
- The Responses API opt-in referenced above landed in #266.
- Related: #256.
- Happy to open the PR if this direction sounds right.

HwangJohn:
## Summary

- Add an explicit `OPENWIKI_OPENAI_COMPATIBLE_REASONING_EFFORT_SUPPORTED` opt-in for OpenAI-compatible gateways that support reasoning effort.
- Route reasoning effort through the existing Responses API transport when enabled, or chat-completions model kwargs otherwise.
- Validate/document the new environment variable and add focused config/model coverage plus a minor changeset.

Fixes #686

## Testing

- Windows: `npm exec --yes pnpm@10.33.2 -- pnpm run format`
- Windows: `npm exec --yes pnpm@10.33.2 -- pnpm run lint`
- Windows: `npm exec --yes pnpm@10.33.2 -- pnpm exec vitest run test/config/constants.test.ts test/config/env.test.ts test/config/env-behavior.test.ts test/agent/create-model.test.ts test/openai-compatible-responses.test.ts test/openai-compatible-streaming.test.ts test/agent/stream-modes.test.ts`; the grouped run passed the new coverage and 200 tests, while two existing slower OpenAI-compatible files exceeded Vitest's 5s per-test timeout when grouped.
- Windows: `npm exec --yes pnpm@10.33.2 -- pnpm exec vitest run test/openai-compatible-responses.test.ts` passed.
- Windows: `npm exec --yes pnpm@10.33.2 -- pnpm exec vitest run test/openai-compatible-streaming.test.ts` passed.
- Windows: attempted `npm exec --yes pnpm@10.33.2 -- pnpm test`; typecheck/build passed, but the coverage phase is blocked on this local Windows account because `fs.symlinkSync` fails with `EPERM`.
- DGX Spark Ubuntu: `pnpm install --frozen-lockfile && pnpm run format:check && pnpm run lint:check && pnpm test` passed. Summary: 225 test files passed, 3040 tests passed, 2 files/3 tests skipped.


colifran (approved):
Thanks, @HwangJohn!

PR #794, opened 2026-09-01
fix: unescape .env values in a single atomic pass

### LINKED ISSUE #793: .env values with a backslash before `n`/`r` (Windows paths) get corrupted with a stray newline/CR on load
## Bug description

`~/.openwiki/.env` values containing a raw backslash immediately followed by `n` or `r` get corrupted with a stray newline or carriage return the next time they're loaded. This silently mangles any saved value that looks like a Windows path, e.g. `GOOGLE_APPLICATION_CREDENTIALS=C:\name\creds.json` or any path where a folder name starts with `n` or `r` (`C:\repos\...`, `C:\notes\...`).

## Root cause

`formatEnvValue` (`src/config/env.ts`) escapes a value with four independent, sequential regex passes:

```js
function formatEnvValue(value) {
  return `"${value
    .replace(/\\/gu, "\\\\")
    .replace(/"/gu, '\\"')
    .replace(/\n/gu, "\\n")
    .replace(/\r/gu, "\\r")}"`;
}
```

`parseEnvValue` tries to reverse this with the same four-pass approach, in the same relative order:

```js
function parseEnvValue(value) {
  if (value.startsWith('"') && value.endsWith('"')) {
    return value
      .slice(1, -1)
      .replace(/\\n/gu, "\n")
      .replace(/\\r/gu, "\r")
      .replace(/\\"/gu, '"')
      .replace(/\\\\/gu, "\\");
  }
  return value;
}
```

The problem: sequential independent `.replace()` calls aren't safe for this escaping scheme in *any* relative order, because unescaping one sequence can produce (or leave adjacent) text that a different pass's pattern also matches.

Concretely, for the raw value `C:\name\creds.json`:

1. `formatEnvValue` escapes the single backslash before `name` to `\\`, producing the literal 20-character sequence `C:\\name\\creds.json` (each `\` is now two `\` characters).
2. On load, `parseEnvValue`'s first pass (`/\\n/g` → real newline) scans left to right. At the position of the *second* backslash of the escaped pair, followed by the literal `n` of `name`, it finds a match — `\` + `n` — and converts it to an actual newline, before the `/\\\\/g` pass (which runs last) ever gets a chance to collapse the original escaped pair back to a single backslash.

Result: `C:\name\creds.json` round-trips to `` C:\ `` + newline + `` ame\creds.json `` — a corrupted value with an embedded newline and a dropped character. The equivalent case with `\r` (e.g. a folder starting with `r`, like `C:\repos\...`) corrupts the same way with a carriage return.

## Steps to reproduce

```js
import { parseEnv, formatEnv } from "./src/config/env.ts";

const original = { GOOGLE_APPLICATION_CREDENTIALS: "C:\\name\\creds.json" };
console.log(parseEnv(formatEnv(original)));
// Expected: { GOOGLE_APPLICATION_CREDENTIALS: "C:\\name\\creds.json" }
// Actual:   { GOOGLE_APPLICATION_CREDENTIALS: "C:\\\name\\creds.json" }
//           (the \n between C:\ and ame is a real embedded newline, not two characters)
```

## Impact

Any credential-setup flow that persists a value shaped like this (very plausible on Windows, which OpenWiki explicitly supports — e.g. `GOOGLE_APPLICATION_CREDENTIALS` pointing at a service-account JSON key under a folder starting with `n`/`r`) gets that value silently corrupted on the very next `.env` load, most likely surfacing as a confusing "file not found" or "invalid credentials" error with no indication the `.env` value itself is the problem.

None of the existing tests in `test/config/env.test.ts` cover a raw backslash immediately adjacent to a literal `n`/`r` character, which is why this has gone unnoticed.

## Environment

Not version-specific — this is a logic bug in `parseEnvValue`/`formatEnvValue` present on current `main`.

Yashwanth-Kumar-Kotla:
### Description

`parseEnvValue` unescaped `.env` values with four sequential, independent `.replace()` calls. That's not safe for this escaping scheme in any order: unescaping one sequence can produce (or leave adjacent) text that a different pass's pattern also matches.

Concretely, a raw backslash escaped to `\\` and immediately followed by a literal `n` or `r` (e.g. a Windows path like `C:\name\creds.json` or `C:\repos\...`) gets misread by the `\n`/`\r` unescape pass as an actual newline/carriage-return escape sequence, before the `\\` pass ever gets a chance to collapse the original escaped pair. The value comes back corrupted with a stray embedded newline/CR and a dropped character.

Fixes #793

### Changes Made

- `src/config/env.ts`: replaced the four sequential `.replace()` calls in `parseEnvValue` with a single left-to-right pass (`/\\(.)/gsu` with a switch on the captured character) that consumes each backslash escape as one atomic unit, so an unescaped sequence can never be re-interpreted by a later (or earlier) pass.
- `test/config/env.test.ts`: added a regression test asserting Windows-path-shaped values (backslash immediately before `n`/`r`) survive a `formatEnv` → `parseEnv` round-trip.

### Testing

- `pnpm run format:check`
- `pnpm run lint:check`
- `pnpm test` (typecheck, build, full Vitest suite with coverage) — all green
- Regression test verified against the actual bug: reverted only the `src/config/env.ts` fix (kept the new test) and confirmed the new test fails with the expected corrupted output, then restored the fix and confirmed it passes; the other 22 existing tests in the file were unaffected throughout.


colifran (approved):
Thanks, @Yashwanth-Kumar-Kotla!

PR #796, opened 2026-09-01
fix: allow the google fonts origins the visualizer page requests

### LINKED ISSUE #795: Visualizer's CSP blocks its own Google Fonts request; Inter never loads
## Bug description

`openwiki visualize` (and any static-exported visualizer output) requests the Inter font from Google Fonts, but its own Content-Security-Policy blocks that exact request. Every browser rendering the page silently drops the font request and falls back to a system font — the visualizer never actually appears in the typeface it's built for.

## Root cause

`src/visualize/page.ts` sets:

```js
export const CSP = [
  "default-src 'none'",
  `script-src 'self' ${CDN}`,
  "style-src 'self' 'unsafe-inline'",
  "img-src 'self' data:",
  "font-src 'self'",
  "connect-src 'self'",
  "base-uri 'none'",
  "form-action 'none'",
].join("; ");
```

and unconditionally includes, in the page `<head>`:

```html
<link rel="preconnect" href="https://fonts.googleapis.com" />
<link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;800&display=swap" rel="stylesheet" />
```

`style-src` only allows `'self'` and `'unsafe-inline'` — neither permits loading an external stylesheet from a different origin (`'unsafe-inline'` only covers inline `<style>` elements/attributes, not `<link rel="stylesheet">` from another host). `font-src 'self'` similarly doesn't allow the actual font files, which `fonts.googleapis.com`'s returned CSS references on `fonts.gstatic.com`.

This CSP is enforced in both modes: as an HTTP `content-security-policy` response header for the live `openwiki visualize` server (`src/visualize/server.ts:172`), and as a `<meta http-equiv="Content-Security-Policy">` tag for a static export (`src/visualize/page.ts`, `staticExport` branch). `styles.css` sets `font-family: "Inter", ...` (lines 55-57 and 650), so this is the typeface the whole UI is designed around.

## Steps to reproduce

1. Run `openwiki visualize` (or export a static visualizer and open the HTML file through any local server) and open browser devtools.
2. Observe a CSP violation in the console:
   ```
   Loading the stylesheet 'https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;800&display=swap' violates the following Content Security Policy directive: "style-src 'self' 'unsafe-inline'". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback. The action has been blocked.
   ```
3. The Inter font never loads; the page silently renders in a fallback font instead.

I reproduced this concretely by rendering the actual `STATIC_PAGE` export from `src/visualize/page.ts`, serving it locally, and opening it in a real browser — the violation above is the exact console output.

## Impact

Cosmetic but universal: every user of `openwiki visualize`, and every consumer of a static-exported visualizer, silently loses the intended typography on every single run. The app's own CSP defeats its own designed appearance.

## Suggested fix

Add the two specific Google Fonts origins the page itself already depends on — `https://fonts.googleapis.com` to `style-src`, `https://fonts.gstatic.com` to `font-src` — rather than loosening the policy any further. Nothing else about the CSP needs to change.

Yashwanth-Kumar-Kotla:
### Description

The visualizer's own Content-Security-Policy blocked its own Google Fonts request. `style-src` only allowed `'self' 'unsafe-inline'` and `font-src` only allowed `'self'`, but the page unconditionally requests Inter via a `<link rel="stylesheet" href="https://fonts.googleapis.com/...">`. Every browser enforcing this CSP (as an HTTP header for the live `openwiki visualize` server, or as a `<meta>` tag for a static export) silently blocks the request, so the visualizer never actually renders in its intended typeface.

Fixes #795

### Changes Made

- `src/visualize/page.ts`: added `https://fonts.googleapis.com` to `style-src` and `https://fonts.gstatic.com` to `font-src` — the two specific origins the page already depends on. Nothing else about the CSP changes.
- `test/visualize/page.test.ts`: added a regression test that parses the CSP directives and the page's actual Google Fonts `<link>` href, then asserts both required origins are allowed.

### Testing

- `pnpm run format:check`
- `pnpm run lint:check`
- `pnpm test` (typecheck, build, full Vitest suite with coverage) — all green
- Verified the actual bug and the fix live in a real browser, not just by reading the CSP spec: rendered the real `STATIC_PAGE` export, served it locally, and opened it in a browser.
  - Before the fix: console shows `Loading the stylesheet 'https://fonts.googleapis.com/css2?family=Inter...' violates the following Content Security Policy directive: "style-src 'self' 'unsafe-inline'"... The action has been blocked.`
  - After the fix: no CSP violation.
- Regression test verified against the actual bug: reverted only the `src/visualize/page.ts` fix (kept the new test), confirmed the new test fails with the pre-fix CSP, then restored the fix and confirmed it passes; the other 9 existing tests in the file were unaffected throughout.


colifran (approved):
thanks, @Yashwanth-Kumar-Kotla!

PR #801, opened 2026-09-02
feat: map gemini reasoning effort to thinking level

### LINKED ISSUE #786: [Feature]: Map OPENWIKI_REASONING_EFFORT onto Gemini's thinkingLevel
### Summary

`OPENWIKI_REASONING_EFFORT` can never be used with the `gemini` provider. `REASONING_CAPABILITIES` in `src/config/reasoning.ts` has entries for `openai`, `openai-chatgpt` and `nvidia`, but no `gemini` key at all, so `resolveReasoningConfig` misses the lookup and throws for every Gemini model:

```
OPENWIKI_REASONING_EFFORT is not supported for provider "gemini" and model "gemini-3.6-flash".
Error: Process completed with exit code 1.
```

This looks like a mapping gap rather than a capability gap: the Gemini API does expose a reasoning control, `generationConfig.thinkingConfig.thinkingLevel`, and it works on the models OpenWiki already lists in its picker. I verified this against a live AI Studio key (numbers below), so the entry can be added with concrete values rather than guessed ones.

Related but distinct: #686 asks for an opt-in escape hatch for `openai-compatible`, where OpenWiki genuinely cannot know what the endpoint serves. Gemini is the opposite case — the models are curated in `constants.ts` and the capability is knowable, so a normal curated entry fits rather than a user-asserted override.

### Motivation

The failure lands on the combination the docs steer you toward. A repo using the recommended `OPENWIKI_PROVIDER: gemini` + `GEMINI_API_KEY` setup, plus the documented `OPENWIKI_REASONING_EFFORT`, hard-fails with exit 1 before doing any work. In our case that silently froze a scheduled wiki refresh: the workflow had been green, someone added the documented env var, and every subsequent run aborted during config resolution.

Two things make it more surprising than a plain unsupported-feature error:

- It is a **startup throw, not a degradation**. There is no warning-and-continue path, so a config value that is merely inapplicable takes the whole run down.
- The error names the model, which reasonably suggests trying a different one. No Gemini model helps — the lookup is keyed on `(provider, modelId)` and the provider has no entries, so 3.5, 3.6, 3.7 and pro all fail identically. We only established that by reading `reasoning.js` in the installed package.

Gemini is also the provider where the knob arguably matters most for this tool: it is the cheap default many CI setups land on, and an unattended docs run is exactly the workload where you would happily trade latency for more thinking.

### Proposed Solution

Add a `gemini` entry to `REASONING_CAPABILITIES` with a transport that maps the effort onto `generationConfig.thinkingConfig.thinkingLevel`:

```ts
const GEMINI_THINKING_LEVEL_CAPABILITY = {
  transport: "gemini-thinking-level",
  values: ["low", "medium", "high"],
};

const REASONING_CAPABILITIES = {
  // ...
  gemini: {
    "gemini-3.7-flash": GEMINI_THINKING_LEVEL_CAPABILITY,
    "gemini-3.6-flash": GEMINI_THINKING_LEVEL_CAPABILITY,
    "gemini-3.5-flash": GEMINI_THINKING_LEVEL_CAPABILITY,
    // ...whichever of the picker's models you want to claim
  },
};
```

The request shape is:

```jsonc
{
  "contents": [ /* ... */ ],
  "generationConfig": {
    "thinkingConfig": { "thinkingLevel": "high" }
  }
}
```

A partial `values` list is already an established pattern here — the existing `nvidia` entry uses `["none", "low", "high"]` — and `resolveReasoningConfig` already produces a good message for a value outside the set, so `xhigh`/`max` on Gemini would report *"Supported values: low, medium, high"* instead of the current blanket "not supported".

### Alternatives Considered

**Leave it and document that Gemini is unsupported.** Cheapest, but the current failure mode is a hard exit during config resolution for a value that is merely inapplicable. If the mapping is not wanted, downgrading the throw to a warning-and-ignore for providers with no entry would at least stop a documented env var from bricking a scheduled run.

**Use `thinkingBudget` instead of `thinkingLevel`.** Also accepted by the API, but it is an integer token budget, which does not map cleanly onto six named efforts and would need per-model tuning. `thinkingLevel` is already a named-tier control, so it lines up with the existing abstraction. Worth noting `thinkingBudget: 0` did **not** eliminate thinking on `gemini-3.7-flash` (152 thought tokens), so it appears advisory on that model — another reason to prefer the level.

**Switch provider to OpenAI.** What the error effectively forces today. It means a second API key and a different cost profile purely to set one env var, which is a poor trade for a docs job.

**An opt-in override like #686.** Right for `openai-compatible`, where the served model is unknowable. Overkill for Gemini, where the model list is curated in-repo.

### Additional Context

Measured against `gemini-3.7-flash` on a live AI Studio key, same prompt each time, reading `usageMetadata.thoughtsTokenCount`:

| `thinkingLevel` | Result | Thought tokens |
| :--- | :--- | ---: |
| `none` | rejected 400 — invalid value | — |
| `minimal` | rejected 400 — *"not supported for this model"* | — |
| `low` | accepted | 169 |
| `medium` | accepted | 240 |
| `high` | accepted | 376 |
| `xhigh` | rejected 400 — invalid value | — |
| `max` | rejected 400 — invalid value | — |

The tiers are monotonic and the spread is real, so the knob does something rather than being accepted and ignored. A separate run showed the same on `gemini-3.6-flash` (958 thought tokens at `high`), so this is not specific to the newest model.

Mapping implication: Gemini supports exactly three of OpenWiki's six effort values. `none`, `xhigh` and `max` would need to be rejected by the existing `capability.values` check rather than forwarded.

Environment: `openwiki@0.4.3` (current latest at time of writing), `OPENWIKI_PROVIDER=gemini`, Google AI Studio key, Node 22 in GitHub Actions.

Happy to open a PR for the capability entry if the transport name and the model set look right to you — the part I would not want to guess at is which of the picker's models you want to claim support for.

ousamabenyounes:
Fix #786

## Summary
- Maps `OPENWIKI_REASONING_EFFORT=low|medium|high` to Gemini's `thinkingLevel` for Gemini 3.6 Flash.
- Adds Gemini reasoning capability resolution and rejection coverage for unsupported values.
- Documents the new Gemini behavior and includes a patch changeset.

- BASE RED: `pnpm exec vitest run test/config/constants.test.ts test/agent/create-model.test.ts`
  - Failed before the production change with `4 failed | 127 passed`.
  - Failure included: `OPENWIKI_REASONING_EFFORT is not supported for provider "gemini" and model "gemini-3.6-flash"`.
- GREEN: same targeted command passed with `2 passed`, `131 passed`.
- Revert RED: stashing only the production source changes made the same targeted command fail again with `4 failed | 127 passed` and the same unsupported-Gemini behavior.
- Full local validation suite baseline on `main`: `3032 passed | 3 skipped`.
- Full local validation suite final on this branch: `3033 passed | 3 skipped`.
- Production diff coverage gate: `2/2` changed executable source lines covered.
- Additional checks: `pnpm run typecheck`, `pnpm run format:check`, `pnpm exec changeset status --since=origin/main`, and `git diff --check origin/main`.

## Local CI Notes
- The dry-run CLI smoke emits the same Ink raw-mode warning in this non-interactive shell on both baseline and final runs, and exits successfully.
- Node 24, Windows, and Trivy replay were not available locally; visible remote checks are monitored after PR creation.


colifran (approved):
thanks, @ousamabenyounes!

PR #817, opened 2026-09-06
fix: handle updates-mode stream chunks for openai-compatible provider

### LINKED ISSUE #803: [Bug]: Interactive chat TUI shows "No assistant output captured" for plain-text replies via openai-compatible provider — openwiki -p works, interactive chat doesn't
### Description

With OPENWIKI_PROVIDER=openai-compatible pointed at a local Ollama model, any tool-free conversational reply (e.g. "who are you?") fails to display in the interactive TUI, showing "No assistant output captured" even though the underlying model response is confirmed complete and well-formed via a logging proxy. `openwiki -p "<question>"` correctly prints the answer for the identical question/model/config — only the interactive TUI is affected. The bug is in how OpenWiki/LangChain handle the model's streamed response internally, not the model or backend.

### Steps to Reproduce

1. Configure OPENWIKI_PROVIDER=openai-compatible against an Ollama "thinking" model that returns a non-standard `reasoning` field alongside `content` (e.g. qwen3 or gemma4).
2. Set OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true (needed to get past a separate, related silent-drop issue — see Additional Context).
3. Run `openwiki` (interactive) and ask a plain question with no expected tool use, e.g. "who are you?"
4. Exit and instead run `openwiki -p "who are you?"` with identical config.

### Expected Behavior

The interactive TUI should display the assistant's answer, consistent with `openwiki -p`, which correctly prints it for the exact same question, model, and configuration.

### Actual Behavior

The interactive TUI shows:

✓ Completed openwiki chat in Xs · <model>
  No assistant output captured.

despite the underlying Ollama response being fully complete and correct (confirmed via a logging proxy — see Additional Context for the raw evidence and full investigation).

### Environment

- OS: Linux (Ubuntu, x86_64)
- Node.js version: v24.19.0
- OpenWiki version: 0.4.3
- Provider: openai-compatible
- Backend: Ollama v0.32.15 (http://localhost:11434/v1)
- Models tested: qwen3.8:27b-128k, gemma4:e4b-128k (custom Ollama builds)
- @langchain/core: 1.2.9
- @langchain/openai: 1.5.10

### Additional Context

_No response_

Christian-Sidak:
## Summary

- The openai-compatible provider uses LangGraph "updates" stream mode by default (to avoid the reasoning-delta crash from issue #659). However, `isAgentStreamChunk` only accepted `"messages"` and `"tools"` mode chunks, so `"updates"` chunks were silently dropped and the interactive TUI showed "No assistant output captured" for every response.
- Adds `parseUpdatesChunk`, which extracts assistant text from the per-node state diff that LangGraph emits in `updates` mode. The diff shape is `{ nodeName: { messages: [...] }, ... }`; iterating its values and delegating to the existing `extractMessageText` routine surfaces the assistant reply.
- Extends `isAgentStreamChunk` to accept the `"updates"` mode string, then routes those chunks to the new handler instead of silently discarding them.

Fixes #803

## Test plan

- [x] Added unit tests in `test/agent/stream-redaction.test.ts` covering: basic updates-mode extraction, subgraph namespace tagging, empty state diff (null), non-record payload (null), and tool-call-only messages (null).
- [x] Manually verified extraction logic with a standalone JS script against all test cases before adding vitest tests.

### DIFF
--- a/.changeset/wobbly-foxes-stream.md
+++ b/.changeset/wobbly-foxes-stream.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: handle updates-mode stream chunks for openai-compatible provider
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1439,6 +1439,10 @@ export function parseAgentStreamChunk(chunk: unknown): OpenWikiRunEvent | null {
     return parseToolStreamEvent(payload);
   }
 
+  if (mode === "updates") {
+    return parseUpdatesChunk(namespace, payload);
+  }
+
   const text = extractMessageText(payload);
 
   return text.length > 0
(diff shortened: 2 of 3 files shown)


colifran (approved):
Thanks, @Christian-Sidak!

Issue #827, opened 2026-09-07
[Bug]: Finalization removes verification metadata from skipped pages


Totoro-qaq:
## Description

After a page is skipped during a repository update, `skipRepositoryPage` correctly restores its original Markdown and Claims sidecar. `finishRepositoryRun` then removes the page's existing OpenWiki `verified` metadata, while retaining its old sidecar and page-manifest entry. Their stored `pageVersion` hashes no longer match the Markdown on disk.

This affects previously verified pages. The page body is preserved, but its verification stamp and coverage consistency are lost. Finalization returns `{ status: "complete" }`; the update metadata still records `interrupted`, so this is not a report that the whole update is marked successful.

## Steps to Reproduce

From a source checkout at `b5d2407679825e10f86eb6ed159418563c9a8e45`, after `pnpm install --frozen-lockfile`, save the self-contained script below as `skipped-page-repro.mts` and run:

```bash
pnpm exec tsx skipped-page-repro.mts
```

The script creates and cleans up a temporary Git repository. It establishes its verified baseline through `begin → submit_plan → submit_page → finish`, then performs `begin(update) → capture → partial write → skip → finish`. It does not seed Claims or manifest JSON manually, and needs no model or API key.

<details>
<summary>Self-contained reproduction</summary>

```typescript
import assert from "node:assert/strict";
import { execFile } from "node:child_process";
import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
import { tmpdir } from "node:os";
import path from "node:path";
import { pathToFileURL } from "node:url";
import { promisify } from "node:util";

// Run from an OpenWiki source checkout after pnpm install --frozen-lockfile.
const checkout = process.cwd();
const directory = await mkdtemp(path.join(tmpdir(), "openwiki-skipped-repro-"));
const root = path.join(directory, "repo");
process.env.OPENWIKI_CONFIG_DIR = path.join(directory, "config");
process.env.OPENWIKI_TELEMETRY_DISABLED = "1";
process.env.DO_NOT_TRACK = "1";
process.env.LANGSMITH_TRACING = "false";
process.env.LANGCHAIN_TRACING_V2 = "false";
const execute = promisify(execFile);
const moduleUrl = (name: string) => pathToFileURL(path.join(checkout, "src", name)).href;
const api = await import(moduleUrl("generation/repository-run.ts"));
const { ClaimsStore } = await import(moduleUrl("claims/brains/code/store.ts"));
[...]

PR #828, opened 2026-09-07
fix: complete legacy page coverage and recover interrupted update plans

### LINKED ISSUE #826: [Bug]: Page coverage guard aborts the whole run on large existing wikis (page-manifest.js:262, 0.5.0)
### Description

`openwiki code --update` aborts near the end of a run with

```
Cannot advance page coverage for /openwiki/<page>.md; Markdown and verified Claims are not durable.
```

The message comes from `throwPageCoverageError` at `dist/generation/page-manifest.js:262` (openwiki 0.5.0), called from `buildManifestEntry` at `dist/generation/page-manifest.js:228`. That guard refuses to write a manifest entry unless `ClaimsStore.loadPage(page)` and `ClaimsStore.hashPage(page)` both succeed, the loaded record carries a `verification` object, and the record's stored `pageVersion` equals the hash of the Markdown currently on disk.

So the run does not fail on a model or network error - it fails on the durability guard of the grounded-claims store, after something earlier left a page half-written: Markdown updated, Claims either missing, unverified, or verified against an older hash.

What makes this hard to live with is that the guard turns a recoverable per-page problem into a total run failure: every page generated before it is discarded and no pull request is produced, so a repository in this state can never make progress. There is no `--skip-unverified` style escape and no partial-output mode we could find.

The failure correlates with repository/wiki size rather than with the model or the version:

- a small repository (one-digit page count) completes green on 0.5.0 with the same free OpenRouter models, the same workflow and the same runner image;
- three larger repositories (about 15 pages of existing wiki and up, generated by earlier openwiki versions) fail on 0.5.0 every time, each on a different page;
- the same repositories were green on 0.3.3 and started failing after 0.4.x, which is the release that introduced grounded claims (`feat: add grounded claims for self-correcting code wikis`, #638). Nothing else about the setup changed.

We suspect an existing wiki written before the claims store existed, or a page whose generation is interrupted mid-way (free models do time out and return empty responses under load), leaves exactly the state the guard rejects - and that a longer run on a larger repository simply gives that far more chances to happen.

### Steps to Reproduce

1. Have a repository with an existing `openwiki/` wiki of roughly 15+ pages, generated by an earlier openwiki version (pre-0.4.0 claims store).
2. `npm install --global openwiki` (0.5.0).
3. Run `OPENWIKI_PROVIDER=openrouter OPENWIKI_MODEL_ID=<a free model> openwiki code --update --print`.
4. Let it run to the end - the failure appears late, after tens of minutes of work, on one page.

The same command on a small repository with a handful of pages completes normally.

### Expected Behavior

A page whose Claims are missing, unverified or stale should not destroy the whole run. Either the page is regenerated and re-verified, or it is reported and skipped so the remaining pages are still written and a pull request is still produced.

At minimum, the error should say which of the three conditions failed (no record / no `verification` / `pageVersion` mismatch), so it can be repaired by hand.

### Actual Behavior

The run aborts with the message above, on the first page that trips the guard. No output is written, no pull request is produced, and the repository stays stuck: every subsequent run hits the same state.

The message does not distinguish the three failure conditions, and there is no stack, so locating it required reading the shipped bundle.

### Environment

- OS: Ubuntu 24.04 (GitHub Actions `ubuntu-latest`, later a self-hosted Linux runner - identical behaviour)
- Node.js version: 24
- OpenWiki version: 0.5.0 (last known-good on the same repositories: 0.3.3)

### Additional Context

Provider is OpenRouter with free models (`nvidia/nemotron-3-super-120b-a12b:free`, `nvidia/nemotron-3-ultra-550b-a55b:free`), `OPENWIKI_PROVIDER_RETRY_ATTEMPTS=10`. Both candidates fail identically, one after the other, so this does not look model-specific.

Two observations that may be relevant:

- On one run the first model worked for 54 minutes before the guard fired; the second failed on the same page within 4 minutes, which suggests the bad on-disk state persists between attempts within the same run.
- We also see a separate `Cannot read properties of undefined (reading '0')` on one other large repository, reported with only `error.message` and no stack. We cannot tell yet whether it shares a root cause with this one; we will file it separately once we have a stack.

Happy to test a patch or provide more detail from a run.

Totoro-qaq:
Fixes #826

An update can accept a plan that omits an existing page without verified Claims. The selected pages complete, but finalization rejects the omitted page. Subsequent runs restore an already-complete queue and fail again without giving the planner or workers a chance to fill the gap.

This change adds required verification jobs for those existing pages and repairs older interrupted queues by adding omitted work while preserving completed jobs and their provenance. The required-page set is persisted so repeated plan submissions remain idempotent, including after recovery and when Claims issues have been resolved.

Explicit deletions and caller-provided page instructions are respected. Pages with valid verified Claims are not rewritten solely because their manifest is missing or their source checkpoint is older. Final durability checks remain enforced. A patch changeset is included.

## Validation

- Added regressions for partial/empty legacy plans, old checkpoint recovery, duplicate submission, overlapping Claims/coverage jobs, source drift, explicit deletion, and valid existing coverage.
- Applying this patch alone: **89 focused generation tests passed**. The full `pnpm test` command passed typecheck, build, and coverage: **3048 tests passed, 3 skipped**. `pnpm run lint:check` and `pnpm run format:check` also passed.
- Actual CLI checks with deterministic local HTTP model responses: two-page and fifteen-page partial plans now complete. Copies of the failed 0.5.0 directories recover by authoring only omitted pages; the already-completed quickstart is not reauthored.

The reproduction establishes one concrete path matching #826; it does not claim to identify every trigger in the reporter's original OpenRouter environment.

### DIFF
--- a/.changeset/restore-legacy-page-coverage.md
+++ b/.changeset/restore-legacy-page-coverage.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Include existing pages without verified Claims in update plans and recover older interrupted queues that omitted them, preserving already-completed work and the final durability checks.
--- a/src/generation/page-jobs.ts
+++ b/src/generation/page-jobs.ts
@@ -117,6 +117,7 @@ export interface ProposedRepositoryPlan {
  * @param proposed - Complete proposed page and deletion set.
  * @param claimIssues - Stable preflight issues that require page work.
  * @param requiredRewritePages - Existing pages requiring language rewrites.
+ * @param requiredCoveragePages - Existing pages requiring verified Claims.
  * @returns Complete normalized plan ready for durable persistence.
  * @throws RepositoryRunError when paths, deletion intent, or init shape is invalid.
  */
@@ -125,6 +126,7 @@ export function createRepositoryPlan(
   proposed: ProposedRepositoryPlan,
   claimIssues: readonly GroundingIssue[],
   requiredRewritePages: readonly string[] = [],
+  requiredCoveragePages: readonly string[] = [],
 ): RepositoryRunPlan {
   const pages = proposed.pages.map(normalizePlanPage);
   const deletePages = uniqueSorted(
@@ -176,6 +178,7 @@ export function createRepositoryPlan(
     const deleted = new Set(deletePages);
     addRequiredClaimIssueJobs(pages, pagePaths, deleted, claimIssues);
     addRequiredRewriteJobs(pages, pagePaths, deleted, requiredRewritePages);
+    addRequiredCoverageJobs(pages, pagePaths, deleted, requiredCoveragePages);
   }
 
   // Quickstart is the synthesis/navigation page; generate it after domain pages.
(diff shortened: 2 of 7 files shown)


Totoro-qaq:
@colifran Could you take a look when you have a moment? This fixes #826, a user-reported 0.5.0 regression where finalization aborts the whole update on large existing wikis. The fork `Checks` workflow still needs approval; Socket and Corridor are green and the full `pnpm test` / lint / format pass locally (see description). #829 is a small independent fix for the sibling finalization bug (#827). Thanks!

Totoro-qaq:
Rebased onto current `main` and resolved the conflict with #848 (only the `paths.js` import overlapped). This is still needed after #848: on `main`, `update` never queues existing pages that lack verified Claims, so they stay uncovered — #848 only keeps finalization from failing on them. The new tests fail on `main` and pass here; the full suite passes.

PR #838, opened 2026-09-08
fix: accept empty-string env vars in MCP connector env resolution

### LINKED ISSUE #800: test(mcp-client): incomplete value-leak check + falsy envValue guard rejects empty-string vars
## Summary

Two bugs found in `test/mcp-client.test.ts` and the underlying `buildChildEnv` implementation.

---

### Bug 1 — Test: value-leak assertion only checks one of six secrets (line 42–44)

**File:** `test/mcp-client.test.ts`

```ts
// A random full-process.env secret must never leak by value either.
expect(Object.values(childEnv)).not.toContain(
  "secret-value-for-ANTHROPIC_API_KEY",
);
```

The `beforeEach` sets six `SECRET_KEYS` to `"secret-value-for-<KEY>"`, but the value-leak assertion only checks one of them (`ANTHROPIC_API_KEY`). A regression that leaked `OPENAI_API_KEY`, `TAVILY_API_KEY`, `SLACK_CLIENT_SECRET`, `GMAIL_ACCESS_TOKEN`, or `X_REFRESH_TOKEN` by value would pass this test undetected.

**Fix:** iterate over all secret keys:
```ts
for (const key of SECRET_KEYS) {
  expect(Object.values(childEnv)).not.toContain(`secret-value-for-${key}`);
}
```

---

### Bug 2 — Source: `!envValue` rejects env vars intentionally set to `""`

**File:** `src/connectors/mcp-client.ts` · `resolveEnvReference` (line ~836)

```ts
const envValue = process.env[envKey];
if (!envValue) {   // ← treats "" the same as undefined
  throw new Error(`${envKey} is required for MCP connector ingestion.`);
}
```

A falsy check means an env var explicitly set to an empty string (`""`) is treated as missing and throws, even though the variable IS present. No test covers this case.

**Fix:** use a strict undefined check:
```ts
if (envValue === undefined) {
  throw new Error(`${envKey} is required for MCP connector ingestion.`);
}
```

And add a test:
```ts
test("accepts an env var set to an empty string", () => {
  process.env.MCP_EMPTY = "";
  const childEnv = buildChildEnv({ MY_KEY: "${MCP_EMPTY}" });
  expect(childEnv.MY_KEY).toBe("");
  delete process.env.MCP_EMPTY;
});
```

---

## Affected files

- `test/mcp-client.test.ts` — value-leak assertion loop (Bug 1)
- `src/connectors/mcp-client.ts` — `resolveEnvReference` guard (Bug 2)

ousamabenyounes:
## What and why

`resolveEnvReference` in the MCP connector guarded its "missing required
credential" error with a falsy check:

```ts
const envValue = process.env[envKey];
if (!envValue) {
  throw new Error(`${envKey} is required for MCP connector ingestion.`);
}
```

`!envValue` treats an env var explicitly set to an empty string (`""`) the same
as an unset one, so an MCP connector declaring `env: { MY_KEY: "${MCP_EMPTY}" }`
with `MCP_EMPTY=""` aborts ingestion even though the variable **is** present.
This is inconsistent with `buildChildEnv`'s own base-var handling a few lines up,
which already accepts empty strings via `typeof value === "string"`.

The fix switches to a strict `envValue === undefined` check so only a truly unset
variable counts as missing. An empty string is a present value and passes through.

This PR also closes a test gap in the same suite: the child-env value-leak
assertion only checked one of the six seeded `SECRET_KEYS`, so a regression
leaking any of the other five by value would have gone undetected. It now loops
over every secret key.

Fixes #800

## How it was tested

New test `accepts a declared env var set to an empty string` in
`test/connectors/mcp-client.test.ts`, plus the widened value-leak loop.

### Test verification (RED → GREEN)

RED — new test on the unmodified source (before the `=== undefined` fix):

```
 ❯ test/connectors/mcp-client.test.ts (53 tests | 1 failed)
     × accepts a declared env var set to an empty string
 Tests  1 failed | 52 passed (53)
```

GREEN — with the fix applied:

```
 Test Files  1 passed (1)
      Tests  53 passed (53)
```

Full suite (`pnpm run format:check`, `lint:check`, `typecheck`, `build`, smoke,
`pnpm test` with coverage):

```
 Test Files  225 passed | 2 skipped (227)
      Tests  3039 passed | 3 skipped (3042)
```

The one changed production line (`src/connectors/mcp-client.ts` guard) is covered
by the suite. A changeset (`patch`) is included.

### DIFF
--- a/.changeset/mcp-empty-env-var.md
+++ b/.changeset/mcp-empty-env-var.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Accept MCP connector env vars that are explicitly set to an empty string. Previously a variable set to `""` was treated as missing and aborted ingestion with "<VAR> is required for MCP connector ingestion.".
--- a/.changeset/mcp-empty-env-var.md
+++ b/.changeset/mcp-empty-env-var.md
@@ -2,4 +2,4 @@
 "openwiki": patch
 ---
 
-Accept MCP connector env vars that are explicitly set to an empty string. Previously a variable set to `""` was treated as missing and aborted ingestion with "<VAR> is required for MCP connector ingestion.".
+fix: accept empty-string env vars in MCP connector env resolution
--- a/src/connectors/mcp-client.ts
+++ b/src/connectors/mcp-client.ts
@@ -900,7 +900,10 @@ function resolveEnvReference(value: string): string {
   validateEnvKey(envKey);
   const envValue = process.env[envKey];
 
-  if (!envValue) {
+  // An env var explicitly set to "" is present, not missing. Only a truly
+  // unset variable (undefined) is a missing required credential; mirror the
+  // `typeof value === "string"` check buildChildEnv uses for base vars.
+  if (envValue === undefined) {
     throw new Error(`${envKey} is required for MCP connector ingestion.`);
   }
 
--- a/test/connectors/mcp-client.test.ts
+++ b/test/connectors/mcp-client.test.ts
@@ -32,6 +32,7 @@ describe("buildChildEnv", () => {
       "APPDATA",
       "LOCALAPPDATA",
       "MCP_SERVER_TOKEN",
+      "MCP_EMPTY",
     ]) {
       saved[key] = process.env[key];
     }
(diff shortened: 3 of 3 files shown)


colifran (approved):
Thanks, @ousamabenyounes!

PR #846, opened 2026-09-09
fix: render model request output

### LINKED ISSUE #844: [Bug]: Interactive chat drops ALL assistant text with deepagents 1.12 — model_request namespace misclassified as subgraph
### Description

With deepagents ≥ 1.12, model calls are wrapped in a `model_request` subgraph. OpenWiki's `getStreamSource()` classifies **any** non-empty LangGraph namespace as `"subgraph"`, so every assistant text event emitted from inside the model call now carries `source: "subgraph"`. The interactive chat run-log reducer intentionally discards subgraph prose (`src/cli/run-log/reducer.ts`: `if (event.source === "subgraph" || event.text.length === 0) return log;`), which means **all** assistant text is silently dropped and the TUI shows:

```
✓ Completed openwiki chat in Xs · <model>
  No assistant output captured.
```

This is distinct from #803 / #817 (updates-mode chunk parsing): even with `messages` stream mode enabled (or with the #817 updates-mode parser), the chunks still arrive with namespace `["model_request:<id>"]`, get tagged `subgraph`, and are discarded by the reducer. As far as I can tell this affects **every provider** on current main, not just openai-compatible — anything that renders chat through the run-log reducer loses the assistant's reply text.

### Root cause

`src/agent/index.ts`:

```ts
function getStreamSource(namespace: unknown): "main" | "subgraph" {
  return Array.isArray(namespace) && namespace.length > 0 ? "subgraph" : "main";
}
```

With deepagents 1.12, a plain conversational turn produces namespaces like:

```
["model_request:5b3d…"]        ← 100% of the model's text chunks
["tools"]                       ← tool executions (correctly subgraph)
```

So the "discard subgraph prose" rule, which was meant to hide sub-agent/tool narration, now hides the primary model response.

### Steps to Reproduce

1. `npm i -g openwiki@0.5.0` (ships deepagents 1.12.x)
2. Configure any provider (verified with `OPENWIKI_PROVIDER=openai-compatible` + MiniMax-M3 via OpenAI-compatible endpoint; also set `OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true` in the shell env to rule out #803)
3. Run `openwiki`, send any tool-free message, e.g. `你好`
4. Observe "No assistant output captured" while the model demonstrably answered (verified by instrumenting the stream loop: 148 chunks, 96 parsed text events, all with `source: "subgraph"` and non-empty text).

### Expected Behavior

Assistant text from the top-level model call is displayed in interactive chat.

### Suggested fix

Treat the single-level `model_request` namespace (the wrapper deepagents puts around the model call itself) as main-graph output, e.g.:

```ts
function getStreamSource(namespace: unknown): "main" | "subgraph" {
  // deepagents ≥1.12 wraps the model call in a single-level `model_request`
  // subgraph; its text chunks are the primary model response, not sub-agent
  // prose, so they must render as main output.
  if (
    Array.isArray(namespace) &&
    namespace.length === 1 &&
    typeof namespace[0] === "string" &&
    namespace[0].startsWith("model_request:")
  ) {
    return "main";
  }
  return Array.isArray(namespace) && namespace.length > 0 ? "subgraph" : "main";
}
```

Verified locally: with this change (and `OPENWIKI_OPENAI_COMPATIBLE_STREAM_MESSAGES=true` to get messages-mode chunks on 0.5.0), interactive chat displays the assistant reply correctly end-to-end. Genuine sub-agent output (deeper namespaces, `tools`, etc.) is still suppressed as intended.

### Environment

- OpenWiki version: 0.5.0 (also confirmed present in `src/agent/index.ts` on `main` as of 2026-09-09)
- deepagents: 1.12.0 (bundled with 0.5.0)
- Node.js: v22.x
- OS: Ubuntu 24.04 (WSL2)
- Provider: openai-compatible (MiniMax-M3); bug is provider-agnostic

ousamabenyounes:
## Motivation

DeepAgents wraps primary model output in a single `model_request:<id>` namespace. OpenWiki treated every non-empty namespace as subgraph output, so interactive chat discarded the assistant response and displayed "No assistant output captured."

Closes #844.

## Changes

- Classify a single top-level `model_request:<id>` namespace as main-agent output.
- Keep deeper model-request namespaces and other non-empty namespaces classified as subgraph output.
- Cover the parser-to-run-log path and the nested-subgraph guard.
- Add a patch changeset for the user-visible fix.

## Validation

- RED on untouched `origin/main`: the focused regression received `source: "subgraph"` instead of `source: "main"`.
- GREEN with the patch: `test/agent/stream-redaction.test.ts` passes 15/15.
- Production-fix revert proof: the same focused regression returns to RED.
- Full local suite: formatting, lint, typecheck, build, CLI dry-run, coverage suite, and Windows-selected tests completed successfully.
- Baseline: 3,046 passed, 2 unrelated timeout failures, 3 skipped. Patched: 3,050 passed, 0 failed, 3 skipped.
- Changed production-line coverage: 100% (3/3 executable lines).

### DIFF
--- a/.changeset/render-primary-model-stream.md
+++ b/.changeset/render-primary-model-stream.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: render assistant text emitted from top-level model request streams
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1838,12 +1838,23 @@ function parseToolStreamEvent(payload: unknown): OpenWikiRunEvent | null {
   return null;
 }
 
+const MODEL_REQUEST_NAMESPACE_PREFIX = "model_request:";
+
 /**
- * Classifies a stream namespace. LangGraph reserves the empty namespace for
- * the root graph; even a single namespace segment therefore belongs to a
- * subgraph.
+ * Classifies a stream namespace. DeepAgents wraps the primary model call in a
+ * single model_request namespace, while deeper namespaces still represent
+ * subgraphs whose prose should stay hidden from the main transcript.
  */
 function getStreamSource(namespace: unknown): "main" | "subgraph" {
+  if (
+    Array.isArray(namespace) &&
+    namespace.length === 1 &&
+    typeof namespace[0] === "string" &&
+    namespace[0].startsWith(MODEL_REQUEST_NAMESPACE_PREFIX)
+  ) {
+    return "main";
+  }
+
   return Array.isArray(namespace) && namespace.length > 0 ? "subgraph" : "main";
 }
 
--- a/test/agent/stream-redaction.test.ts
+++ b/test/agent/stream-redaction.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, test } from "vitest";
 import { parseAgentStreamChunk } from "../../src/agent/index.ts";
+import { appendRunLogEvent } from "../../src/cli/run-log/reducer.ts";
+
+const MODEL_REQUEST_NAMESPACE = "model_request:fake-request-id";
+const MODEL_RESPONSE_TEXT = "Hello from the primary model.";
 
 function makeChunk(
   contentBlocks: unknown[],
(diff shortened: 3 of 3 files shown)


colifran (approved):
Thanks, @ousamabenyounes!

PR #829, opened 2026-09-07
fix: preserve verification metadata on skipped pages

Totoro-qaq:
Fixes #827

---

When a page worker fails, OpenWiki restores that page's original Markdown and Claims. Finalization then visits the skipped page again and removes its existing OpenWiki `verified` metadata, while preserving the Claims sidecar and page-manifest entry. Their stored hashes consequently no longer match the restored Markdown.

Pass the existing excluded-page set through Claims verification projection so skipped pages retain their exact restored bytes and matching coverage. Other pages continue through normal verification reconciliation. Includes a patch changeset.

Validation:

- 19 focused Claims/verification tests passed. Regression coverage spans verification projection, the Claims runtime, and the repository lifecycle.
- Applying this patch alone: `pnpm test` passed typecheck, build, and coverage (**3041 tests passed, 3 skipped**). `pnpm run lint:check` and `pnpm run format:check` also passed.
- A real-filesystem lifecycle probe creates the verified baseline through normal generation APIs. Before the fix, skipping restores consistent state, but finalization removes the verification stamp and breaks both hash comparisons. After the fix, the original Markdown, Claims sidecar, and manifest entry are preserved, and both hashes match. The completed-page control remains consistent.

The probe uses local fixture content without model calls. A skipped update still records `interrupted` metadata so its unfinished work can be retried.

PR #848, opened 2026-09-09
fix: only restamp page-manifest entries a run actually regenerated

### LINKED ISSUE #837: Concurrent runs always conflict on openwiki/.page-manifest.json, even on disjoint pages: every run restamps every entry with a whole-repo sourceFingerprint
### Description

Two `update`-mode runs made on separate branches from the same base cannot be merged by git, even when they regenerate completely different pages. Every run rewrites the `gitHead` and `sourceFingerprint` of every entry in the page manifest, so both branches modify every line of the file. The only safe resolution is to discard one run and replay it on top of the other, because a hand merge risks pairing one side's page body with the other side's `pageVersion` hash.

### Steps to Reproduce

1. On a repo with an existing wiki (63 pages in our case), branch A: run `openwiki` in update mode scoped to page X. Commit.
2. From the same base, branch B: run it scoped to page Y (no overlap with X). Commit.
3. Merge A into main, then try to merge B.

### Expected Behavior

Only the manifest entries for X and Y change, and the merge is clean.

### Actual Behavior

All 63 entries' `gitHead` and `sourceFingerprint` lines differ on both branches. Observed on the second such collision: 61 conflicting hunks, 60 of them the repeated stamp pair and one real page-entry disagreement. Six consecutive wiki commits each changed all 126 stamp lines while touching 2–6 pages.

Cause, as read in `dist/`: run finish calls `replaceRepositoryPageManifest(root, discoverPages(), checkpoint, skippedPages)` (`generation/repository-run.js:878`); only `skippedPages` are preserved (`generation/page-manifest.js:157-171`), so untouched-but-tracked pages are rebuilt with the run's checkpoint. That checkpoint's `sourceFingerprint` is one value per run hashing every model-visible file plus `git status --porcelain` (`agent/utils.js:169-218`), so it changes on every commit anywhere in the repo.

### Environment

openwiki 0.5.0, macOS, Node via Homebrew; driven through the MCP lifecycle (`openwiki_begin` mode `update`) and through `openwiki code --update`.

### Additional Context

Would either of these be acceptable: (a) preserve the previous entry for any page the run did not complete, so the manifest diff is scoped to regenerated pages; or (b) a per-page `sourceFingerprint` over the page's cited sources rather than the whole repo. Happy to test a patch. Possibly related: #827 (finalization removing metadata from skipped pages touches the same preserve path).

kowshikdev:
Fixes #837

## Root cause

`finishRepositoryRun()` calls `replaceRepositoryPageManifest(root, discoverPages(), checkpoint, skippedPages)`, and `replaceRepositoryPageManifest()` only preserved the prior manifest entry for pages in `skippedPages`. Every other currently-tracked page — including pages entirely outside this run's plan — was rebuilt with the run's own `checkpoint`.

That checkpoint's `sourceFingerprint` hashes every model-visible file plus `git status --porcelain` for the *whole repository*, so it's a new value on every single run regardless of which pages that run actually regenerated. The result: two `update` runs on separate branches touching disjoint pages both restamp every entry in `openwiki/.page-manifest.json`, and the branches can never be merged cleanly by git (every line conflicts, as described in the issue with 61/126 conflicting hunks from a 2–6 page change).

## Fix (issue's suggested approach a)

`finishRepositoryRun()` already computes `producerActorsByPage` — pages whose manifest entry has `completedRunId === run.state.runId`, i.e. pages this run actually completed. I reuse that as the "regenerated this run" set and pass `replaceRepositoryPageManifest()` a broader preserve set: every currently-tracked page *except* the ones this run regenerated (not just formally-skipped ones).

`replaceRepositoryPageManifest()` itself now:
- keeps a preserved page's exact prior entry unchanged when one exists (the common case for a mature wiki with existing coverage — this is what actually fixes the issue), and
- only builds a fresh entry for a preserved page when it has no prior coverage at all (first-time tracking), tolerating that build failing by leaving the page uncovered rather than failing the whole run — this matches the existing behavior for skipped pages that were never durably covered.

I considered approach (b) (per-page `sourceFingerprint` over each page's own cited sources) but (a) is the smaller, more surgical change: it doesn't touch how `sourceFingerprint` is computed or how completion currency is checked (`getCurrentRepositoryPageCompletion`), it only changes which pages get restamped at manifest-replace time.

## What I verified

- `npx tsc -p tsconfig.json --noEmit` — clean.
- `npm run build` — clean (tsc x2 + asset copy).
- `test/generation/repository-run.test.ts` and `test/generation/page-manifest.test.ts` — all tests pass except two pre-existing failures unrelated to this change and reproducible on `main` before my edits (`preserves durable state and cleans up when atomic rename fails` — a Windows fs-rename-simulation quirk, and `preserves mixed page baselines in a fresh checkout after partial progress` — pre-existing flake); confirmed both fail identically on a clean checkout of `main`.
- Added a new regression test, `finishRepositoryRun > only restamps pages a disjoint run actually regenerated`, reproducing the issue's scenario: run 1 completes page A and finishes, an unrelated source file changes, run 2 completes disjoint page B and finishes — asserts page A's manifest entry is byte-for-byte unchanged and only page B gets the new checkpoint.
- Updated two existing tests (`applies explicit existing-page deletions with Claims cleanup`, `finalizes completed work once without advancing a drifted source`) whose assertions asserted the old universal-restamp behavior (i.e. asserted the bug); updated them to assert only genuinely-regenerated pages get the new checkpoint stamp.

Related: #829 (open PR for #827) touches the same preserve path for a different concern (verification metadata on skipped pages) — this change doesn't overlap with or duplicate that scope.


colifran (approved):
thanks, @kowshikdev!

PR #856, opened 2026-09-10
fix: route vertex grok ids to the openai-compatible surface

### LINKED ISSUE #855: [Bug]: Vertex AI Grok model IDs route to the native Gemini surface and always fail
### Description

On the `gemini-enterprise` provider, xAI Grok model IDs are classified as the `gemini` surface and sent to Google's native `generateContent` endpoint. Grok is a Model Garden partner model and is only served over the OpenAI-compatible `/endpoints/openapi/chat/completions` surface, so every run fails immediately.

`resolveVertexSurface()` in `src/agent/vertex-surface.ts` keys off `OPENAI_MAAS_MODEL_PATTERN`, which lists `ai21|codellama|codestral|deepseek|jamba|llama|meta|mistral|qwen`. Neither `xai` nor `grok` is in it, so `xai/grok-4.6` falls through to the `gemini` default. `createGeminiEnterpriseModel()` then builds a `ChatGoogle` client, and `stripPublisherPath()` drops the `xai/` prefix, leaving the bare `grok-4.6`. `@langchain/google` hardcodes `get publisher() { return "google"; }`, so the request goes to:

```
POST https://aiplatform.googleapis.com/v1/projects/<project>/locations/global/publishers/google/models/grok-4.6:generateContent
```

That model does not exist under the `google` publisher.

### Steps to Reproduce

1. Enable an xAI Grok model in Vertex AI Model Garden on a project with ADC available.
2. `export OPENWIKI_PROVIDER=gemini-enterprise`, `OPENWIKI_MODEL_ID=xai/grok-4.6`, `GOOGLE_CLOUD_PROJECT=<project>`.
3. Run `openwiki --init -p` in any repository.

### Expected Behavior

Grok IDs route to the `openai-maas` surface, the same as every other partner model, so the request goes to `.../endpoints/openapi/chat/completions` with the model ID `xai/grok-4.6` (which `toVertexPublisherModel()` already passes through unchanged).

### Actual Behavior

The run dies on the first model call:

```
Invalid JSON payload received. Unknown name "exclusiveMinimum" at
'tools[0].function_declarations[4].parameters.properties[3].value.any_of[0]': Cannot find field.
```

The message points at a second, independent bug. `exclusiveMinimum` is emitted by zod for constraints like `.positive()`, and LangChain's Gemini tool converter only strips `additionalProperties` and `$schema` (`removeAdditionalProperties` in `@langchain/google/converters/tools`), so it reaches the wire for every model on this path. Fixing the routing does not fix that — it only moves Grok onto the OpenAI-compatible surface, which accepts standard JSON Schema. Conversely, fixing the schema would not fix Grok: `publishers/google/models/grok-4.6` 404s regardless.

> **Correction.** This paragraph originally claimed that real Gemini models accept `exclusiveMinimum`, citing a successful `gemini-3.8-flash` run on this repo. That was wrong. The run predated openwiki 0.5.1 and used 0.5.0 / deepagents 1.12.0, which has no `grep.max_count` param and emits no exclusive bounds at all, so it never sent the keyword — it was not a valid control. deepagents 1.13.2 (openwiki 0.5.1) introduced `max_count: z.coerce.number().int().positive()`, which is where it comes from. See [@gbmerrall's comment](https://github.com/langchain-ai/openwiki/issues/855#issuecomment-5627364543) for a minimal curl repro against AI Studio.

There is no way to work around this from configuration: no environment variable overrides the surface.

### Environment

- OS: Windows 11 / ubuntu-latest runner (reproduced on both)
- Node.js version: 22
- OpenWiki version: 0.5.1

### Additional Context

[Vertex AI docs for xAI Grok](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/partner-models/grok) confirm the OpenAI-compatible endpoint and the `xai/`-prefixed model ID.

The fix is to add `grok` and `xai` to `OPENAI_MAAS_MODEL_PATTERN` — both tokens, matching how `meta`/`llama` are both listed so an ID is accepted whether given as the publisher or the model family. Happy to open a PR.

radekfojtik:
Fixes #855.

`resolveVertexSurface()` had no `xai`/`grok` token, so Grok model IDs fell through to the `gemini` default and got sent to `publishers/google/models/grok-…:generateContent`, which doesn't exist. Grok is a Model Garden partner model and is only served over the OpenAI-compatible endpoint, so `gemini-enterprise` with `xai/grok-4.6` failed on the first model call every time.

Adds both tokens to `OPENAI_MAAS_MODEL_PATTERN`, the same way `meta`/`llama` are both listed, so the ID works whether you pass the publisher or the model family. Nothing else needed changing — `toVertexPublisherModel()` already passes `xai/grok-4.6` through unchanged, which is exactly what the endpoint wants.

Tested with three new cases in `test/agent/vertex-surface.test.ts` covering `xai/grok-4.6`, the bare `grok-4.1-fast-reasoning`, and the full `publishers/xai/models/grok-4.6` path. They fail on main and pass with the change. `pnpm run typecheck`, `pnpm run build`, `pnpm run lint:check` and `pnpm run format:check` are all clean.

### DIFF
--- a/.changeset/route-vertex-grok-to-maas.md
+++ b/.changeset/route-vertex-grok-to-maas.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: route Vertex AI xAI Grok model IDs to the OpenAI-compatible surface
--- a/src/agent/vertex-surface.ts
+++ b/src/agent/vertex-surface.ts
@@ -12,18 +12,20 @@ import { ANTHROPIC_API_KEY_ENV_KEY } from "../config/constants.js";
  * - `anthropic`: Claude over `rawPredict`/`streamRawPredict`, Anthropic's own
  *   wire protocol, reached via the Anthropic Vertex SDK.
  * - `openai-maas`: partner/open-weight models (Llama, Mistral, DeepSeek, Qwen,
- *   …) over the OpenAI-compatible `/endpoints/openapi/chat/completions` surface.
+ *   Grok, …) over the OpenAI-compatible `/endpoints/openapi/chat/completions`
+ *   surface.
  */
 export type VertexSurface = "anthropic" | "gemini" | "openai-maas";
 
 // These two patterns are the maintenance point when Vertex adds publishers:
-// extend the relevant alternation. `meta`/`llama` are both listed intentionally
-// so IDs are tolerated whether given as the publisher (`meta/…`) or the model
-// family (`llama-…`). `codellama` is listed separately because the `(^|\/)`
-// boundary means the `llama` token would not match the bare `codellama-…` form.
+// extend the relevant alternation. `meta`/`llama` and `xai`/`grok` are both
+// listed intentionally so IDs are tolerated whether given as the publisher
+// (`meta/…`, `xai/…`) or the model family (`llama-…`, `grok-…`). `codellama` is
+// listed separately because the `(^|\/)` boundary means the `llama` token would
+// not match the bare `codellama-…` form.
 const ANTHROPIC_MODEL_PATTERN = /(^|\/)(anthropic|claude)/u;
 const OPENAI_MAAS_MODEL_PATTERN =
-  /(^|\/)(ai21|codellama|codestral|deepseek|jamba|llama|meta|mistral|qwen)/u;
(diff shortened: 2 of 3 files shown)


radekfojtik:
@colifran PTAL

colifran (approved):
Nice. Thanks, @radekfojtik!

PR #859, opened 2026-09-10
fix: ignore windows ctime drift while fingerprinting

### LINKED ISSUE #857: [Bug]: "Error Source path changed while fingerprinting <file>" with v0.5.0
### Description

- **Bug:** Repeated `Source path changed while fingerprinting <file>` errors
- **Version:** OpenWiki 0.5.0
- **Platform:** Windows 11 Enterprise, Node.js v24.14.0

Hi @colifran, hope you’re doing well! 

Thanks for addressing #778 in #781. But I’ve encountered a related edge case on Windows where `ctime` are updated after file is opened because I work with an enterprise-managed windows machine and my company forces a file metadata change after opening a file. 

My understanding is this check aims to ensure **the opened file handle refers to the same file object that was inspected**(correct me if I’m wrong), rather than requiring its contents or metadata to remain unchanged. Exact timestamp equality seems too restrictive for that purpose. 

---

### Steps to Reproduce

What I have reproduced, independently of OpenWiki, is:
1. Write a file, then inspect it with `lstat()`.
2. Open it read-only and query its metadata through the handle.
3. `ctimeNs` differs, while `dev`, `ino`, size, `mtimeNs`, and `birthtimeNs` remain unchanged.

Here's my test script

```js
import { mkdtemp, writeFile, lstat, open, unlink, rmdir } from 'node:fs/promises';
import { tmpdir } from 'node:os';
import path from 'node:path';

const dir = await mkdtemp(path.join(process.argv[2] ?? tmpdir(), 'openwiki-ctime-'));
const file = path.join(dir, 'AGENTS.md');
const fields = ['dev', 'ino', 'size', 'mtimeNs', 'ctimeNs', 'birthtimeNs'];
try {
  await writeFile(file, '# Disposable reproduction\n', { flag: 'wx' });
  for (let attempt = 1; attempt <= 2; attempt++) {
    const before = await lstat(file, { bigint: true });
    const handle = await open(file, 'r');
    try {
      const after = await handle.stat({ bigint: true });
      console.log(JSON.stringify({
        attempt,
        changed: fields.filter(key => before[key] !== after[key]),
        ctimeBefore: String(before.ctimeNs),
        ctimeAfter: String(after.ctimeNs)
      }));
    } finally {
      await handle.close();
    }
  }
} finally {
  await unlink(file);
  await rmdir(dir);
}
```

Observed here: attempt 1 reports ["ctimeNs"]; attempt 2 reports []. No file contents are read between the two metadata checks. This is environment-specific evidence, not a claim that every Windows installation reproduces it.

### Expected Behavior

-

### Actual Behavior

-

### Environment

- OS: Windows 11 Enterprise, version 10.0.26200 (build 26200)
- Node.js: v24.14.0
- OpenWiki: 0.5.0

### Additional Context

_No response_

HwangJohn:
﻿## Summary
- Stop treating Windows-only `ctimeNs` drift between `lstat()` and `FileHandle.stat()` as a source path replacement during repository source fingerprinting.
- Keep the fail-closed replacement checks for non-regular files and for Windows `size`, `mtimeNs`, or `birthtimeNs` changes.
- Add deterministic regression coverage for ctime-only drift and for replacement-like Windows metadata changes.

Fixes #857.

## Testing
- Windows: `pnpm run format`
- Windows: `pnpm run lint`
- Windows: `pnpm exec vitest run test/agent/repository-source-fingerprint.test.ts -t "Windows file handles" --reporter=dot`
- Windows: `git diff --check`
- DGX Spark Linux: `pnpm exec vitest run test/agent/repository-source-fingerprint.test.ts --reporter=dot`
- DGX Spark Linux: `pnpm run format:check`
- DGX Spark Linux: `pnpm run lint:check`
- DGX Spark Linux: `pnpm run typecheck`
- DGX Spark Linux: `pnpm test`


colifran (approved):
looks good. thanks!

PR #861, opened 2026-09-10
fix: normalize openai-compatible text content wrappers

### LINKED ISSUE #626: openwiki --init fails with 400 against self-hosted vLLM: doubly-nested message content rejected
## Bug description

`openwiki code --init` fails with a 400 from the model provider when running against a self-hosted vLLM server exposed through an OpenAI-compatible endpoint (`OPENAI_COMPATIBLE_BASE_URL` / `OPENAI_COMPATIBLE_API_KEY`, model `gpt-oss-120b` served via vLLM with `--tool-call-parser=openai --enable-auto-tool-choice --reasoning-parser=openai_gptoss`).

## Error

```
400: 1 validation error for Message
content.0
  Input should be a valid dictionary or instance of Content [type=model_type, input_value=[{'text': '     1\t# Tech...y/dir', 'type': 'text'}], input_type=list]
    For further information visit https://errors.pydantic.dev/2.11/v/model_type None
```

The `input_value` is numbered file-listing content (line-numbered text, likely the output of a `read_file`-style tool call being replayed into a follow-up message).

## Root cause (best guess)

vLLM's OpenAI-compatible endpoint applies strict Pydantic validation on incoming `messages[*].content`. It appears OpenWiki/DeepAgents is sending `content` as a doubly-nested list — `[[{"type": "text", "text": "..."}]]` — instead of the expected flat `[{"type": "text", "text": "..."}]`. OpenAI and OpenRouter appear to tolerate/normalize this malformed shape silently, but vLLM correctly rejects it. This matches a similar report in another LangChain-based agent framework: https://github.com/apecloud/ApeRAG/issues/1396

## Environment

- openwiki v0.3.1
- Node v22.22.1
- Provider: OpenAI-compatible endpoint (self-hosted vLLM behind an internal MLflow AI Gateway)
- Model: `gpt-oss-120b`

## Repro

1. Configure `~/.openwiki/.env` with `OPENAI_COMPATIBLE_BASE_URL`, `OPENAI_COMPATIBLE_API_KEY`, `OPENWIKI_MODEL_ID` pointed at a vLLM server.
2. Run `openwiki code --init` in a non-trivial repository (large enough that a file-read tool call happens).
3. Observe the 400 error above; the run ends early and `openwiki/index.md` is written with no real content despite `.last-update.json` reporting `"status": "complete"`.

Note: a small, single-turn tool-calling test against the same vLLM endpoint (a simple `list_files` tool call, no multi-turn history) succeeds — this only reproduces once a tool result gets replayed back into the conversation as a follow-up message.

HwangJohn:
﻿## Summary
- Normalize redundant nested text content arrays on OpenAI-compatible Chat Completions requests before strict vLLM servers validate the payload.
- Keep the change scoped to the OpenAI-compatible chat-completions transport; other ChatOpenAI providers, Responses API opt-in, non-chat requests, non-JSON bodies, already-flat content, and non-text/multimodal blocks pass through unchanged.
- Add regression coverage that reproduces the nested wire payload before the fetch normalizer and verifies the pass-through cases.

Fixes #626.

## Testing
- Windows: `pnpm exec vitest run test/agent/create-model.test.ts test/openai-compatible-streaming.test.ts test/openai-compatible-responses.test.ts --reporter=dot`
- Windows: `pnpm run format`
- Windows: `pnpm run format:check`
- Windows: `pnpm run lint`
- Windows: `pnpm run lint:check`
- Windows: `git diff --check`
- DGX Spark Linux: `pnpm exec vitest run test/agent/create-model.test.ts test/openai-compatible-streaming.test.ts test/openai-compatible-responses.test.ts --reporter=dot`
- DGX Spark Linux: `pnpm run format:check`
- DGX Spark Linux: `pnpm run lint:check`
- DGX Spark Linux: `pnpm run typecheck`
- DGX Spark Linux: `pnpm test`

### DIFF
--- a/.changeset/openai-compatible-vllm-content.md
+++ b/.changeset/openai-compatible-vllm-content.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+Normalize redundant nested text content arrays for OpenAI-compatible Chat Completions requests so strict vLLM servers accept replayed tool output.
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1241,14 +1241,21 @@ export function createModel(
   }
 
   const baseURL = resolveProviderBaseUrl(provider);
+  const configuration =
+    provider === "openai-compatible" && !chatOpenAiUsesResponsesApi
+      ? {
+          ...(baseURL ? { baseURL } : {}),
+          fetch: createOpenAiCompatibleFetch(),
+        }
+      : baseURL
+        ? {
+            baseURL,
+          }
+        : undefined;
 
   return new ChatOpenAI({
     apiKey: getProviderApiKey(provider),
-    configuration: baseURL
-      ? {
-          baseURL,
-        }
-      : undefined,
+    configuration,
     model: modelId,
     useResponsesApi: chatOpenAiUsesResponsesApi,
     ...maxTokensOptions,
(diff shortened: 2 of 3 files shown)


colifran (approved):
Thanks, @HwangJohn!

Issue #813, opened 2026-09-04
[Bug]: Copilot provider (claude-sonnet-5) repository generation crashes with "expected AIMessage or Command, got object" in OpenWikiRepositoryWorkerNoDelegation


RajeshLTIM:
## Description

`openwiki --init` (repository/code mode) fails on the first model call when the provider is `copilot` with a Claude model, in the repository-generation page-job path (`src/agent/repository-runner.ts`), not the chat-mode agent path:

```
Error Invalid response from "wrapModelCall" in middleware "OpenWikiRepositoryWorkerNoDelegation": expected AIMessage or Command, got object
```

This looks related to #213 / #659 (same validator, same error text) but is a **distinct, unfixed path**:

- #213/#659's fix (PR #660) only special-cases `provider !== "openai-compatible"` in `src/agent/index.ts`'s `runOpenWikiAgentCore`, dropping the `"messages"` stream mode to avoid `ChatMessageChunk` aggregation from providers that stream reasoning deltas before the first `role:"assistant"` delta.
- The repository-generation path used by `--init`/`--update` (`runNativeRepositoryGeneration` → `runPlanningAgent`/`runPageAgent` in `repository-runner.ts`) never uses `"messages"` stream mode at all — `streamWorkerTools` only streams `["tools"]`. So the #660 fix's root cause (chunk-aggregation from a streamed roleless-first delta) does not apply here, meaning something else is producing a non-`AIMessage` object from the model call in this path.
- The middleware named in the error, `OpenWikiRepositoryWorkerNoDelegation`, only exists in `repository-runner.ts` (added to strip the `task` tool from DeepAgents 1.12+). It just forwards `wrapModelCall`'s result untouched — it isn't the source of the invalid object, only the outermost middleware that surfaces the agent loop's rejection.

## Steps to Reproduce

1. Configure the GitHub Copilot provider with a Claude model:
   ```
   OPENWIKI_PROVIDER=copilot
   OPENWIKI_MODEL_ID=claude-sonnet-5
   COPILOT_API_KEY=<GitHub OAuth token>
   ```
2. Run `openwiki --init` in any repository (code/repository mode).
3. The first agent turn (planning worker) crashes with the error above. No wiki is produced.

## Expected Behavior

The run completes and the wiki is generated, as it does with other providers/models.

## Actual Behavior

[...]

PR #862, opened 2026-09-10
fix: coerce roleless repository worker stream messages

### LINKED ISSUE #849: OPENWIKI_OPENAI_COMPATIBLE_STREAMING=true fails in wrapModelCall middleware (0.4.3 and 0.5.0)
## Description

Setting `OPENWIKI_OPENAI_COMPATIBLE_STREAMING=true` (the documented remedy for gateways that only serve the streaming transport) makes every repository run fail at its first model call, in both **0.4.3** and **0.5.0**:

```
Invalid response from "wrapModelCall" in middleware "OpenWikiRepositoryWorkerNoDelegation": expected AIMessage or Command, got object
```

## Environment

- openwiki 0.4.3 and 0.5.0 (npm, verified against both)
- provider: `openai-compatible` against a GLM endpoint (z.ai coding plan, `https://api.z.ai/api/coding/paas/v4`), model `glm-5.3-flash`
- Node 22

## Repro

```bash
export OPENWIKI_PROVIDER=openai-compatible
export OPENAI_COMPATIBLE_API_KEY=…
export OPENAI_COMPATIBLE_BASE_URL=https://api.z.ai/api/coding/paas/v4
export OPENWIKI_MODEL_ID=glm-5.3-flash
export OPENWIKI_OPENAI_COMPATIBLE_STREAMING=true
openwiki code --update --print
```

Fails within ~30 seconds, before any page work. Remove the streaming flag and the same command runs.

## Analysis

The flag reaches the model constructor as `streaming: true` (`agent/index.js`, the
`providerUsesStreaming(provider) ? { streaming: true } : {}` spread). With that set, the value
returned through the middleware-wrapped model call is an aggregate object rather than an
`AIMessage` instance, and the LangGraph middleware contract check rejects it — the middleware
itself (`OpenWikiRepositoryWorkerNoDelegation`) only filters the `task` tool and passes the
handler's return through unchanged, so the breakage is in the interaction, not the middleware's
own logic.

## Why it matters

The non-streaming transport is not viable on some gateways independently of this flag: Node's
undici defaults `headersTimeout` to 300s and a non-streaming chat completion sends no response
headers until generation completes, so any turn that reasons longer than five minutes is killed
by the HTTP client, mid-run, with retries walking into the same wall. The streaming flag is the
only OpenAI-side escape from that — which makes it load-bearing, and currently unusable.

## Workaround we use

Switch to an Anthropic-compatible endpoint (`OPENWIKI_PROVIDER=anthropic` +
`ANTHROPIC_BASE_URL`): that client streams SSE by default inside the agent loop, so response
headers arrive immediately and the undici wall cannot fire, without the constructor flag that
trips the middleware.

HwangJohn:
## Summary

- Coerce repository worker `wrapModelCall` responses that are assistant-like `ChatMessage` / `ChatMessageChunk` objects into `AIMessage` / `AIMessageChunk` before LangChain validates the middleware return value.
- Preserve message content, response metadata, reasoning kwargs, and raw OpenAI tool calls during the coercion.
- Leave non-assistant generic model responses untouched so unrelated invalid responses still fail closed.

## Why

OpenAI-compatible endpoints that require `OPENWIKI_OPENAI_COMPATIBLE_STREAMING=true` can emit a reasoning delta before the first `role: "assistant"` delta. LangChain then aggregates the response as a generic `ChatMessageChunk`, and the repository worker fails its first model call with:

```text
Invalid response from "wrapModelCall" in middleware "OpenWikiRepositoryWorkerNoDelegation": expected AIMessage or Command, got object
```

This keeps the compatibility shim at the repository worker boundary where the invalid response surfaces, without changing the lower-level OpenAI-compatible request normalization path.

Fixes #849.
Related #813.

## Tests

- Windows: `pnpm exec vitest run test/openai-compatible-streaming.test.ts test/agent/stream-modes.test.ts test/agent/repository-runner.test.ts --reporter=dot`
- Windows: `pnpm run format:check`
- Windows: `pnpm run lint:check`
- Windows: build-equivalent `tsc -p tsconfig.json`, `tsc -p tsconfig.client.json`, `node scripts/copy-visualize-assets.cjs`, and postbuild chmod
- DGX Spark Ubuntu fresh clone: `pnpm install --frozen-lockfile`
- DGX Spark Ubuntu fresh clone: focused 3-file Vitest suite above
- DGX Spark Ubuntu fresh clone: `pnpm run format:check`
- DGX Spark Ubuntu fresh clone: `pnpm run lint:check`
- DGX Spark Ubuntu fresh clone: `pnpm run typecheck`
- DGX Spark Ubuntu fresh clone: `pnpm test` (225 files passed, 2 skipped; 3069 tests passed, 3 skipped)

### DIFF
--- a/.changeset/openai-compatible-worker-streaming.md
+++ b/.changeset/openai-compatible-worker-streaming.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: coerce roleless OpenAI-compatible streaming worker responses
--- a/src/agent/repository-runner.ts
+++ b/src/agent/repository-runner.ts
@@ -1,6 +1,17 @@
 import { scheduler } from "node:timers/promises";
 import type { BaseChatModel } from "@langchain/core/language_models/chat_models";
-import { ToolMessage } from "@langchain/core/messages";
+import {
+  AIMessage,
+  AIMessageChunk,
+  ChatMessage,
+  ChatMessageChunk,
+  ToolMessage,
+  collapseToolCallChunks,
+  defaultToolCallParser,
+  type InvalidToolCall,
+  type ToolCall,
+  type ToolCallChunk,
+} from "@langchain/core/messages";
 import { DynamicStructuredTool } from "@langchain/core/tools";
 import { createDeepAgent, createFilesystemMiddleware } from "deepagents";
 import { createMiddleware } from "langchain";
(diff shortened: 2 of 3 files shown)


colifran (approved):
Thanks, @HwangJohn!

PR #869, opened 2026-09-10
fix: add claude opus 5 to the anthropic and vertex model lists

### LINKED ISSUE #868: [Bug]: Claude Opus 5 warns that it belongs to GitHub Copilot on the anthropic and gemini-enterprise providers
### Description

Claude Opus 5 is missing from the `modelOptions` of the two providers that serve Claude directly (`anthropic` and `gemini-enterprise`), but it is listed under `copilot`. `isModelIdForOtherProvider()` therefore treats it as a plain provider/model mismatch, and every run prints:

```
Warning: model "claude-opus-5" is not a known Gemini Enterprise (Vertex AI) model
(it belongs to GitHub Copilot). The request may fail. Set OPENWIKI_MODEL_ID to a
Gemini Enterprise (Vertex AI) model, or switch providers.
```

The warning is wrong on both counts: `claude-opus-5` is the correct bare model ID on the Anthropic API *and* on Vertex AI (current-generation Claude models on Vertex use the bare first-party ID; only dated snapshots take the `@version` separator, which is why the curated list already carries `claude-haiku-4-5@20251001`). Routing is fine — `resolveVertexSurface()` matches on `claude` and sends it to the Anthropic Vertex SDK — so the run proceeds and succeeds. It is purely the curated list being a generation behind: `PROVIDER_CONFIGS.anthropic` and `PROVIDER_CONFIGS["gemini-enterprise"]` both still top out at `claude-opus-4-8`.

Beyond the noise, Opus 5 is also absent from the setup picker for both providers, so it can only be reached by typing the ID by hand.

### Steps to Reproduce

1. `export OPENWIKI_PROVIDER=gemini-enterprise` and `OPENWIKI_MODEL_ID=claude-opus-5` (same with `OPENWIKI_PROVIDER=anthropic`).
2. Run `openwiki --init -p` in any repository.
3. Read the first lines of stderr.

### Expected Behavior

No warning. `claude-opus-5` is a valid model for both providers, and it should be offered in the setup picker alongside the other Claude tiers.

### Actual Behavior

Every run warns that the model "belongs to GitHub Copilot" and suggests switching providers. Generation then works normally, so the warning only trains people to ignore a diagnostic that exists to catch real misconfiguration.

### Environment

- OS: ubuntu-latest runner / Windows 11
- Node.js version: 22
- OpenWiki version: 0.5.1

### Additional Context

`openrouter` is unaffected — its IDs are namespaced (`anthropic/claude-opus-4-8`) and `getProvidersForKnownModelId()` matches exactly, which the existing tests already pin. So the fix is scoped to the two providers that use bare Claude IDs.

Happy to open a PR adding `claude-opus-5` to both lists and keeping `claude-opus-4-8` as an explicitly versioned entry.

radekfojtik:
Fixes #868.

`claude-opus-5` was only listed under `copilot`, so running it on `anthropic` or `gemini-enterprise` printed "it belongs to GitHub Copilot" on every single run. The ID is correct for both — Vertex serves current-generation Claude models under the bare first-party ID — and routing already worked, so the warning was pure noise on a working config. It also meant Opus 5 never showed up in the setup picker for either provider.

Added it to both lists and kept `claude-opus-4-8` as an explicitly versioned entry rather than dropping it, since it's still served and the existing tests use it as their fixture for "known to more than one provider". `openrouter` needs nothing — its IDs are namespaced, so it never collided in the first place.

Tested with a new case in `test/config/constants.test.ts` asserting `isModelIdForOtherProvider("claude-opus-5", ...)` is false for both providers; it fails on main and passes here. The other 107 assertions in that file are untouched and still pass, and `pnpm run typecheck`, `lint:check` and `format:check` are clean.

### DIFF
--- a/.changeset/add-claude-opus-5-model-options.md
+++ b/.changeset/add-claude-opus-5-model-options.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: add Claude Opus 5 to the Anthropic and Gemini Enterprise model lists
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -369,7 +369,8 @@ export const PROVIDER_CONFIGS: Record<OpenWikiProvider, ProviderConfig> = {
     modelOptions: [
       { id: "claude-haiku-4-5", label: "Haiku" },
       { id: "claude-sonnet-5", label: "Sonnet" },
-      { id: "claude-opus-4-8", label: "Opus" },
+      { id: "claude-opus-5", label: "Opus" },
+      { id: "claude-opus-4-8", label: "Opus 4.8" },
     ],
   },
   gemini: {
@@ -392,7 +393,8 @@ export const PROVIDER_CONFIGS: Record<OpenWikiProvider, ProviderConfig> = {
       ...GEMINI_MODELS,
       { id: "claude-haiku-4-5@20251001", label: "Claude Haiku" },
       { id: "claude-sonnet-5", label: "Claude Sonnet" },
-      { id: "claude-opus-4-8", label: "Claude Opus" },
+      { id: "claude-opus-5", label: "Claude Opus" },
+      { id: "claude-opus-4-8", label: "Claude Opus 4.8" },
     ],
   },
   openrouter: {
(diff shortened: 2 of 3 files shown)


colifran (approved):
Thanks, @radekfojtik!

PR #881, opened 2026-09-12
fix: classify symlinked claim evidence as unresolved instead of aborting update

### LINKED ISSUE #867: `update` aborts with an opaque error when a claim's evidence path resolves through a symlinked directory
## Environment

- openwiki 0.5.1 (also reproduced on 0.4.3), installed globally via npm
- macOS 15 (Darwin 25.6.0), Node via Homebrew
- Host-driven mode: `openwiki mcp --host claude`

## What happens

`openwiki_begin` with `mode: "update"` fails with

```
OpenWiki MCP operation failed.
```

as soon as **any** file cited by an existing claim's evidence is reached through a
symlinked directory. The message carries no detail — same text on stdout and
stderr, with `force: true`, and with `OPENWIKI_DEBUG`, `DEBUG`, `LOG_LEVEL`,
`OPENWIKI_LOG_LEVEL` and `NODE_ENV=development`.

## What I expected

Either resolve the symlink and validate the evidence normally, or skip that
evidence with a warning. A symlinked source directory is a common layout — it is
how a repository can expose a submodule tree, a vendored checkout, or a shared
working copy under a stable path.

## Minimal reproduction

Two files, no dependencies. `repro_run.py` (attached below) drives one full
host-driven lifecycle in a single process.

```bash
# 1. a repo with one source file
mkdir -p /tmp/repro /tmp/external/lib
cat > /tmp/external/lib/greeter.py <<'PY'
def greet(name):
    """Return a greeting."""
    return f"Hello, {name}!"
PY
cd /tmp/repro && git init -q .
mkdir lib && cp /tmp/external/lib/greeter.py lib/
printf '# Demo\n\nA tiny library with one function.\n' > README.md
git add -A && git -c user.email=r@example.com -c user.name=r commit -qm "initial commit"

# 2. init a wiki — produces a claim whose evidence is repo://lib/greeter.py
python3 repro_run.py /tmp/repro init

# 3. change the file and commit, then update — WORKS
printf '\n# touched\n' >> lib/greeter.py
git add -A && git -c user.email=r@example.com -c user.name=r commit -qm touch
python3 repro_run.py /tmp/repro update      # -> status=active, phase=planning

# 4. replace lib/ with a symlink to byte-identical content, then update — FAILS
cp lib/greeter.py /tmp/external/lib/greeter.py
rm -rf lib && ln -s /tmp/external/lib lib
python3 repro_run.py /tmp/repro update      # -> "OpenWiki MCP operation failed."
```

Step 3 and step 4 differ **only** in whether `lib` is a directory or a symlink;
the file content is identical (`cmp` clean).

## Additional findings

Narrowed down on a real repository with 35 source trees under one directory,
testing each individually and resetting `.openwikiignore` to a known state at
every step:

- **It takes both conditions.** A symlinked tree that no claim cites is fine; a
  tree with claims is fine when it is a real directory. 27 trees failed and 8
  passed, and the split correlated perfectly, in both directions, with whether
  any existing claim cited that path. Deterministic over 5 repetitions each.
- A rename test isolates the path from the content: the same tree under a name no
  claim cites passes, and an otherwise-passing tree renamed to a name with claims
  fails.
- **Not data volume.** The full tree as a real copy — 26.313 files — passes,
  while a symlinked tree holding 2 files fails.
- **Not git nesting.** Same result whether the symlink target is a git submodule
  (`.git` as a gitdir-pointer file), an independent clone, or has no git at all.
- `.openwikiignore` does not help: excluding the symlinked path as `repos/` or as
  `repos` (no trailing slash) changes nothing — the failure happens before the
  ignore rules apply. Note also that a rule **without** a trailing slash on a
  directory-shaped entry makes `begin` fail on its own, even for a tree that
  otherwise passes; that may be a second, separate bug.

## Second, separable issue: the error is undiagnosable

`"OpenWiki MCP operation failed."` with no stack, no path, no cause, on every
channel and at every log level, turned a one-line root cause into a long
bisection. v0.5.1 added a debug diagnostics panel with error stacks (#832) —
surfacing the same detail through the MCP tool result (or any env-gated stderr
trace) would make failures like this self-diagnosing.

## Workaround

Run against a real copy of the tree instead of the symlink
(`rsync -a --exclude '.git' <target>/ <repo>/<path>/`). Evidence paths stay
unchanged, so no claims are invalidated.

---

## `repro_run.py`

<details>
<summary>Script used in the reproduction above (drives one full host-driven lifecycle in a single process)</summary>

```python
#!/usr/bin/env python3
"""Drive one full OpenWiki host-driven lifecycle over stdio in a single process."""
import json, subprocess, sys, pathlib

ROOT = sys.argv[1]
MODE = sys.argv[2] if len(sys.argv) > 2 else "init"

p = subprocess.Popen(["openwiki", "mcp", "--host", "claude"],
                     stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                     stderr=subprocess.DEVNULL, text=True, bufsize=1)
_id = 0
def rpc(method, params=None, notify=False):
    global _id
    msg = {"jsonrpc": "2.0", "method": method}
    if params is not None: msg["params"] = params
    if not notify:
        _id += 1; msg["id"] = _id
    p.stdin.write(json.dumps(msg) + "\n"); p.stdin.flush()
    if notify: return None
    while True:
        line = p.stdout.readline()
        if not line: raise SystemExit("server closed")
        try: r = json.loads(line)
        except json.JSONDecodeError: continue
        if r.get("id") == _id: return r

def tool(name, args):
    r = rpc("tools/call", {"name": name, "arguments": args})
    res = r.get("result", {})
    txt = (res.get("content") or [{}])[0].get("text", "")
    if res.get("isError"): return {"_error": txt}
    try: return json.loads(txt)
    except json.JSONDecodeError: return {"_raw": txt}

rpc("initialize", {"protocolVersion": "2024-11-05", "capabilities": {},
                   "clientInfo": {"name": "repro", "version": "1"}})
rpc("notifications/initialized", notify=True)

st = tool("openwiki_begin", {"root": ROOT, "mode": MODE, "language": "en"})
if "_error" in st: raise SystemExit("begin failed: " + st["_error"])
if st.get("status") == "noop":
    print("noop — nothing to update"); raise SystemExit(0)
run = st["runId"]
print(f"begin: status={st['status']} phase={st.get('phase')}")

if st.get("phase") == "planning":
    plan = {"runId": run, "pages": [
        {"path": "/openwiki/library/greeter.md", "title": "Greeter",
         "purpose": "The greeting function and its contract.",
         "seedPaths": ["lib/greeter.py"]},
        {"path": "/openwiki/quickstart.md", "title": "Quickstart",
         "purpose": "How to use the demo library.", "seedPaths": ["README.md"]}]}
    r = tool("openwiki_submit_plan", plan)
    if "_error" in r: raise SystemExit("submit_plan failed: " + r["_error"])
    print("plan submitted")

BODIES = {
 "/openwiki/library/greeter.md": ("""---
type: module
title: Greeter
description: The single greeting function exposed by the demo library.
tags: [library, api]
---

# Greeter

`lib/greeter.py` exposes one function, `greet(name)`, which returns a
greeting string built from its single argument. It performs no I/O and
raises nothing, so callers need no error handling.
""", [{"statement": "greet(name) returns a greeting string built from its single argument and performs no I/O.",
       "evidence": [{"resource": "repo://lib/greeter.py"}]}]),
 "/openwiki/quickstart.md": ("""---
type: guide
title: Quickstart
description: How to call the demo library.
tags: [guide, onboarding]
---

# Quickstart

Import `greet` from `lib/greeter.py` and call it with a name. The
repository ships no build step and no dependencies.
""", [{"statement": "The repository ships no build step and no dependencies beyond the standard library.",
       "evidence": [{"resource": "repo://README.md"}]}]),
}

while True:
    job = tool("openwiki_next_page", {"runId": run})
    if "_error" in job: raise SystemExit("next_page failed: " + job["_error"])
    if job.get("status") == "complete":
        print("all pages complete"); break
    j = job.get("job") or {}
    path, jid = j.get("path"), j.get("id")
    body, claims = BODIES.get(path, (None, None))
    if body is None: raise SystemExit(f"no body prepared for {path}")
    f = pathlib.Path(ROOT + path)
    f.parent.mkdir(parents=True, exist_ok=True)
    f.write_text(body, encoding="utf-8")
    r = tool("openwiki_submit_page", {"runId": run, "jobId": jid, "claims": claims})
    if "_error" in r: raise SystemExit(f"submit_page failed for {path}: " + r["_error"])
    print(f"page done: {path}")

r = tool("openwiki_finish", {"runId": run})
print("finish:", json.dumps(r)[:200])
```

</details>

marliechorgan:
## What this fixes

If a repository has a claim whose evidence file is now reached through a symlinked directory, `openwiki --update` (and `openwiki_begin` with `mode: "update"` in host-driven mode) stops before doing anything, with only:

```
OpenWiki MCP operation failed.
```

One claim, one symlink, and the whole update is dead. That is #867, and the reporter's reproduction is exact: a repository where `lib/` is a real directory updates fine; replace `lib/` with a symlink to byte-identical content and the same update aborts.

After this change, that claim is reported as **unresolved**, the same way OpenWiki already reports a claim whose file was deleted. The update runs, every other page and claim is untouched, and the page that owns the claim is asked to re-cite it or retract it.

## Why it was happening

OpenWiki refuses to read evidence through a symlink. That is deliberate and correct: `repo://lib/greeter.py` has to mean the file physically inside the repository at that path, otherwise a claim could quietly cite something outside the repo. The resolver enforces this by throwing `EvidenceSecurityError`.

The problem is what preflight did with that error. Before an update, preflight re-checks every persisted claim's evidence, and it has a sensible rule: if resolving evidence *errors* (as opposed to returning "not found"), let the error propagate, because a permissions or I/O hiccup must never be mistaken for "the file was deleted". `EvidenceSecurityError` is a subclass of `EvidenceResolutionError`, so it fell under that rule. A permanent "this path is not allowed" was being handled as a transient "could not read it right now", and it took the whole run down instead of flagging one claim. The MCP layer then hides everything except `HostIntegrationError`, which is why the reporter saw no detail.

## What the change is

In `runClaimsPreflight`, when resolving one evidence resource throws `EvidenceSecurityError`, that resource is recorded as unresolved. Any other error is rethrown exactly as before. That is the entire code change, plus a comment on the function explaining the exception:

```ts
try {
  current = await cachedResolver.resolve(evidence.resource, evidence.version);
} catch (error) {
  if (!(error instanceof EvidenceSecurityError)) {
    throw error;
  }
  current = null;
}
```

Things I deliberately did not do:

- **Did not make the resolver follow symlinks.** That would let a claim cite a file outside the repository, which the containment check exists to prevent. The resolver is unchanged and still throws before it reads anything.
- **Did not add a new issue kind.** `unresolved` already means "this evidence does not resolve under the current rules, recheck it". A symlinked path is exactly that. No schema, prompt, or tool-output change.
- **Did not loosen anything on the write side.** If a page worker tries to re-cite the symlinked path when submitting, the same resolver refuses it (`src/claims/core/mutations.ts` resolves through it with no catch). If a security error reaches finalization, `isRecoverableFinalizationError` still treats it as fatal. Both boundaries are as they were.

## How I tested it

Three tests added to `test/claims/brains/code/preflight.test.ts`:

1. **The fix itself.** With a fixture resolver, a security error on one resource marks that claim unresolved. A healthy resource on the same claim, and a second healthy claim, are unaffected.
2. **The fence.** A plain `EvidenceResolutionError` (the parent class, an operational failure) still aborts preflight. This is what proves the catch is narrow.
3. **The issue's scenario, for real.** Using the actual `RepositoryEvidenceResolver`: a real `lib/` directory resolves cleanly, then `lib/` is replaced with a symlink to byte-identical content outside the root, and preflight reports the claim unresolved instead of throwing.

Without the source change, tests 1 and 3 fail and test 2 passes. With it, `pnpm test` (typecheck, build, Vitest with coverage) passes: 3069 tests, 0 failures, 3 pre-existing skips. `pnpm run format` and `pnpm run lint:check` are clean.

Not tested: a live host-driven run on a large repository like the reporter's 35-tree case. The change is confined to the deterministic preflight, so I would not expect a difference, but I have not measured it.

## Not in this PR

Two things from the issue that I left alone so this stays one change:

- **The masked error message.** `executeTool` still reduces every non-`HostIntegrationError` to `OpenWiki MCP operation failed.` The reporter is right that surfacing the cause would have made this a one-line diagnosis. Worth its own PR.
- **`.openwikiignore` exclusions.** An evidence path excluded by `.openwikiignore` still aborts preflight, through `EvidenceResourceError`. The reporter flagged it as a possible second bug. It is a different error class with different semantics (an ignore rule is something the user can edit), so it deserves a separate decision.

Changeset included (`patch`). Closes #867.

### DIFF
--- a/.changeset/symlinked-evidence-unresolved.md
+++ b/.changeset/symlinked-evidence-unresolved.md
@@ -1,0 +1,5 @@
+---
+"openwiki": patch
+---
+
+fix: report a claim whose evidence now traverses a symbolic link as unresolved instead of aborting the update
--- a/src/claims/brains/code/preflight.ts
+++ b/src/claims/brains/code/preflight.ts
@@ -1,5 +1,6 @@
+import { EvidenceSecurityError } from "../../core/errors.js";
 import { cacheEvidenceResolver } from "../../core/resolver-cache.js";
-import type { EvidenceResolver } from "../../core/types.js";
+import type { EvidenceResolver, ResolvedEvidence } from "../../core/types.js";
 import type { GroundingIssue, PageClaims } from "./types.js";
 import { ClaimsStore } from "./store.js";
 
@@ -28,6 +29,9 @@ export interface ClaimsPreflightResult {
  *
  * Each evidence resource and prior version resolves once per preflight.
  * Resolution errors propagate so they cannot be mistaken for deleted evidence.
+ * A containment refusal (`EvidenceSecurityError`) is the one exception: it is
+ * permanent for the cited resource rather than operational, so the claim is
+ * reported as unresolved for reconciliation instead of aborting the run.
  *
  * @param store - Repository claim persistence.
  * @param resolver - Repository evidence resolver.
@@ -57,10 +61,18 @@ export async function runClaimsPreflight(
       const unresolvedResources: string[] = [];
 
       for (const evidence of claim.evidence) {
-        const current = await cachedResolver.resolve(
-          evidence.resource,
-          evidence.version,
-        );
+        let current: ResolvedEvidence | null;
+        try {
+          current = await cachedResolver.resolve(
+            evidence.resource,
+            evidence.version,
+          );
+        } catch (error) {
+          if (!(error instanceof EvidenceSecurityError)) {
+            throw error;
+          }
+          current = null;
+        }
         if (!current) {
           unresolvedResources.push(evidence.resource);
         } else if (current.evidence.version !== evidence.version) {
--- a/test/claims/brains/code/preflight.test.ts
+++ b/test/claims/brains/code/preflight.test.ts
@@ -1,9 +1,14 @@
-import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, rm, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import { afterEach, beforeEach, describe, expect, test } from "vitest";
 import { runClaimsPreflight } from "../../../../src/claims/brains/code/preflight.ts";
 import { ClaimsStore } from "../../../../src/claims/brains/code/store.ts";
+import {
+  EvidenceResolutionError,
+  EvidenceSecurityError,
+} from "../../../../src/claims/core/errors.ts";
+import { RepositoryEvidenceResolver } from "../../../../src/claims/evidence/repository/resolver.ts";
 import { CODE_CLAIMS_SCHEMA_VERSION } from "../../../../src/claims/brains/code/types.ts";
 import type { PageClaims } from "../../../../src/claims/brains/code/types.ts";
 import type {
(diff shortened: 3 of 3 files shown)


colifran (approved):
Thanks, @marliechorgan!

PR #885, opened 2026-09-14
fix: tighter redaction of API keys in CredentialDiagnosticsPanel

### LINKED ISSUE #878: [Bug]: CredentialDiagnosticsPanel reveals more of a live API key than the codebase's own redaction standard allows elsewhere
### Description

There are two independent implementations of "redact a secret for debug display" in this codebase, and they disagree:

- `src/agent/index.ts:1300`, `formatDebugValue()` — for any key ending in `_API_KEY`, returns only `set(length=N)`. **No preview substring at all.**
- `src/env.ts:113`, `createCredentialPreview()` (used by `getCredentialDiagnostics()` → `CredentialDiagnosticsPanel`, `src/cli.tsx:673`) — for **any** value over 10 characters, including live values of `OPENROUTER_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, etc., returns a `first-6-chars...last-4-chars` preview with no exception for `_API_KEY` keys.

The result: with `OPENWIKI_DEBUG=1` (or `OPENWIKI_DEBUG_CREDENTIALS=1`) set, the interactive `CredentialDiagnosticsPanel` prints something like:

```
OPENROUTER_API_KEY source=process.env over ~/.openwiki/.env
length=73 preview="sk-or-...d1b8"
```

That's a real, partial fragment of a live credential rendered directly to the terminal — and a natural thing for a user to paste into a public bug report (I nearly did, before catching it), since it looks like an intentional, safe debug summary rather than a partial secret leak. The presence of `formatDebugValue()`'s stricter behavior elsewhere in the same codebase indicates this isn't the intended standard — `createCredentialPreview()` just didn't get the same treatment.

### Steps to Reproduce

1. `export OPENWIKI_DEBUG=1` (or `OPENWIKI_DEBUG_CREDENTIALS=1`).
2. Run `openwiki` with a real `OPENROUTER_API_KEY` (or any other `_API_KEY`) configured, longer than 10 characters.
3. Trigger the credential diagnostics view (e.g., via a run failure) and observe the `preview=` field for the API key.
4. Compare with any debug path that goes through `formatDebugValue()` in `src/agent/index.ts`, which withholds the preview entirely for the same key.

### Expected Behavior

`createCredentialPreview()` should apply the same `_API_KEY`-suffix exception `formatDebugValue()` already uses — i.e., report length only, no substring preview, for any credential-shaped key. Non-secret config values (`OPENWIKI_MODEL_ID`, `OPENWIKI_PROVIDER`) can keep showing full/partial values as they already do elsewhere in `createCredentialDiagnostic()`.

### Actual Behavior

Live API keys get a 10-character preview (6 prefix + 4 suffix) printed to the terminal under debug mode, inconsistent with the redaction level the codebase already enforces in its other debug-formatting path.

### Suggested Fix

In `src/env.ts`, make `createCredentialPreview()` (or its caller, `createCredentialDiagnostic()`) special-case keys ending in `_API_KEY` the same way `formatDebugValue()` in `src/agent/index.ts` does, so both debug surfaces agree on what's safe to show. Might also be worth extracting a single shared redaction helper so the two call sites can't drift again.

### Environment

- OpenWiki: local dev checkout, `main`-adjacent branch
- Discovered via `OPENWIKI_DEBUG=1` output after an OpenRouter "Key limit exceeded (total limit)" run failure (per-key spend cap, unrelated to this bug — just what triggered the diagnostics view).

green-creeper:
Resolves #878 by only revealing the last 4 characters in the panel's credential preview, rather than the first 6 and last 4, preventing partial leakage of structured tokens like `sk-` or `sk-or-v1-`. 

- Exported `createCredentialPreview` and updated its logic to only show the last 4 characters.
- Added unit tests to verify behavior.


colifran (dismissed):
Thanks, @green-creeper!

colifran (approved):

PR #903, opened 2026-09-17
feat: run repository page workers in parallel

### LINKED ISSUE #891: [Feature]: Add configurable parallel generation for independent documentation pages
### Summary

Add a configurable concurrency limit so documentation pages that do not depend on one another can be generated in parallel, reducing end-to-end generation time.

### Motivation

Generating documentation for medium or large repositories can take a significant amount of time. Many planned pages are independent, so they do not need to wait for one another. Allowing those pages to run concurrently would reduce total generation time while still preserving the correct order for pages that have dependencies.

### Proposed Solution

Add a configuration option that controls the maximum number of page-generation workers. For example:

```yaml
generation:
  max_concurrency: 4
```

The option could also be exposed through the CLI:

```bash
openwiki generate --max-concurrency 4
```

OpenWiki should:

- Identify pages that have no dependencies on one another and generate them concurrently.
- Preserve dependency ordering where dependencies exist.
- Use a safe default, with `1` preserving sequential behavior.
- Respect provider rate limits and existing retry policies.
- Isolate failures so one page-generation task does not silently cancel or corrupt other tasks.
- Keep progress and errors clear when multiple pages are running.

Acceptance criteria:

- Users can configure the maximum number of concurrent page-generation workers.
- Independent pages are generated concurrently.
- Dependent pages are generated in the correct order.
- Sequential generation remains available with concurrency set to `1`.

### Alternatives Considered

Keep generation fully sequential. This is simpler, but it leaves independent work waiting unnecessarily and does not address generation time for large repositories.

### Additional Context

This would be especially useful for repositories with many independent modules or components, where most documentation pages can be generated without waiting for other pages to finish.

eugeneliu-86:
## Summary

Adds `OPENWIKI_PAGE_CONCURRENCY` (1–8, default 1) so the native repository runner can document several pages at once. With the default unset, behavior is byte-for-byte what it is today: one fresh worker per page, in queue order, same progress line, same events.

Page workers are independent by design (each owns one page, gets only its own plan entry and Claims), so the change is mostly in the durable core's ownership model, not in the workers.

Theoretically, should be upwards of 8x faster!

## What changed

**Core (`src/generation/repository-run.ts`)**
- `inspect`, `snapshot`, `skip`, and `submit` accept any *pending* job by id instead of only the first pending job in queue order. Queue order remains the resume discipline; it was never a data dependency.
- `nextRepositoryPage` takes an optional `exclude` set so a driver can hand distinct jobs to several workers. Ownership stays process-local; `.run.json` still records only `pending | skipped | complete`, so resume is unchanged.
- `submit` and `skip` serialize their shared-state bookkeeping (Claims reconcile/finalize, page manifest, checkpoint write) through a per-run mutation lock keyed by run identity. Page-local validation and all model work stay outside the lock.
- A submit now finalizes with the other pending jobs' pages excluded. Previously finalization stamped verification into every eligible page's front matter and rehashed every sidecar; with one worker that was idempotent, with several it wrote into pages another worker was mid-edit on (found in a hard-kill test, see below).

**Runner (`src/agent/repository-runner.ts`)**
- Bounded pool of `N` worker loops with serialized job acquisition. Each worker is still a fresh, non-delegating deep agent confined to its page.
- With `N > 1`, quickstart is held back until every other page is done so its routing map links to pages that exist. With `N = 1` the plan's ordering already guarantees this, so nothing changes.
- A worker that fails on a provider rate limit (429 / "rate limit" in the error chain) lowers the live pool size by one, never below 1. The page is skipped and restored exactly as before.
- Runs with `N > 1` default to 5 model retries unless `OPENWIKI_PROVIDER_RETRY_ATTEMPTS` is set. First-wave worker starts are staggered by one second per slot.
- A fatal submission error stops new work, lets in-flight workers submit or skip, then rethrows, so `finish` never sees pending jobs.

**Config / UI / docs**
- New managed env key with diagnostics, `resolvePageConcurrency`, README section next to the retry-attempts entry.
- Progress events gain `completedCount` and `inFlightPages`; tool events gain `page`. The line becomes `Documenting 5 of 20 · 3 in flight: …` only when more than one page is in flight; otherwise the existing `Documenting page i of N · path` line is unchanged. Print mode gets the same line.
- MCP host path is unchanged and still drives one job at a time. Tool descriptions were reworded from "current pending" to "pending page job".

## Testing

**Unit.** New core tests: any pending job by id, concurrent submits land in one checkpoint and manifest, skip while a sibling is in flight, sibling submit leaves an in-flight page and sidecar byte-identical (this one fails on the previous code). New runner tests: N agents constructed before any submit and quickstart last, rate-limit failure lowers the pool, ordinary failure does not, fatal submission waits for in-flight workers, `N = 1` emits the historical event shape. Plus config, reducer, and progress-format tests. `pnpm lint:check`, `pnpm format:check`, and `pnpm test` pass (the 7 `test/visualize/client-interaction.test.ts` failures on my machine are the pre-existing Node 26 `localStorage` global issue and reproduce on `main`).

**Real runs** against a 20-page repository with `gpt-5.6-terra`:
- `N = 1` and `N = 4` full refreshes both finish clean: `.run.json` removed, 20 manifest entries under the run id, all 20 sidecar `pageVersion` hashes equal the page bytes, and a follow-up `--update` reports "already current" through the strict Claims preflight with zero model calls.
- Durability: `N = 4`, `kill -9` at 3/20 with 4 workers in flight. Checkpoint and manifest agreed on exactly the 3 durable pages. Resume kept the run id, did not re-plan or redo the 3 pages, finished the remaining 17 with 4 workers, and passed the same checks.
- The wall-clock tail is set by the single heaviest page (~30 min alone in this corpus), so gains are bounded by that page. Ordering heavy pages first would help; left as a follow-up.

## Notes for review

- Default is 1, so nobody's runs change unless they opt in.
- Changeset: minor.

---

Closes: #891


colifran (approved):
Awesome! Ran against openwiki with concurrency of 3 - total time was around 11min. Then re-ran with concurrency of 1 and total time was around 26min.

This is a great efficiency bump!

PR #913, opened 2026-09-19
fix: retry a page worker once before skipping its page

### LINKED ISSUE #850: Page worker ending without submit_page is skipped with no retry; run finalizes interrupted and the documented base never advances
## Description

A page worker whose stream ends **without calling `submit_page`** is marked `skipped`, its
snapshot restored, and the run finalizes with `status: "interrupted"` — **without advancing the
documented base** (`gitHead` in `openwiki/.last-update.json`) and with `openwiki/.run.json`
deleted. Neither 0.4.3 nor 0.5.0 retries the page. The next `--update` therefore diffs against
the same old base and re-researches the **entire changeset** again.

For a tool whose architecture is "resumable page jobs, partial progress banked", the missing
retry turns a single unlucky final turn into a full-cost regeneration loop.

## Observed trigger

A reasoning model that spends its entire completion budget on thinking and returns an empty final
message. With `glm-5.3-flash` (via an Anthropic-compatible endpoint), one failing worker made
**94 clean LLM calls, then ended its stream with no text and no tool call** — the final turn's
AIMessage content blocks are `thinking` only (diagnosed by pulling the worker's turns from
LangSmith and printing the content-block kinds; healthy turns are
`thinking,tool_use` / `thinking,text,tool_use`). The agent loop simply ends; nothing throws.

## Measured cost

Over four consecutive daily runs (2026-09-01..04), each ended `interrupted` on one such worker,
the base stayed frozen at the same commit, and every run re-researched the full 74-file changeset
— 8h, then 13h × 3. The completed pages *were* banked each time, but the un-advanced base means
the next run re-derives them anyway. The first run to finalize cleanly took 1.7h for what the
frozen-base runs were paying 13h.

A secondary observation, possibly by design but worth a look: a run that finalizes `interrupted`
deletes `.run.json`, so "rerunning after an interruption resumes the durable queue" holds only
for hard kills, not for finalized-interrupted runs — there is nothing left to resume, and the
skipped page's work is fully discarded.

## Suggested fix

One fresh retry of the page worker before marking it skipped — a new agent instance over the
same page job. Transient empty final turns then cost one retry instead of the whole page; the
deterministic case still skips as today.

We carry a local patch implementing exactly this (rewrites `runPageAgent` to construct the
worker agent per attempt and retry once when the stream ends without `submit_page`, preserving
the skip path when the retry also fails) and would be glad to turn it into a PR against either
line if the approach sounds right.

sloemo01:
Closes #850.

## What

A page worker whose stream ends without calling `submit_page` is now retried once with a fresh worker over the same page job. Between attempts the page and its Claims are restored to the pre-run snapshot, so both attempts start from the same state and edits from an unsubmitted attempt cannot leak into the retry. If the retry also exits without submitting, the page is skipped exactly as before, with the same warning and the same interrupted finalization.

A fatal submission failure still propagates immediately and is not retried.

## Why

The issue measured the cost: one worker ending its stream without `submit_page` freezes the documented base, so the next update re-researches the entire changeset (8h, then 13h three times, against 1.7h for a clean run). The completed pages are banked each time, but the frozen base means they get re-derived anyway. `skipRepositoryPage` already performed this rollback; `restoreRepositoryPage` is that rollback extracted so a retry can reuse it, and `skipRepositoryPage` behaves as before.

## How tested

- Existing skip-path tests updated to the new contract (a page is skipped only after every attempt exits without submitting), plus new tests: retry completes the page after an empty stream; retry completes the page after a stream failure; page skipped when both attempts exit without submitting; fatal submission failure propagates with no retry. Reverting the retry (attempt limit back to 1) makes all four fail.
- `repository-run.test.ts` covers `restoreRepositoryPage` directly: Markdown and Claims return to the snapshot while the job stays `pending` and the queue still offers it.
- `pnpm test` runs the new tests green. The 13 failures elsewhere (`test/visualize/client-interaction.test.ts` and the agent graph suites) reproduce on a clean `main` checkout, so they are not from this diff.

The issue's secondary observation, that a finalized `interrupted` run deletes `.run.json` so only hard kills are resumable, is not touched here. Changeset included (patch).


sloemo01:
Reopened after a rebase onto current `main`.

The branch was written against the page worker before #903 landed, which changed `runPageAgent` to return `PageAgentOutcome` and moved the skip handling out to `runWorkerLoop`. So the retry is applied on top of that contract: `runPageAgent` now drives up to two attempts, `runPageWorkerAttempt` is a single bounded attempt that returns its outcome, and the skip happens in the driver only once every attempt has exited without submitting.

One thing the port had to preserve: `runWorkerLoop` reads `outcome.status === "skipped"` and calls `isRateLimitError(outcome.error)` to drop a concurrency slot on a 429. My first version of the driver returned a skip without the error, which would have quietly disabled that. A rate-limited attempt is no longer retried at all and is surfaced to the pool with the error intact.

`pnpm run test` exits 0 on the current head (232 files, 3254 tests). Each piece of the retry is sabotage-checked: attempt limit back to 1, reset skipped before the retry, rate limits retried anyway, and the extracted helper restoring Markdown without Claims each fail only their own arms.

PR #917, opened 2026-09-21
fix(visualize): resolve URL-encoded filenames in graph links

changingshow:
## Summary

Graph generation matches Markdown link targets against filesystem paths without decoding them. For example, a link to `%E6%9C%B1%E6%A3%A3.md` does not resolve to the existing `朱棣.md` node, leaving out its edge and backlink.

Decode each link target once before resolving it. If decoding fails, retain the original path so filenames such as `progress100%.md` continue to work. Missing targets remain ignored.

Add regression coverage for Unicode and special characters, relative directories, anchor deduplication, backlinks, single decoding, unescaped percent signs, and malformed escapes. Includes an `openwiki` patch changeset.

## Validation

- All three added regression tests pass. The unescaped-percent regression failed before the fallback was added and passes afterward.
- Graph tests: 8 passed and 1 skipped when excluding the existing symlink test. Running that test on this Windows machine fails during symlink creation with `EPERM`; the test has not been modified.
- Typecheck, build, and repository-wide ESLint passed before the final wording-only edits. The changed files also pass targeted formatting and lint checks.
- Repository-wide Prettier passed after temporarily normalizing checkout line endings to the repository's LF format. Unchanged files were then restored to their original checkout format; no repository-wide line-ending changes are included.
- Full Windows test run before the wording-only edits: 3038 passed, 80 failed, 23 skipped. Failures include symlink permissions, platform paths, file modes, and timeouts. Their causes have not all been verified; this is not a passing full-suite result. Linux CI verification is still needed.

## Related work

PR #886 addresses directory links resolving to `index.md`, while this PR addresses URL-encoded filenames. Both touch the link-resolution code; either change should preserve the other's behavior if merged later.


colifran (approved):
thanks, @changingshow!

